PR #16, opened 2014-11-21
updated new functions for gmap_display

ryanlei:


shankari:
Not using google maps any more

PR #46, opened 2015-01-13
Improve the documentation in preparation for cs194

shankari:
- Add some figures
- Make the dependencies clearer
- Move the design discussion to the end, and focus on quick setup first


shankari:
Also includes a small typo fix from earlier.

PR #57, opened 2015-01-30
for purposes of pull request

gaukes:


shankari:
What is the actual change? Are you adding a new file that has your name in it?

gaukes:
Sorry this was for the pull request you told us to make in class. All I did
was add a file with my name.

shankari:
Right, but can you make it a useful pull request instead?
Maybe update the README with something that wasn't clear?

If you can't think of anything, then we can fall back to adding a file with your name...

gaukes:
No problem. I remember now that you wanted us to make a change to something
like the readme file with any issues or problems we came across. I'll
submit another pull request soon.

PR #59, opened 2015-01-30
Update README with Windows Mongodb notes

lechiang:


shankari:
I've merged this change now, but you should really create a branch to make changes instead of making them in the master branch.

PR #60, opened 2015-01-30
Updated README

gaukes:
Added instructions for installing pip on mac. Made slight modification to installation instructions for mongodb.


shankari:
I've merged this change for now, but in the future, your changes should really be on a separate branch, not on master.

PR #65, opened 2015-01-31
Update README.md

jeffdh5:
Listed out some python dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,9 @@ fill them in
         $ python api/cfc_webapp.py
 
    You may need to install some of the python dependencies from above
+   Amongst the dependencies include:
+   -cherrypy
+   -pygeocoder
 
 1. Browse to 
 [http://localhost:8080](http://localhost:8080)


README.md
```diff
@@ -66,6 +66,9 @@ fill them in
         $ python api/cfc_webapp.py
 
    You may need to install some of the python dependencies from above
+   Amongst the dependencies include:
+   -cherrypy
```
shankari: I believe you shouldn't need cherrypy unless you are running over SSL. Did you have an issue with cherrypy when you started with the default config?

jeffdh5: Using the default config, I got this traceback:


  File "api/cfc_webapp.py", line 409, in <module>

    run(host=server_host, port=server_port, server='cherrypy', debug=True)

  File "/Users/jeffdh5/Desktop/e-mission-server/CFC_WebApp/api/bottle.py",
line 2999, in run

    server.run(app)

  File "/Users/jeffdh5/Desktop/e-mission-server/CFC_WebApp/api/bottle.py",
line 2651, in run

    from cherrypy import wsgiserver

ImportError: No module named cherrypy

shankari: Ah, make sense. I should really move that import inside the "if ssl" code block.
I'll merge this change now - can you file an issue for the code fix so that I can get to it later?

PR #68, opened 2015-02-01
fixed buggy profiling statements

jeffdh5:


CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -83,20 +87,43 @@ def getSectionQueryWithGroundTruth(groundTruthMode):
...
+    logging.debug("Querying stage modes %s" % (datetime.now()))
+    begin = time.time()
     modeList = []
     for mode in MongoClient('localhost').Stage_database.Stage_Modes.find():
         modeList.append(mode)
         logging.debug(mode)
+    duration = time.time() - begin
+    logging.debug("Querying stage modes took %s" % (duration))
+    
+    logging.debug("Section query with ground truth %s" % (datetime.now()))
```
shankari: The log statements are automatically prefixed with the time (e.g.)
2015-01-30 00:30:02,699:INFO:loadTrainingDataStep DONE

So I don't think that you need another datetime.now() call.

CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -83,20 +87,43 @@ def getSectionQueryWithGroundTruth(groundTruthMode):
...
+    logging.debug("Querying confirmedSection took %s" % (duration))
+    
+    logging.debug("Querying stage modes %s" % (datetime.now()))
+    begin = time.time()
     modeList = []
     for mode in MongoClient('localhost').Stage_database.Stage_Modes.find():
         modeList.append(mode)
         logging.debug(mode)
+    duration = time.time() - begin
+    logging.debug("Querying stage modes took %s" % (duration))
```
shankari: It would be nice to put this into a separate function.

def logtime(function):
    begin = ...
    result = function()
    duration = ...
    logging.debug(....)
    return result

But it seems hard to refactor the inner code into nice functions that can be passed into log time.
Where are #DEFINES when you need them? :)

shankari:
Just remove the datetime.now() at the end of the log messages (no point in making unnecessary calls), and I'll merge the request.

Issue #48, opened 2015-01-22
Store raw sensed data to help in debugging conversion


shankari:
We parse the raw data from the phone in the collect method. As we change the raw data format, we need to ensure that the parsing is correct. This is easiest if we have the raw data stored on the server so that we can write tests more easily.

jesca:
we will be working on this !

jesca:
shankari --- it looks like you merged a branch (dumping the raw input data into log_base_dir)  -- is this issue still open?

shankari:
Fixed via #69

PR #69, opened 2015-02-01
We dump the raw input data into the log_base_dir

shankari:
We put this here since it is not clear that we want to do this after we have
stabilized the coversion from the raw data to the database format.

This addresses issue #48.

@bellettif, let's test out the admin workflow.
If you approve of this change, please merge the pull request into the main repo.


shankari:

PR #71, opened 2015-02-01
Move cal distance

shankari:
So that we get into the habit of doing code reviews, @bellettif, can you review and pull if you approve?
I don't want to keep merging my own requests...

So that we can start cleaning up this web of circular dependencies.
The main change is to tripManager.py and common.py.
All the others just fix dependencies.
    
All references to calDistance in tripManager are gone.
    
bash-3.2$ grep -r tripManager.*calDistance .
Binary file ./.git/.COMMIT_EDITMSG.swp matches
bash-3.2$
    
And the testCommon tests pass.
    
Ran 12 tests in 0.490s
OK

### DIFF
--- a/CFC_WebApp/main/DTW.py
+++ b/CFC_WebApp/main/DTW.py
@@ -1,5 +1,5 @@
 __author__ = 'Mogeng'
-from tripManager import calDistance
+from common import calDistance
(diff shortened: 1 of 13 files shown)


shankari:

PR #72, opened 2015-02-02
Handle corner cases in the feature matrix creation

shankari:
@bellettif - here's one more pull request for you to review. With this, I think that all the changes in my tree have been submitted as pull requests to the repo.

When we switched to our own data collection, we initially had some weird trips
generated, for example:
- trips with zero time, or
- trips with only one point

These exposed several corner cases where feature matrix creation failed.

While these trips were clearly incorrect and have since been fixed, it is also
good to handle the corner cases so that we fail gracefully.

This change fixes the corner cases and adds new test cases to avoid regressions
in the future.


shankari:

PR #73, opened 2015-02-03
Modified the readme to include db path requirement

Wuhandson:
Added the db path requirement to the Troubleshooting and a bonus fix was
to fix the formatting.


shankari:
Good job, Handson :)

PR #76, opened 2015-02-04
set up backend server

abhinavasingh16:


shankari:
Your pull request contains a bunch of key files.

These were intentionally NOT pushed to the public repo - we don't want any random person on the internet knowing the API key associated with our server, specially since sensitive data is stored in the server. If random people on the internet want to use this code, they need to get their own keys. I just made a set of keys available to the class as a courtesy since we are all UC Berkeley students.

Similarly with the keystore - the key stored in the keystore is used to sign the apk and restrict access to apps that are allowed to talk to the corresponding server. We don't want to share all this data with the rest of the world.

Please send an alternate pull request that is not such a huge security hole.

abhinavasingh16:
Sorry Shankari,

I was not aware, I will do that right now!

Thanks,

Abhi

PR #77, opened 2015-02-05
Got rid of keys

abhinavasingh16:


shankari:
Hi Abhinava,

All keys have not been removed. In fact, I see `CFC_DataCollector/keys.json` in this list.

Two of the other files are log files (moves_collect.log and pipeline.log) that are generated when you run code. Such files are typically not checked in, since they are regenerated for every user when they run code.

Finally, it is not clear that it is correct to check in a moves from config.json.ios.localhost - some people may want to use android, for example, which is why different sample files are checked in.

I could be wrong, but it seems to me that you are trying to check in the changes that you made in order to start the server. But most of those changes were intentional - either for security or customization reasons.

A better approach might be to check in additional documentation to clarify areas that were unclear by modifying the README, and then talk about whether it is possible to automate things so that the clarification is not necessary.

PR #78, opened 2015-02-05
Made more changes and checked in updated README as well as updated documentation.

abhinavasingh16:
I am unsure what you mean by what to check in? I am not as used to using github's pull request functionality so thank you so much for you patience, I will have this down by next week!


shankari:
This is better, but you are still renaming the config file, and I'm not sure that's the right option. Let's discuss in the hands-on section of class tomorrow (10:00 to 11:30) and get something checked in then.

Issue #21, opened 2014-11-28
We currently filter short sections as part of the getUnclassifiedSections call


shankari:
There's is no need to do this as part of a REST API call and slow it down. It should be done as part of batch processing, presumably while loading data from moves.

zacatac:
***Proposed solution***

**Database schema**
Section now has a new boolean field called 'filter'. This field defines whether or not the section should be filtered in the call to queryUnclassifiedSections 
```python
newSec['filter'] = True/False
```

**~Frontend**
When a client queries for unclassifiedSections a call is made to `queryUnclassifiedSections`, but the call to `filter_UnclassifiedSections` is no longer needed. In the `queryUnclassifiedSections` we filter out any section entry that is labeled with `section['filter'] = True`.

**Backend**
When new sections come from the moves database we will call a modified version of `filter_UnclassifiedSections` that will return `True/False` based on whether or not the section should be filtered on the next call to `'/tripManager/getUnclassifiedSections'`.

**Testing**
Unit tests for both the frontend and backend changes described above.


Contributing: @zacatac @wuhandson @liyeungjohn

shankari:
> ##Backend##
> When new sections come from the moves database we will call a modified version of 
Just to make sure that you have thought through the design sufficiently, what will you do with the `True/False` result?

##~Frontend##
Is this the front end or is the phone code the front end? This is an example of how there is backend and back backend, or in other words, layered architectures are popular.

zacatac:
We will use the `True/False` result to filter the results that we get from `queryUnclassifiedSections` by adding a parameter to the query found in `queryUnclassifiedSections`: 

```python
defaultQueryList = [ {'source':'Shankari'},
                         {'user_id':user_uuid},
                         {'predicted_mode': { '$exists' : True } },
                         {'confirmed_mode': ''},
                         { 'type': 'move' },
                         {'section_end_datetime': {"$gt": weekago}}]
```

As for existing trips we would have to look into adding the `filter` to all sections currently in the database. This may not be the best way, but we couldn't come up with another way to memoize the trips that are labeled as being to short. 

As for the frontend label, I wasn't sure what to label the distinction. 

Thanks for the feedback

shankari:
> As for existing trips we would have to look into adding the filter to all sections currently in the database. > This may not be the best way, but we couldn't come up with another way to memoize the trips as being > too short. 

[...]

PR #79, opened 2015-02-05
Prefiltering short sections when they are pulled from the moves server. Resolves issue #21

zacatac:
***Database schema***
Section now has a new boolean field called 'filter'. This field defines whether or not the section should be filtered in the call to queryUnclassifiedSections

```python
newSec['filter'] = True/False
```

**~Frontend**
When a client queries for `unclassifiedSections` a call is made to `queryUnclassifiedSections`, but the call to `filter_UnclassifiedSections` is no longer needed. In the `queryUnclassifiedSections` we filter out any section entry that is labeled with `section['filter'] = True`.

**Backend**
When new sections come from the moves database we will call a modified version of `filter_UnclassifiedSections` that will return `True/False` based on whether or not the section should be filtered on the next call to '/tripManager/getUnclassifiedSections'.

**Testing**
Unit tests for both the frontend and backend changes described above. Tests can be found in `

Contributing: @zacatac @wuhandson @liyeungjohn


CFC_WebApp/main/tripManager.py
```diff
@@ -87,6 +41,13 @@ def stripoutNonSerializable(sectionList):
         strippedList.append(section)
     return strippedList
 
+def filter_unclassifiedSections(UnclassifiedSections):
+    filtered_Sections=[]
+    for section in UnclassifiedSections:
+        if section['filtered']:
```
shankari: I have four comments here:

1. This check doesn't match the spec. Here, if section['filtered'] == True, then we retain the section, but the spec says:
> In the queryUnclassifiedSections we filter out any section entry that is labeled with section['filter'] = True

2. The code that sets the `'filtered'` field is also consistent with this description, and not with the spec. For example, in `label_filtered_section`, if the `distance > = minimum_travel_time`, we set `is_filtered = True`, not False. However, in this case, I think that maybe a different label name should have been used to minimize confusion.
1. if we are only filtering based on a simple flag, there is no need to read all the sections and iterate through them. Instead, this can be incorporated into the query for the sections, which should improve performance.
1. Are you sure that the unit test for this code works? There are three test sections in your file, only two of which are of type `move`, and both of them have `confirmed_mode` set. I would assume that all three of the sections would be filtered out by the `defaultQueryList`, and the test would fail.

CFC_WebApp/tests/TestFilterTrips.py
```diff
@@ -0,0 +1,99 @@
...
+    (resultPre, resultReg) = client.preRegister("this_is_the_super_secret_id", fakeEmail)
+    self.assertEqual(resultPre, 0)
+    self.assertEqual(resultReg, 1)
+
+    user = User.fromEmail(fakeEmail)
+    self.assertEqual(user.getFirstStudy(), 'testclient')
+
+    unclassifiedSections = tripManager.getUnclassifiedSections(User.fromEmail(fakeEmail).uuid)['sections']
+    # Check that of the valid sections in the testFilterFile (2/3), only one of them is returned by the query
+    self.assertEqual(len(unclassifiedSections), 1)
```
shankari: Are you sure that this works? There are three test sections in your file, only two of which are of type `move`, and both of them have `confirmed_mode` set. I would assume that all three of the sections would be filtered out by the `defaultQueryList`, and the test would fail

CFC_WebApp/tests/TestFilterTrips.py
```diff
@@ -0,0 +1,99 @@
...
+import sys
+import os
+from datetime import datetime, timedelta
+from dao.user import User
+from dao.client import Client
+import tests.common
+
+logging.basicConfig(level=logging.DEBUG)
+
+class TestFilterTrips(unittest.TestCase):
```
shankari: Why is this not a part of the existing TestTripManager unit test class? Will avoid copy/pasting a bunch of code in order to add one new test...

zacatac: I just thought that there was a lot of functionality already being tested in the TripManager so moving the testing of this feature to a different file would help seperate the testing functionality. If we add an __init__.py file in the test directory then we can avoid importing in each testing file.

shankari: Fair enough, although most of the stuff tested in TestTripManager is around retrieving unclassified sections, so this seems like an obvious addition there.

CFC_WebApp/tests/TestFilterTrips.py
zacatac: Yeah all of the tests passed before I made the request. I'll check again though/

CFC_WebApp/tests/TestFilterTrips.py
zacatac: Why don't I add the __init__.py file, and then move the filter tests back into the tripManager tests? That way we can avoid having to do all those imports in the future, but then have the tests in a more reasonable place.

CFC_WebApp/main/tripManager.py
zacatac: 1,2,3. Yeah there is a difference between the spec and the code. I will change the spec and then maybe change the new field name. I'm not sure what an appropriate name would be though, other than `keep` or something along those lines.

4. That is an interesting point, but the tests do pass. I will double check that though.

CFC_WebApp/tests/TestFilterTrips.py
shankari: Sounds good to me.

shankari:
So I think that the following tasks are pending:
- have the logic match the "filtered" field. An easy way to do this would be to rename "filtered" to "retained". This is not a technical requirement, but a maintainability requirement.
- filter by adding it to the query and not post-retrieval. This is a performance requirement.
- confirm that the test case passes and explain why it does
- (optional, but nice to have) merge `TestFilterTrips` into `TestTripManager`

zacatac:
What is the best way to make edits on this pull request? Or should I just create a new pull request with these changes included?

shankari:
Just push more commits to the branch that you created the pull request
from. They should automatically be appended to the request.

You *are* working off a branch, right?

Shankari

zacatac:
Haha yeah I am. I’ll push those before class tomorrow.

zacatac:
Updated with suggested changes!:

  • have the logic match the "filtered" field. An easy way to do this would be to rename "filtered" to "retained". This is not a technical requirement, but a maintainability requirement.
  • filter by adding it to the query and not post-retrieval. This is a performance requirement.
  • confirm that the test case passes and explain why it does
  • (optional, but nice to have) merge TestFilterTrips into TestTripManager

shankari:
Note to future self:

The reason that the TestTripManager test case passes in spite of the confirmed mode being non-empty in the test file is that we set the confirmed mode to empty in the `updateSections` method.

Merging the request now...

zacatac:
Woo! Thanks

PR #80, opened 2015-02-05
Verify jwt locally

jimmydief:


shankari:
@jesca and I discussed the testing for this. A simple test might be:
- get a valid JWT
- test1: use the client to verify it - verification should be successful
- test2: perturb the JWT (change some characters) - verification should be unsuccessful
- test3 (optional): create a mismatched JWT (validly signed but with a different client ID) and call the cfc_webapp method on it - verification should be unsuccessful. I am not sure 

Do you know how to generate a JWT to do test1 and test2? If not, I can send you a JWT from a throwaway google account for your testing (you don't want to use a real one!).

shankari:
@jesca, @jimmydief, @neerajbaid @abhinavasingh16, are you planning to test this? What's the ETA on the test code?

jimmydief:
@shankari just saw this, will talk to my team

shankari:
@jimmydief, no worries, not going to pull without seeing testing results anyway :)
Note that this week's deadline has been moved to Sunday so you have some additional time to finish working on this.

jimmydief:
@shankari Cool, thanks. Also, I'd be interested in that throwaway JWT

jimmydief:
@shankari We are having trouble importing the verifyUserToken function for testing since importing it runs the server. Is this a problem that has been addressed in the past? I'm not seeing any existing tests for verifyUserToken, or any other methods in that file actually.

jesca:
@shankari I think we just need to generate a JWT to test it -- right now, it looks like you are getting the the ID via a post request to /profile/create. We don't want to generate the JWT this way, I'm assuming.  What's a good way to generate a test token id?

shankari:
Sorry for the late response - I was on my way back home, and it takes me ~ 2 hours to get back home on the train.

@jimmydief, you are right that there are currently no unit tests for the api layer - it was intended to be a really simple shim layer that could be replaced easily if we chose to move to a different webserver. One possible solution might be to refactor the verifyAuth code into something in main (`main/auth.py` seems to be a pretty obvious choice), and the refactoring is a Good Thing in general.

@jesca, the JWTs are created on the phone. When I originally posted this, I thought that I would just generate a JWT from the phone code using a throwaway account and send it to you. But as I was trying to do that, I realized that the JWT is only valid for ~ 15 minutes, so this wouldn't work.

Let me think about the best way to do this...

shankari:
So it seems to me that there are actually two things that we need to test:

- unit test: whether the JWT code works at all, which you can test by constructing a JWT using the google OAuth python libraries (you can use a dummy payload) and then testing whether the payload is returned correctly when it is verified. This can be implemented as a unit test, but we largely expect it to pass, because the google OAuth libraries are unit tested. So this would just be making sure that you have imported everything correctly and are using the correct parameters and so on
- integration test: whether the code verifies JWTs actually generated on the phones, with the payload expected by google. The easiest way that I can think of to test this is to connect to your local server with your phone and attempt to log in. This is much harder to automate because it has two parts, and requires manipulating the UI on the phone, so I won't insist on automating it. You can just copy-paste results showing that this works.

In fact, if you show up in lab tomorrow, I'll walk you through the steps and help you connect your phone to your server and finish the integration testing...

jesca:
@shankari i'll be there

shankari:
Resubmitted as 0cd868a88361c7d02667fced624feba1641bff33 and merged.

PR #85, opened 2015-02-09
Mini-Project 1

sdsingh:
Cached reverse_geocode + testing

### DIFF
--- a/CFC_WebApp/main/zipcode.py
+++ b/CFC_WebApp/main/zipcode.py
@@ -9,7 +9,6 @@ from collections import defaultdict
 from get_database import get_section_db,get_profile_db
 from modeshare import get_user_mode_share_by_distance
 
-
 # zipcode_list = ["other", "94720", "94709", "94705", "94706", "94703", "94704"]
 
 def getDistinctUserCount():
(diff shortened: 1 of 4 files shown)


CFC_WebApp/main/zipcode.py
```diff
@@ -17,18 +16,26 @@ def getDistinctUserCount():
...
     logging.debug("Found %s distinct users " % distinctUserCount)
     return distinctUserCount
+   
 
-def get_userZipcode(user):
+def get_userZipcode(user, valid_zip):
     location = detect_home_from_db(user)
-    if location!='N/A':
+    current_db = get_profile_db()
+    user_pro = current_db.find_one({"$and":[{'source':'Shankari'},{'user_id':user}]})
```
shankari: Would have been nice to use the User DAO, which didn't exist at the time this code was written, but was added later. That has a built-in method to get the profile:
User.from_uuid(user).getProfile()

CFC_WebApp/tests/TestProfile.py
```diff
@@ -0,0 +1,86 @@
...
+            self.assertEquals(func1.called, True)
+            self.assertEquals(func2.called, True)
+            func1.reset_mock()
+            func2.reset_mock()
+            prof_1 = self.Profiles.find_one({'user_id':'1'})
+            self.assertEquals(prof_1['zip'],  '94709')
+            #zip should be stored, no API call
+            Profile.update_profiles(True)
+            prof_1 = self.Profiles.find_one({'user_id':'1'})
+            self.assertEquals(prof_1['zip'],  '94709')
```
shankari: Shouldn't you check here that the API call is not performed? Shouldn't there be an assertEquals for it?

CFC_WebApp/tests/TestProfile.py
```diff
@@ -0,0 +1,86 @@
...
+            #new zip retrieval phase, should make API call and obtain zipcode
+            Profile.update_profiles(True)
+            self.assertEquals(func1.called, True)
+            func1.reset_mock()
+            prof_1 = self.Profiles.find_one({'user_id':'1'})
+            self.assertEquals(prof_1['zip'],  '94720')
+            #zip should be stored, no API call
+            Profile.update_profiles(True)
+            prof_1 = self.Profiles.find_one({'user_id':'1'})
+            self.assertEquals(prof_1['zip'],  '94720')
```
shankari: Ditto.

PR #90, opened 2015-02-17
Outlier detection

sdsingh:


CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -163,51 +202,69 @@ def getGroundTruthMode(self, section):
...
-    
-    featureMatrix[i, 19] = mode_start_end_coverage(section, self.bus_cluster,105)
-    featureMatrix[i, 20] = mode_start_end_coverage(section, self.train_cluster,600)
+    points = section["track_points"]
+    if len(points) > 4:
+        points = np.array([[(datetime.datetime.strptime(point['time'].split('-')[0], "%Y%m%dT%H%M%S") - datetime.datetime(1970,1,1)).total_seconds(), 
+            point["track_location"]["coordinates"][0], point["track_location"]["coordinates"][1]] for point in points])
+        time_stamp = points[:,0].reshape(len(points[:,0]), 1)
+        lon_lat = points[:,1:]
+        model_ransac = linear_model.RANSACRegressor(linear_model.LinearRegression())
```
shankari: Can you provide some intuition around why we would want to put the RANSAC regression code into the machine learning pipeline?

This seems to have two problems:

1.  It seems to me that you would be running RANSAC over and over on the same set of trips. For example, consider the scenario in which I took 3 trips (t1, t2, t3) today and classified them today.
  - Then, (t1, t2 and t3) would become part of the training set tomorrow. So I would run RANSAC on them then.
  - Then, they would continue to be part of the training set day after tomorrow, so I would run RANSAC on them again.
2. Running RANSAC at this point will leave the values in the database untouched. This will mean that the users will see the zigzag points as well. Having zigzag, unsmoothed points in the map displayed to the user makes it harder for users to classify trips, and to have faith in our data collection.

Running RANSAC at the point that the data comes in from the phone (`tripManager.storeSensedTrips`) seems like it would avoid both of these problems, which is why I suggested it in the issue description.

CFC_DataCollector/modeinfer/experiment.py
```diff
@@ -0,0 +1,66 @@
+import scipy.io
+import random
+import numpy as np
+from sklearn.cross_validation import KFold
+
+mat = scipy.io.loadmat('original_data.mat')
```
shankari: what is `original_data.mat`? How is it generated?

CFC_DataCollector/modeinfer/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+    #model = buildModelStep(orig_training_set, orig_training_labels)
+    model = buildModelStep(mat['X'][train], mat['y'][train])
+    #new_model = buildModelStep(mat['X'], mat['y'].ravel())
+    
+    #predicted_labels = model.predict(orig_test_set)
+    predicted_labels = model.predict(mat['X'][test])
+    print(predicted_labels)
+    print(orig_test_labels)
+    
+    def score_classifier(predicted_labels, true_labels):
```
shankari: This is not a great comparison because a high accuracy on the `walking` mode, for example, can pull up the overall accuracy and mask the really low accuracy of the motorized modes. That is why we used stratified k-fold validation for our estimation.

I have an ipython file that implements stratified k-fold validation and generates a confusion matrix. Let me see if I can find it...

CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -33,30 +34,44 @@ def __init__(self):
...
   def runPipeline(self):
     allConfirmedTripsQuery = ModeInferencePipeline.getSectionQueryWithGroundTruth({'$ne': ''})
+
+
+
     (self.modeList, self.confirmedSections) = self.loadTrainingDataStep(allConfirmedTripsQuery)
     logging.debug("confirmedSections.count() = %s" % (self.confirmedSections.count()))
     
+    #commented out for now, because the training set is not large enough and there is some
+    #issue with this query, it doesn't load any sections for me
```
shankari: The issue is probably that there is no Backup_database in your local mongodb. It's fine to comment it out for now...

CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -83,20 +98,42 @@ def getSectionQueryWithGroundTruth(groundTruthMode):
...
   def loadTrainingDataStep(self, sectionQuery, sectionDb = None):
+    logging.debug("START TRAINING DATA STEP")
     if (sectionDb == None):
       sectionDb = self.Sections
+    begin = time.time()
     logging.debug("Section data set size = %s" % sectionDb.find({'type': 'move'}).count())
+    duration = time.time() - begin
+    logging.debug("Getting dataset size took %s" % (duration))
+        
+    logging.debug("Querying confirmedSections %s" % (datetime.datetime.now()))
```
shankari: Please don't put the datetimes at the end of the log statements. The log statements are already timestamped, so this is a bunch of unnecessary system calls. It is true that the calls are pretty lightweight, but why do more work than we need to. if your log statements don't print the time, you can look into the log format that is getting invoked.

CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -122,13 +159,15 @@ def generateFeatureMatrixAndResultVectorStep(self):
 
       # This will crash the script because we will try to access a record that
       # doesn't exist.
-
       # So we limit the records to the size of the matrix that we have created
       for (i, section) in enumerate(self.confirmedSections.limit(featureMatrix.shape[0]).batch_size(300)):
-        self.updateFeatureMatrixRowWithSection(featureMatrix, i, section)
+        #self.updateFeatureMatrixRowWithSection(featureMatrix, i, section)
+        featureMatrix[i] = self.updateFeatureMatrixRowWithSection(featureMatrix, i, section)
```
shankari: Why return the row instead of updating the feature matrix directly? I don't have a strong reason why we need to update as a side effect, but am just curious about the reason for the change...

CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -122,13 +159,15 @@ def generateFeatureMatrixAndResultVectorStep(self):
...
         resultVector[i] = self.getGroundTruthMode(section)
         if i % 100 == 0:
             logging.debug("Processing record %s " % i)
+      featureMatrix = np.nan_to_num(featureMatrix)
```
shankari: Was is the nan_to_num call here? I had already checked that fix as part of commit 990ea697b363dabcd4493858b4eb6a76929a79d1

Also, `updateFeatureMatrixRowWithSection` already calls np.nan_to_num() in the return statement, so won't this make the call twice?

data_processing.py
```diff
@@ -0,0 +1,76 @@
...
+What we are interested in is called SECTIONS: components of trips
+corresponsing to one type of transportation. Thus, we need 
+to run RANSAC on each of these.
+'''
+path = "our_collection_data/android/after_elapsed_time_fix/"
+#processed_data = np.zeros((1,2))
+total_data = []
+for path, direc, filenames in os.walk(path):
+    for filename in filenames:
+        if filename == "1422220481.05":
```
shankari: Why hardcoded value?

data_processing.py
```diff
@@ -0,0 +1,76 @@
+import os
```
shankari: Maybe add a short comment describing what this script does? `data_processing` is not quite as descriptive as `check_increasing_timestamp`.

CFC_DataCollector/modeinfer/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+    		else:
+                        wrong += 1
+    		c += 1
+    	return (correct, wrong)
+    
+    correct, wrong = score_classifier(predicted_labels, mat['y'][test])
+    #correct, wrong = score_classifier(predicted_labels, orig_test_labels)
+    print("Your classifier got " + str(correct) + " correct labels.")
+    print("Your classifier got " + str(wrong) + " wrong labels.")
+    print("Total accuracy: " + str((float(correct)/float(correct+wrong))*100.0))
```
shankari: What were the results? Add them to the pull request or to the issue for the record?

check_increasing_timestamp.py
```diff
@@ -0,0 +1,33 @@
+import os
```
shankari: Put into a util or bin directory in CFC_DataCollector to avoid cluttering up the main directory?

CFC_DataCollector/modeinfer/experiment.py
sdsingh: From the given dataset (about 130 points), with 5-fold cross-validation, the average accuracy was around 75% pre-RANSAC. The addition of RANSAC brought us to around 78%. However, I think we may do better in practice, since some features seem to be calculated before the ML pipelining step (e.g. first filter mode), but the impact of these is hard to evaluate in the given context, as we had processed data and did not run the "first filter".

shankari: Hm, so it looks like it didn't help very much but it didn't hurt either. This could be because the underlying learning was not really affected as strongly by the outliers, because our features are relatively coarse.

I would not hold up the pull request for this, but you might want to consider alternate evaluation methods. The path filter paper had some examples, such as the entropy of the points, that could provide a better measure of the improvement that you are bringing to the table.

CFC_DataCollector/modeinfer/pipeline.py
```diff
@@ -163,51 +202,69 @@ def getGroundTruthMode(self, section):
...
+            point["track_location"]["coordinates"][0], point["track_location"]["coordinates"][1]] for point in points])
+        time_stamp = points[:,0].reshape(len(points[:,0]), 1)
+        lon_lat = points[:,1:]
+        model_ransac = linear_model.RANSACRegressor(linear_model.LinearRegression())
+        model_ransac.fit(lon_lat, time_stamp)
+        inlier_mask = model_ransac.inlier_mask_
+        outlier_mask = np.logical_not(inlier_mask)
+        print "total size: " + str(len(outlier_mask))
+        remove = [index for index,v in enumerate(outlier_mask) if v]
+        section["track_points"] = [v for j,v in enumerate(section["track_points"]) if j not in frozenset(remove)]
```
shankari: One more comment: We should retain the initial (unsmoothed) data points separately to make it easier to explore alternate approaches. Two potential solutions:
- stick them into a separate "unsmoothed" collection, indexed by a shared trip + section ID
- stick them into a separate field of the same collection

           section['unsmoothed_track_points'] = section['track_points']
           section['track_points'] = [v for j,v in enumerate(section["track_points"]) if j not in frozenset(remove)]

CFC_DataCollector/modeinfer/experiment.py
sdsingh: cleaned_data_results
original_data_results
First is the confusion matrix for the cleaned data, then the confusion matrix for the original. Although this does show some improvement, I think more importantly this shows that we are not harming accuracy while visual inspection shows significant improvement in the quality of our trip paths.

shankari: I'll buy that. How about including some visual inspection results as well?

shankari:
Is this ready for me to review a second time? Have all my original comments been addressed?

sdsingh:
I think the main code is ready for review, I'm going to pull some of my
auxiliary scripts out of the commit though

CFC_DataCollector/moves/collect.py
```diff
@@ -87,6 +87,7 @@ def fillSectionWithMovesData(sec_from_moves, newSec):
    newSec['duration'] = sec_from_moves["duration"] if "duration" in sec_from_moves else None
    newSec['distance'] = sec_from_moves["distance"] if "distance" in sec_from_moves else None
    newSec['track_points'] = [{'track_location':{'type':'Point', 'coordinates':[point["lon"],point["lat"]]}, 'time':point["time"]}for point in sec_from_moves["trackPoints"]] if "trackPoints" in sec_from_moves else []
+   newSec['cleaned_points'] = _cleanGPSData(newSec['track_points'])
```
shankari: This won't work, because the rest of the code (ML pipeline, client map display...) all use `track_points`. So this will include cleaned points, but won't affect anything else.

In order to really fix this, you need to:

1. change all the uses of `track_points` to use `cleaned_points` instead, OR
2. change this so that the *cleaned* points are stored in `track_points` and the original points are stored in `original_points` or something.

This will ensure that all the rest of the code works with the cleaned points, but we still have the original points around for further analysis.

sdsingh: Fixed. Sorry about that :/

shankari:
Final cleanup: remove `data_processing.py`, or add a description on what it does and move it into `CFC_WebApp/utils`.

I think that once that is done, I can merge this pull request and you can move on to the project for the next week.

PR #92, opened 2015-02-19
Issue 91 (server-side) completed: Added user's mpg and footprint calculation

shanthistyle:
Includes tests for the storing of the list of mpgs and recalculation of the carbon footprint

### DIFF
--- a/CFC_WebApp/api/cfc_webapp.py
+++ b/CFC_WebApp/api/cfc_webapp.py
@@ -244,15 +244,8 @@ def createUserProfile():
 @post('/profile/update')
 def updateUserProfile():
   logging.debug("Called updateUserProfile")
-  userToken = request.json['user']
+  user_uuid = getUUID(request)
   mpg_array = request.json['mpg_array']
-  # This is the only place we should use the email, since we may not have a
-  # UUID yet. All others should only use the UUID.
-  userEmail = verifyUserToken(userToken)
-  logging.debug("userEmail = %s" % userEmail)
-  user = User.register(userEmail)
-  logging.debug("Looked up user = %s" % user)
-  logging.debug("Returning result %s" % {'uuid': str(user.uuid)})
   return user.setMpgArray(mpg_array)
 
 
--- a/CFC_WebApp/api/cfc_webapp.py
+++ b/CFC_WebApp/api/cfc_webapp.py
@@ -245,6 +245,7 @@ def createUserProfile():
 def updateUserProfile():
   logging.debug("Called updateUserProfile")
   user_uuid = getUUID(request)
+  user = User.fromUUID(user_uuid)
   mpg_array = request.json['mpg_array']
   return user.setMpgArray(mpg_array)
 
--- a/CFC_WebApp/api/cfc_webapp.py
+++ b/CFC_WebApp/api/cfc_webapp.py
@@ -241,6 +241,21 @@ def createUserProfile():
   logging.debug("Returning result %s" % {'uuid': str(user.uuid)})
   return {'uuid': str(user.uuid)}
 
+@post('/profile/update')
+def updateUserProfile():
+  logging.debug("Called updateUserProfile")
+  userToken = request.json['user']
+  mpg_array = request.json['mpg_array']
+  # This is the only place we should use the email, since we may not have a
+  # UUID yet. All others should only use the UUID.
+  userEmail = verifyUserToken(userToken)
+  logging.debug("userEmail = %s" % userEmail)
+  user = User.register(userEmail)
+  logging.debug("Looked up user = %s" % user)
+  logging.debug("Returning result %s" % {'uuid': str(user.uuid)})
+  return user.setMpgArray(mpg_array)
+
+
 @post('/profile/consent')
 def setConsentInProfile():
   user_uuid = getUUID(request)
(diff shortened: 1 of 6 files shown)


shankari:
So far looks good. I want to review the conversion calculations more carefully tomorrow, but I don't see any obvious issues that need to be addressed.

shankari:
One high level reminder/nit: You should really be working off a branch.

github basically assumes one change per branch, so if you get review feedback that you need to fix, you can make changes in the same branch and push them to the same pull request.

CFC_WebApp/api/cfc_webapp.py
```diff
@@ -241,6 +241,21 @@ def createUserProfile():
...
+@post('/profile/update')
+def updateUserProfile():
+  logging.debug("Called updateUserProfile")
+  userToken = request.json['user']
+  mpg_array = request.json['mpg_array']
+  # This is the only place we should use the email, since we may not have a
+  # UUID yet. All others should only use the UUID.
+  userEmail = verifyUserToken(userToken)
+  logging.debug("userEmail = %s" % userEmail)
+  user = User.register(userEmail)
```
shankari: Why do you register the user again here? Shouldn't we be registering in `/profile/create`, and updating the profile in `/profile/update`?

CFC_WebApp/tests/TestCarbon.py
```diff
@@ -249,6 +249,13 @@ def testGetCarbonFootprintsForMap(self):
     footprintMap = carbon.getCarbonFootprintsForMap(testDistanceMap, testModeFootprintMap)
     self.assertEqual(footprintMap, {'a': 0.001, 'b': 0.004, 'c': 0.009})
 
+  def testGetCarbonFootprintsForMap(self):
+    testDistanceMap = {'a': 1, 'b': 2, 'c': 3}
+    testModeFootprintMap = {'a': 1, 'b': 2, 'c': 3}
+
+    footprintMap = carbon.getCarbonFootprintsForMap(testDistanceMap, testModeFootprintMap)
```
shankari: Isn't this the same as the previous test? Shouldn't you be making a call to user instead?

shankari:
@shanthistyle, conversion calculations look fine.
So for me to pull the changes, you need to address the other two comments:

- registering the user versus updating an existing profile.
- duplicating a test case

shanthistyle:
Hi Shankari,

Thank you for reviewing the code :) I will address these changes and submit another pull request tomorrow.

Shanthi



—

jimmydief:
@shanthistyle @shankari I think it would make more sense if the server received the specific vehicle ID(s) and made the call to the FuelEconomy.gov MPG API. Since this is shared behavior between the iOS and Android apps and does not require user input, it would make sense to make that call on the server. The server could return the MPG value in its response if we want to display that value to the user.

Also, does it make sense to take in an array of vehicle IDs / MPGs if we're not asking the user which vehicle they drove for each trip? This wouldn't require any change, just wondering if that's future functionality that should be added.

shankari:
@jimmydief, would that really help you though? Wouldn't you still need to access the fueleconomy.gov MPG API to get the vehicle ID (through the server proxy as we discussed)?

jimmydief:
The user will still have to select the year, make, model, and option for their vehicle, but it makes more sense to me to do the query for MPG on the serverside for the reasons stated above. It's one less thing to repeat on both devices and would be just a few lines with the XML processing stuff that Josh provided us.

    import xmltodict

    def test_xml_to_json():
        price = urllib2.urlopen('http://www.fueleconomy.gov/ws/rest/fuelprices')
        xml = price.read()
        print xml_to_json(xml)

    def xml_to_json(xml_string):
        temp = xmltodict.parse(xml_string)
        return json.dumps(temp)

shankari:
@jimmydief, I have no objections to making that call from the server. But just so that I understand the proposed workflow:

- you will make a call to EPA and retrieve the list of years, which you will use to populate a spinner
- the user will select a year
- you will make a call to the EPA and retrieve a list of makes, which you will use to populate a spinner
- the user will select a make
- you will make a call to the EPA and retrieve a list of models, which you will use to populate a spinner
- the user will select a model
- you will make a call to the EPA and retrieve the vehicle ID, which you will send to the server

Is that the workflow that you had in mind? Because that's not how I understood Prof. Culler's suggestion.

shankari:
@jimmydief, @joshzarrabi, @shanthistyle, @Wuhandson, to elaborate further now that I am home, I believe that Prof. Culler was suggesting the following.
We write an XML -> JSON converter proxy on the server. In other words, we make an REST API call with a URL that is something like:
`asJSON/<originalXMLWebserviceURL>`

The implementation of that API call is to call Josh's code and return the corresponding JSON for the XML. Then, on the phone, we have the following workflow:

* make a call to our asJSON service with the EPA URL to retrieve the list of years
* the proxy makes a call to EPA, formats the result as JSON and returns it
* you parse the JSON code with a single line and populate the spinners
* the user selects a year
* you make a call to our asJSON service with the EPA URL to retrieve the list of makes
* the proxy makes a call to EPA, formats the result as JSON and returns it

and so on. We can then reuse this solution in the future if we need to integrate with other XML based services.

I think that this is a really elegant solution - Prof. Culler is very smart!

In order to get a broader context for this, you might want to re-read the REST API reading and the use of proxies for changing the presentation of data.

jimmydief:
Shankari, your second post is what I have in mind. Also, I've switched from using the UIPickerView "spinner" to simple TableViews which make a lot of sense.

shankari:
@jimmydief, good. You are smart too :)

Shankari

On Fri, Apr 3, 2015 at 1:16 PM, James Diefenderfer <notifications@github.com

PR #95, opened 2015-02-21
added oauth to verify jwt before falling back to api call

jesca:
Uses client-side library to verify JWT instead of making an API call using the OAuth2.0Client library. 

Tested and verify that everything is working correctly. 
Tested using wrong client_key for the OAuthLibrary -- defaulted to API call. 
Tested with right client_key -- didn't make the API call. The token was verified using Oauth library. 

Note: added to requirements.txt (need to run pip install -r requirements.txt) to use the library. 
@jimmydief

### DIFF
--- a/CFC_WebApp/api/cfc_webapp.py
+++ b/CFC_WebApp/api/cfc_webapp.py
@@ -10,10 +10,12 @@ from datetime import datetime
 import time
 # So that we can set the socket timeout
 import socket
-# For decoding tokens using the google decode URL
-# We want to switch this to something offline later
+# For decoding JWTs using the google decode URL
 import urllib
 import requests
+# For decoding JWTs on the client side
+import oauth2client.client
+from oauth2client.crypt import AppIdentityError
 
 config_file = open('config.json')
 config_data = json.load(config_file)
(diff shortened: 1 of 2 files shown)


shankari:
So how did you know that the API call was made or not made?

jesca:
This debug statement: logging.debug("OAuth failed to verify id token, falling back to constructedURL")
The log appeared only when the client side verification failed.

CFC_WebApp/api/cfc_webapp.py
```diff
@@ -351,15 +353,24 @@ def after_request():
...
-  if (in_client_key != client_key):
-    if (in_client_key != hack_client_key):
-      abort(401, "Invalid client key %s" % in_client_key)
-  logging.debug("Found user email %s" % tokenFields['email'])
-  return tokenFields['email']
+    try:
+        # attempt to validate token on the client-side
+        logging.debug("Using OAuth2Client to verify id token")
+        tokenFields = oauth2client.client.verify_id_token(token,client_key)
+        print tokenFields
```
shankari: This print is not really necessary and should really be removed because it will clutter up the console. However, this change has already taken so long to get through that I won't insist on it. I'll remove it myself before I deploy the changes on the real server.

PR #99, opened 2015-03-02
Alternative api endpoint for /compare that utilizes get request with custom auth header

zacatac:
Used for making caching of the results summary page possible.

### DIFF
--- a/CFC_WebApp/api/cfc_webapp.py
+++ b/CFC_WebApp/api/cfc_webapp.py
@@ -34,7 +34,7 @@ hack_client_key = key_data["ios_client_key"]
 BaseRequest.MEMFILE_MAX = 1024 * 1024 * 1024 # Allow the request size to be 1G
 # to accomodate large section sizes
 
-skipAuth = True
+skipAuth = False
 
 app = app()
 
--- a/CFC_WebApp/api/cfc_webapp.py
+++ b/CFC_WebApp/api/cfc_webapp.py
@@ -34,7 +34,7 @@ hack_client_key = key_data["ios_client_key"]
 BaseRequest.MEMFILE_MAX = 1024 * 1024 * 1024 # Allow the request size to be 1G
 # to accomodate large section sizes
 
-skipAuth = False
+skipAuth = True
 
 app = app()
 
@@ -262,16 +262,18 @@ def setStats():
   inStats = request.json['stats']
   stats.setClientMeasurements(user_uuid, inStats)
 
-@post('/compare')
-# @get('/compare')
+# @post('/compare')
+@get('/compare')
 def getCarbonCompare():
+  for key, val in request.headers.items():
+    print("  %s: %s" % (key, val))
+  # token = request.headers.get('Authorization')
+  # print token
+
   from clients.default import default
-
-  if request.json == None:
-    return "Waiting for user data to become available..."
-
-  if 'user' not in request.json:
-    return "Waiting for user data to be become available.."
+  
+  # if token is None or 'Bearer' not in token:
+  #   return "Waiting for user data to become available..."
 
   user_uuid = getUUID(request)
   clientResult = userclient.getClientSpecificResult(user_uuid)
@@ -382,7 +384,8 @@ def getUUID(request):
     retUUID = user_uuid
     logging.debug("skipAuth = %s, returning fake UUID %s" % (skipAuth, user_uuid))
   else:
-    userToken = request.json['user']
+    userToken = request.headers.get('Authorization').split()[1]
+    # userToken = request.json['user']
     retUUID = getUUIDFromToken(userToken)
   request.params.user_uuid = retUUID
   return retUUID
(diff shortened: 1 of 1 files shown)


CFC_WebApp/api/cfc_webapp.py
```diff
@@ -282,6 +281,26 @@ def getCarbonCompare():
     logging.debug("No overriding client result for user %s, returning default" % user_uuid)
   return default.getResult(user_uuid)
 
+@get('/compare')
```
shankari: Have you actually tested this with both @get and @post endpoints and ensured that it works? Can you update the testing done section with detailed tests.

shankari:
this is ok to test

zacatac:
I’ll write some tests, for these endpoints. I’ll take a look at the tests already written for the post request, and then model my additional tests off of that.

shankari:
I don't know that there are any tests written for the post request.
There are tests for the carbon footprint calculation, but no API tests.
It would be great if you could write some, but I won't insist on them given the rest of your workload.
A clearly documented step by step manual test will suffice.

For example, you could build a client with your changes and a client from the main repo (without your changes) and ensure that they both worked.

zacatac:
Sounds good.

CFC_WebApp/api/cfc_webapp.py
```diff
@@ -282,6 +281,26 @@ def getCarbonCompare():
...
+def getCarbonCompare():
+  for key, val in request.headers.items():
+    print("  %s: %s" % (key, val))
+
+  from clients.default import default
+  
+  if 'User' not in request.headers or request.headers.get('User') == '':
```
shankari: lowercase "user" instead of "User" to be consistent with the POST messages?

CFC_WebApp/api/cfc_webapp.py
```diff
@@ -382,7 +401,10 @@ def getUUID(request):
     retUUID = user_uuid
     logging.debug("skipAuth = %s, returning fake UUID %s" % (skipAuth, user_uuid))
   else:
-    userToken = request.json['user']
+    if inHeader:
+      userToken = request.headers.get('Authorization').split()[1]
```
shankari: Is the header called "Authorization" or "User"? It looks like you assume it to be "User" in `getCarbonCompare()` and in the client code, but you assume it is "Authorization" here.

CFC_WebApp/api/cfc_webapp.py
zacatac: Sure. I'll update it.

CFC_WebApp/api/cfc_webapp.py
zacatac: Woops. It is 'User'. So I will need to change this, and the phone side code to reflect the change from 'user' to 'User'.

shankari:
@zacatac, is there an ETA on finishing the testing to ensure that this endpoint is backwards compatible? Once that is done, I will deploy both your and Shaun's changes to the server.

Then, we can get one round of testing over https against the real server before merging the phone side changes. @neerajbaid has tested with the changes on localhost, but would like to run the more comprehensive tests before updating the phone pull request.

zacatac:
Yeah let me finish that up tomorrow, and then I'll make a pull request by tomorrow evening.

I agree that testing on the real server is a good idea, just to make sure the new Android code that communicates with the server actually works.

zacatac:
# Testing strategy
## To test the old api 

1. Clone the [most recent e-mission-phone commit that does not have the new api call](https://github.com/e-mission/e-mission-phone/commit/6398f3f8f5131a02e62f1b1d97aa45d59824252f)
2. Clone the [most recent e-mission-server commit that has both api calls for /compare](https://github.com/zacatac/e-mission-server/commit/07ac9a4ca7adb32ab19eace0d63591cc2f045365) (should update this link when this request is merged)
3. Run the server on 0.0.0.0:8080 to make sure it can communicate externally
4. Build one of the phone apps. This should cause the results summary screen to open automatically in a simulator or hardware device. 
5. Observe in the server terminal window that a GET request has been made to the /compare endpoint.
6. Reload the results summary screen
7. Observe in the server terminal window that a second GET request has been made to the /compare endpoint.

## To test the new api

1. Clone the [most recent e-mission-phone commit that has the new api call](https://github.com/jesca/e-mission-phone/commit/9cca71125bde844fd61b47303c78c5dfd6fddc05)
2. Clone the [most recent e-mission-server commit that has both api calls for /compare](https://github.com/zacatac/e-mission-server/commit/07ac9a4ca7adb32ab19eace0d63591cc2f045365) 
3. Run the server on 0.0.0.0:8080 to make sure it can communicate externally
4. Build one of the phone apps. This should cause the results summary screen to open automatically in a simulator or hardware device. 
5. Observe in the server terminal window that a POST request has been made to the /compare endpoint.
6. Reload the results summary screen
7. Observe in the server terminal window that a second POST request to the /compare endpoint is NOT made.

Notes:
* Only tested on Android, but this should be sufficient. 
* We could work on writing E2E tests in the future to automate this. I attempted to do this, but it is a big pain.

shankari:
> We could work on writing E2E tests in the future to automate this. I attempted to do this, but it is a big pain.

Agreed that we should not hold up this change for E2E tests. We need to do that as a separate project.

shankari:
@zacatac, a quick clarification. You actually finished  the testing, right?

zacatac:
Yes



—

shankari:
Great!

PR #101, opened 2015-03-05
Outlier detection--Passes all tests

sdsingh:
Fixed small bugs, incorporated new changes to main repo.


shankari:
Comment from Shaun over email:

> Here is the original file visualized. There are actually only 7 points visible, if I'm seeing this correctly, despite 10 being present in the actual code.
> I think this is good justification for the cleaned version only containing 7 points which correspond to these visible points (I don't have a visualization yet, since I'd have to rewrite the script to deal with processed data, but I checked each point by hand).

shankari:
This is correct. I checked the raw file, and the last 4 points are identical.

    {"time":"20150102T000252-0500","lat":37.385461,"lon":-122.078265,"accuracy":3.700000047683716},
    {"time":"20150102T000212-0500","lat":37.385461,"lon":-122.078265,"accuracy":3.700000047683716},
    {"time":"20150102T000232-0500","lat":37.385461,"lon":-122.078265,"accuracy":3.700000047683716},
    {"time":"20150102T000252-0500","lat":37.385461,"lon":-122.078265,"accuracy":3.700000047683716}

This is probably because I got this data from a unit test run, in which the last n points are duplicated to test the end of a trip. It does seem reasonable that the linear model will combine those points into one.

Please push a change to the test and I will merge once the automatic tests pass.

PR #106, opened 2015-03-30
changed profile.py and others to be consistent with lon, lat format

Mogeng:


shankari:
jenkins, test this please

shankari:
Mogeng deleted this fork in order to clean up his code.

PR #110, opened 2015-04-07
XML to JSON Proxy

jimmydief:
URL parameter must be UTF-8 encoded to work correctly.

screen shot 2015-04-06 at 5 15 11 pm

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ six==1.9.0
 utm==0.3.1
 wsgiref==0.1.2
 google-api-python-client==1.1
+xmltodict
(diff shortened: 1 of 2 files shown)


jimmydief:
@shankari @Wuhandson here's the proxy. Super simple and should simplify things a lot.

requirements.txt
```diff
@@ -7,4 +7,7 @@ requests==2.5.1
 six==1.9.0
 utm==0.3.1
 wsgiref==0.1.2
-google-api-python-client==1.1
\ No newline at end of file
+google-api-python-client==1.1
+jsonpickle
```
shankari: nit: This change does not require jsonpickle or python-crontab.
Also, don't you need to add urllib2?

shankari:
One minor nit: otherwise looks good.

requirements.txt
jimmydief: @shankari good catch

PR #111, opened 2015-04-08
[WIP] Recommender changes

jimmydief:
Making another PR for comments on further changes to the recommender system.


CFC_WebApp/clients/recommendation/result_template.html
```diff
@@ -0,0 +1,107 @@
...
+        width: 100%;
+      }
+    </style>
+  <body ng-controller="RootCtrl" hide-nav-bar="true">
+
+    <div id="rec-info">
+      <h1>Recommended Trip</h1>
+      <p>Leave at <span id="departure-time">1:30 PM</span> and arrive at <span id="arrival-time">1:56 PM</span></p>
+      <div id="rec-reasons">
+        <p id="time-rec">Recommended because this trip is 4 minutes faster than your original trip.</p>
```
shankari: @jimmydief, this needs to be for the

CFC_DataCollector/recommender/recommendation_pipeline.py
```diff
@@ -14,26 +15,24 @@ def retrieve_alternative_trips(self, trip_list):
         return []
 
     def get_selected_user_utility_model(self, user_id):
-        return []
+        return UserUtilityModel.find_from_db(user_id)
 
     def recommend_trips(self, trip_id, utility_model):
         return []
```
shankari: @sdsingh, where are recommendations being generated?

sdsingh: Thanks to a silly hack from demo time, recommendations are being generated
in utility model pipeline.. I'm moving the code over to recommendation
pipeline however.

Do you remember our discussion a while ago on implementing the user model?
The issue was that alternatives were not attached to their original trips
in any way, and so the model could learn crazy things. What was the final
solution?

shankari: Sorry, I am not sure I remember the discussion.
How about something naive like adding an "original_trip" field to each alternative?
It should be pretty easy to then query the alternatives db for all alternatives for an original trip, and to lookup the original trip from the alternatives.

sdsingh: I ask because I'm rewriting all of our modelling to use numpy, and so this
is a good time to fix things.

Also, Shankari, do you think we could just wait until Thursday to deploy
onto production? I know that pushing back is not good with such little time
left, but there are going to be some tricky things with deploying. The
current code is fairly inefficient, and so we will need to start collecting
alternatives in batches. In addition, since Josh's code was deleted, the
pipeline will only be generating google maps alternatives right now (he
didn't push a working version last night) and then it will be hard to
convert our db to otp.

I think it would be easiest to do a full end-to-end demo of recommendations
from the baby server on Thursday, and then work on moving to production.

Again, I'm sorry to delay, but since bits and pieces are not ready to go
(notably otp and trip clustering/canonical trips), I think we should wait
on deploying into production.

sdsingh: Wait, that's not the issue. At query time, trips and alternatives are
linked. The issue is when we throw it all into the model building phase

shankari: What specifically is the issue in the model building phase?

shankari:
> I think it would be easiest to do a full end-to-end demo of recommendations
> from the baby server on Thursday, and then work on moving to production.

How concretely would we do an end to end demo from the baby server? I don't think anybody's phones are hooked up to the baby server. I guess somebody could hook their phone up to the baby server and show their recommendation.

CFC_DataCollector/recommender/recommendation_pipeline.py
sdsingh: To build the user model, we obtain all trips and their alternatives. If we
fit on all of them together, we can learn bad things. This is what we have,
roughly:
Data:
[  (trip, [alternatives]), (trip2, [alternatives2]), .... ]
So if we fit on all of these together, we have a problem, because we could
be learning things like: "User chose alternative2-walking instead if
trip1because..."

sdsingh:
That is an issue, but Jimmy and I could at least connect our phones.

CFC_DataCollector/recommender/recommendation_pipeline.py
shankari: Ah, the solution for that was not to not use scikit-learn logistic regression, but implement the maximum likelihood optimization manually. In that, I think that you optimize for each [trip,{alternatives}] pair and then multiply the probabilities together (see conversion of eqn 3 to eqn 5 in the Vij/Shankari paper).

Vij sent me his code to do this, and I can send it to you.

But I suspect that we will get similar results by using scikit-learn logistic regression for each trip,{alternatives} set and then multiplying the probabilities together, or at least, good enough results for the first pass. scikit-learn gives you the probabilities.
http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression.predict_proba

There is also a package that Sid is learning in his discrete choice class that he can show you how to use if you want something more complex.

shankari: Actually, I am not sure that makes sense. I am pretty sure that Robert said that the answer was to multiply the probabilities together, but I don't think that we want to multiply the probabilities of the parameters together.

shankari:
Fair enough. What do you think is a possible timeline for production?

CFC_DataCollector/recommender/recommendation_pipeline.py
shankari: Let me send you Vij's code and we can see how easy it is to use.

sdsingh: Ok, thanks. Multiplying probabilities doesn't quite make sense with our
model... Can you send me vijs? Thanks. As far as production, early next
week (Monday?) should be fine, especially if other students are available
to help.

shankari: I did. Today morning. I forwarded his email so maybe it looks like it came
from him?

Shankari

sdsingh: Got it. Thanks!

shankari:
jenkins, please test this

shankari:
retest this please

shankari:
Now the import errors are gone, but the count is back to 15.

shankari:
Looks like 5 is more stable than 15.
Merging the pull request now.
No time for code review because of deadline pressure.

CFC_DataCollector/recommender/alternative_trips_module.py
```diff
@@ -22,40 +24,29 @@
...
+    total_stagger = 0
+    for existing_trip in user_trips:
+        #if not existing_trip.pipelineFlags.alternativesStarted:
+        existing_trip.pipelineFlags.startAlternatives()
+        existing_trip.pipelineFlags.savePipelineFlags()
+        if immediate:
+            schedule_queries(existing_trip.trip_id, existing_trip.user_id, [existing_trip], immediate, total_stagger)
+            total_stagger += stagger
+        else:
+            schedule_queries(existing_trip.trip_id, existing_trip.user_id, [existing_trip], immediate)
```
shankari: If we are not running immediately, we don't care about the stagger?

CFC_DataCollector/recommender/alternative_trips_pipeline.py
```diff
@@ -1,9 +1,13 @@
-from trip import E_Mission_Trip
+import optparse
+import sys
+import json
 from alternative_trips_module import calc_alternative_trips
 import tripiterator as ti
-from common import get_uuid_list
+from common import get_uuid_list, uuid_to_name
```
shankari: We don't want to make it easy to map from uuid back to name.

CFC_DataCollector/recommender/alternative_trips_pipeline.py
```diff
@@ -11,17 +15,30 @@ def __init__(self):
     def get_trips_for_alternatives(self, user_uuid):
         return ti.TripIterator(user_uuid, ["trips", "get_all"])
 
-    def runPipeline(self):
+    def runPipeline(self, immediate=False):
         for user_uuid in get_uuid_list():
+	    print "Finding Trips for User: ", uuid_to_name(str(user_uuid)), user_uuid
```
shankari: We don't want to be associating trips with actual users or user email addresses.

CFC_DataCollector/recommender/common.py
```diff
@@ -1,75 +1,108 @@
...
 import json
-import jsonpickle
-from trip import E_Mission_Trip
+from trip import *
+import copy
+import uuid
+import xml.etree.ElementTree as ET
+
+DATE_FORMAT = "%Y%m%dT%H%M%S-%W00" #This is a great hack thought of by Shaun
+CLASS_UUIDS = {
```
shankari: !!!
Why don't we just put all our UUIDs on a giant screen in the middle of the internet for everybody to look at
!!!

CFC_DataCollector/recommender/common.py
```diff
@@ -1,75 +1,108 @@
...
+	cost = 0
+	parent_id = trip_id
+	mode_list = [str(mode)]
+        return Alternative_Trip(_id, user_id, trip_id, sections, org_start_time, end_time, start_trip, end_trip, parent_id, cost, mode_list)
 
 def meters_to_miles(meters):
 	return meters * 0.000621371
 
-def calc_car_cost(trip_id, distance):
-	uuid = sectiondb.find_one({'trip_id': trip_id})['user_id']
```
shankari: @shanthistyle, why are we not using the users' MPG here?

CFC_DataCollector/recommender/testTrip.py
```diff
@@ -0,0 +1,33 @@
+from get_database import *
```
shankari: Shouldn't this be in tests?

CFC_WebApp/clients/choice/choice.py
```diff
@@ -43,13 +44,28 @@ def getResult(user_uuid):
...
+  #                         variables = json.dumps({'curr_view': getCurrView(user_uuid),
+  #                                      'uuid': str(user_uuid),
+  #                                      'client_key': Client("choice").getClientKey()}),
+  #                         gameResult = base64.b64encode(gamified.getResult(user_uuid)),
+  #                         dataResult = base64.b64encode(default.getResult(user_uuid)),
+  #                         recommendationResult = base64.b64encode(recommendation.getResult(user_uuid)))
+  # return renderedTemplate
+
+  # DEMO
+  originalTrip = get_trip_db().find_one({'user_id': UUID(user.uuid), 'recommended_alternative': {'$exists': True} })
```
shankari: @jimmydief, I can't merge this and deploy to the real server - showing a hardcoded trip from a particular user is not going to work. Also, other teams have their own screens. You need to fix this the right way before midnight tomorrow. Reverting now...

CFC_DataCollector/recommender/trip.py
```diff
@@ -97,15 +126,15 @@ def _end_location(cls, points):
...
                       {"$set": {"distance" : self.distance, "mode" : self.mode, "confirmed_mode" : self.confirmed_mode}},
                        upsert=False, multi=False)
 
 
 class E_Mission_Trip(Trip):
 
-    def __init__(self, _id, user_id, trip_id, sections, start_time, end_time, trip_start_location, trip_end_location, alternatives, perturbed_trips, 
-                 mode_list, confirmed_mode_list):
+    def __init__(self, _id, user_id, trip_id, sections, start_time, end_time, trip_start_location, trip_end_location, alternatives, perturbed_trips,
+                 mode_list, confirmed_mode_list, recommended_alternative=None):
```
shankari: @jimmydief, @sdsingh, why link from the trip to the recommendation? Why not from the recommendation to the original trip? That would allow flexibility to have multiple recommendations. An acceptable answer is "there's no particular reason, we just did something easy".

I just want to know so that we have a sense of what to change later.

CFC_DataCollector/recommender/traffic.py
```diff
@@ -0,0 +1,40 @@
+## A wrapper for the 511 traffic api
+import xml.etree.ElementTree as ET
+import urllib2, datetime
+
+def get_orig(orig_city):
+	url = "http://services.my511.org/traffic/getoriginlist.aspx?token=76495c63-cb01-4aac-af28-f717ffee11c8"
```
shankari: The token shouldn't be hardcoded and checked in.
Other projects will need other tokens, and we don't want to share our token and leave it open to abuse.

@joshzarrabi, I am removing this file for now since it is not being used, please fix later.

CFC_DataCollector/recommender/common.py
```diff
@@ -1,75 +1,108 @@
...
+        uuids.add(x['user_id'])
+    return uuids
 
-def update_perturbations(_id, perturbed_trip):
-	db = get_perturbed_trips_db()
-	_id = _id.replace('.', '')
-	json_trip = db.find_one({"our_id" : _id})
-	json_trip[perturbed_trip._id] = jsonpickle.encode(perturbed_trip)
+def get_recommender_uuid_list():
+    uuids = set()
```
shankari: @jimmydief @sdsingh, this is extremely slow. The correct way is to use "distinct" instead.

CFC_DataCollector/recommender/traffic.py
joshzarrabi: Sounds good,
we dont need it

shankari:
Failure is again 5 vs 15.
retest this please
If it fails again, I am going to disable the tests.

shankari:
retest this please

PR #118, opened 2015-05-06
Some more code around the switch from default -> data

shankari:
A couple of fallback places have been retained as default to indicate that we
should still have a reasonable default and that's what those should be.


shankari:
retest this please

PR #121, opened 2015-05-07
Integrate the canonical trip calculation with the rest of the recomme…

shankari:
…nder pipeline

The mode inference team had changed the canonical trip calculation to make it
work. However, the recommender pipeline was still returning the training trips
as canonical trips. This change hooks things up so that the recommender
pipeline returns the canonical trips

Extended the recommender pipeline test to test this case and fixed a bunch of
bugs found in testing. Notably, we should return an iterator, we should pass in
the Canonical_E_Mission_trip as a class to the iterator, we were assuming that
the input user UUID was encoded as a string?!

All tests currently PASS

### DIFF
--- a/CFC_DataCollector/recommender/filter_modules.py
+++ b/CFC_DataCollector/recommender/filter_modules.py
@@ -28,19 +28,28 @@ from main.get_database import get_routeCluster_db, get_section_db
 def get_clusters_info(uid):
         c_db = get_routeCluster_db()
         s_db = get_section_db()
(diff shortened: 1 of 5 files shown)


shankari:
@zacatac, @gaukes, @sdsingh, @jimmydief, I felt like it was easier for me to just do the integration than to argue about which team should do it, and how much testing they should do. I plan to merge this fairly soon after the tests finish passing, but if one of you wants to review the change in the interim, that would be great!

jimmydief:
Looks great to me @shankari

PR #127, opened 2015-05-09
Bunch of fixes encountered while pushing to production

shankari:
- Add pygmaps to the path just before and after generating the map
- Remove some hardcoded stuff in recommender that appears to have been
  (re)introduced as part of a merge conflict
- Fix a totally bizarre issue in which, on android, the webview is unable to
  correctly parse the CommonTripsCtrl if it is named anything other than
  "FooCtrl". Even "CooCtrl" doesn't work. I've only tried this in the emulator -
  maybe it works on a real phone. However, people might want to test in the
  emulator, so renamed to FooCtrl. At this point, I can see all the options on
  both emulators and my physical phone.

### DIFF
--- a/CFC_WebApp/clients/choice/front/ionic-tabs-myapp.js
+++ b/CFC_WebApp/clients/choice/front/ionic-tabs-myapp.js
@@ -138,11 +138,11 @@ angular.module('e-mission-choice', ['ionic'])
   }
 })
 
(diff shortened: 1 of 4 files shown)


shankari:
@zacatac, @jimmydief, can you review this change?

jimmydief:
@shankari other than the asinine FooCtrl thing, looks good!

PR #128, opened 2015-05-11
Add a version to the javascript to force download of the file

shankari:
This is moderately bizarre, but basically, the version currently deployed to
the app store appears to aggressively cache the supporting javascript files.
That means that it doesn't find the controllers that we added later. So I'm
using a trick that I found on stackoverflow - adding a version tag forces the
javascript to be retrieved again.

With this change, the existing production version also show all the new tabs


shankari:
@jimmydief @zacatac @lechiang 
Another small change. At this point, the new tabs are visible on the existing production e-mission as well.
This is a super small change, so I'm not going to wait long for a code review.

PR #131, opened 2015-05-27
Server changes to simplify development by skipping authorization

shankari:
When we are connected to a development server, skip the creation of the JWT and
just send the unencrypted email address instead. This makes it much easier to
start end to end development.


shankari:
Related pull requests in the data collection phone app:
e-mission/e-mission-data-collection#27

and the main phone app.
https://github.com/e-mission/e-mission-phone/pull/45

PR #133, opened 2015-05-31
Socialmedia Init

sfwatergit:
[Fixed PR from the other day]
Added a bit more functionality to get a working login flow initialized. Need to write some tests, but there's some meat here for people to start adding to, if they want to get going on this. Also, may be useful for me to get comments at this point, since, if I'm implementing this incorrectly, I'd like to make sure I don't get too far off track. 

Mainly, the idea is to use this piece of functionality as an alternative registration or login mechanism on the phone, which retrieves the facebook access token of the user, retrieves basic info, registers them via their facebook e-mail (assuming it's the same one used for a study, or this can be stipulated), and then passes the access token through a redirect to another page (function), which can then do whatever it wishes with the token (query friends, sociodemographics, etc.).

The testing machinery is partially in the expected places, but there are some additional things to know. Facebook lets us create and employ test users, so we can start messing around and robustly testing anything that requires a social graph component at this point. Many useful functions towards this end are on the FacebookUser class, which as the module evolves, I'd like to try to generalize to other social media.


shankari:
I can provide detailed comments on the code, but I'd like to get the high level discussion out of the way. Is the assumption here that the user can login via facebook instead of the current gmail-based login, or is the idea that the facebook integration would be used _in addition to_ the gmail-based login? It looks from the FacebookUser code that the goal is the former, in which case this won't work.

The user's email is not just used to uniquely identify them. It is not used to gain access to any google services (contacts etc) via an access token. It is also used for client -> server OAuth based authentication using a JWT (http://openid.net/specs/draft-jones-json-web-token-07.html), which is in turn used for cross-client authentication (https://developers.google.com/identity/protocols/CrossClientAuth).

I am pretty sure that Facebook can generate JWTs as well, but in order to do that, you need to authenticate with Facebook and cache the access and refresh tokens _on the phone_. That's because the phone needs to regenerate JWTs automatically in order to talk to the server - regenerating tokens on the server won't help.

Fortunately, with cordova, there is a plugin that can supposedly do this - http://ngcordova.com/docs/plugins/oauth/. I am not quite sure how to use it to retrieve a JWT - I spent ~ 20 mins poking around, and while I can see that the Facebook iOS SDK seems to do this (https://developers.facebook.com/docs/facebook-login/ios/v2.3, currentAccessToken), I am not quite sure how to get the javascript code to do this.

Alternatively, you can leave the client -> server authentication mechanism as google JWT and have the facebook login be used primarily to access facebook services, similar to the moves login. At that point, you can refresh the access tokens on the server, since we will be talking from our server directly to the facebook server.

Let me know which way you want to go, and I'll provide a more detailed code-level review.

sfwatergit:
Hi Shankari,

Right now, the tokens are generated by a call to /fbregister, which simply stores them as a cookie. The e-mail is simply used as a way to retrieve the uuid. If a user does not yet have a uuid, I’ve called the registration function for testing purposes, but that is one line and can be commented out. Alternatively, we can use the Facebook uid to retrieve the uuid. 

I think since Facebook is an adjunct and optional service, we can continue to use Google as the main sign-in. As long as a user is already signed into Facebook, there shouldn’t be much friction in having the two be separate. 

As for persistence, I’m not caching tokens yet. I’m using the javascript sdk at the moment, and retrieving the access token from cookies, then passing it to subsequent calls (one for now, really). If it is more efficient to cache the token, we can save the expiry date and there is a function in the python sdk to extend the token server-side. I figured this was the way to go, since it would keep the code platform independent. 

So, in summary, the design intent was more along the lines of the moves api. I don’t see much use in a separate login. 

Best, 


Sid Feygin
sid.feygin@gmail.com

shankari:
Good! That makes it much easier. I will now give a more detailed technical review, and merge once the issues are addressed, since this will unblock you for now. But from the long-term, I don't think that this is the correct architecture. This is because the "clients" are intended to make it easier to adapt the results to the user - one example of this would be running studies.

Some examples are:
- for the carshare study, they did not want to see any of the result or goal screens (carbon footprint, game, etc) in order to avoid conflating results through information with results through providing carshare
- if we want to do a randomized controlled trial, we can assign users to groups or "clients" and thus see completely different result screens

It is not clear to me that logging in to Facebook is a "result". Is it something that we only want to enable for certain groups or people? If not, we should probably include this as part of the onboarding process, and also make it available under "settings".

FYI, at that point, you could also use the ngcordova plugin to log into facebook and/or other social media.

sfwatergit:
OK. I agree that logon is not necessarily a client. 

I should have explained why I have `socialmedia` as a client then.

The `socialmedia`client is intended to return results to users participating in a study where a social network of connected e-mission users would be useful. 

The SocialmediaUser.FacebookUser are part of the general user dao library, which seems like the natural place for them. I intend these to handle most of the login and authorization functionality in a manner analogous to User (though my understanding of the full sign-on flow was not yet complete when I wrote this, I’m working through it now w/ from the app side as well, following your suggestion).

By keeping them separate, we can later add the Facebook login as an option separate to the Google flow. If we want to have additional authorization providers, we may want to further abstract common features into a parent class. There’s also this: https://github.com/avelino/bottle-auth <https://github.com/avelino/bottle-auth>, which can serve as a nice template.

On a side note, there seems to be a lot of internalized knowledge (for example what you mentioned about the clients), which would be useful to have in the developer docs. Would organizing all of the Readme.md files into a wiki make more sense at this point?

Best,

Sid Feygin
sid.feygin@gmail.com

shankari:
Before you start moving everything to the phone side, I just want to make sure that we have thought through the requirements and that I understand them, to understand where this fits in.

I think that we need to think through the optionality of the social media part a bit to decide where it needs to fit in to the rest of the screens.

Do you think that logging in to social media is an integral part of the base? Does it need to be hidden from studies/clients who don't want to use social media, or is it OK to have it there, but not show any results based on it?

This is a question of the level of customization required for a client. So far, the requirement was to customize the results. If we want to customize the settings screen as well, that's a reasonable request, but we'd have to build in additional customization there (and FYI, cordova will help immensely with this, because you can change pretty much every cordova screen on the fly without redeploying the app). This is why if you are building a platform, you want to have at least three clients - each client brings up new requirements, and then you can synthesize them into a single architecture, and by the time you are done, you are likely to have enough functionality, and a clear direction for most other clients. You are client #2 (the carshare study was #1), so there will be some extra pain here, but it will all be good at the end... :)

> By keeping them separate, we can later add the Facebook login as an option separate to the Google flow. 

I would like to be careful here to distinguish between authentication that is used for e-mission app <-> e-mission server communication, and authentication/authorization that is used for access to other services. The google auth flow is used for authentication between e-mission app <-> e-mission server. As I discussed before, if we want to provide alternatives to the google auth flow, we need to create a JWT on the phone and so on.

From our earlier discussion, what you want to do is to authenticate _and_ authorize with facebook for e-mission server <-> facebook communication. That is a fundamentally different requirement, and so it cannot replace the google flow, as we discussed. Maybe this is what you are trying to say, but it wasn't super clear to me from reading it, and I just want to make sure that we are on the same page.

If we are authenticating and authorizing for e-mission <-> external service communication, that is much more similar to the moves flow than the google flow. This would involve:
- the option to sign in during onboarding, 
- similar to the "Link to moves" button in the settings page, have other "Link to x" buttons in the settings page, or maybe make a new "services" page that can contain all the "link" buttons. The latter may be easiest since we won't have to change an existing native screen.
- since we are logging in directly from the phone and not from the server, you can simply use the ng-cordova auth plugins. I think that these will use the app installed on the phone if it exists, at least, I hope so :)

sfwatergit:
Comments inline


> On Jun 2, 2015, at 9:05 AM, shankari <notifications@github.com> wrote:
> 
> Before you start moving everything to the phone side, I just want to make sure that we have thought through the requirements and that I understand them, to understand where this fits in.
> 
> I think that we need to think through the optionality of the social media part a bit to decide where it needs to fit in to the rest of the screens.
> 
> Do you think that logging in to social media is an integral part of the base?
> 
Yes, but still optional.
> Does it need to be hidden from studies/clients who don't want to use social media, or is it OK to have it there, but not show any results based on it?
> 
Hiding is easy. Just add a tag or some other semantic element to the sign-on block. I am aiming to make this modular, but there is no clear architecture for this sort of module yet (though that is understandable considering the text of the following para).
> This is a question of the level of customization required for a client. So far, the requirement was to customize the results. If we want to customize the settings screen as well, that's a reasonable request, but we'd have to build in additional customization there (and FYI, cordova will help immensely with this, because you can change pretty much every cordova screen on the fly without redeploying the app). This is why if you are building a platform, you want to have at least three clients - each client brings up new requirements, and then you can synthesize them into a single architecture, and by the time you are done, you are likely to have enough functionality, and a clear direction for most other clients. You are client #2 <https://github.com/e-mission/e-mission-server/pull/2> (the carshare study was #1 <https://github.com/e-mission/e-mission-server/issues/1>), so there will be some extra pain here, but it will all be good at the end... :)
> 
> By keeping them separate, we can later add the Facebook login as an option separate to the Google flow.
> 
> 
Since modifications to the User are probably going to be common, but we don’t necessarily want to override the basic functionality provided by your pre-existing User, perhaps we should have an abstract Profile class (maybe named differently) that would be optional to each client.
> I would like to be careful here to distinguish between authentication that is used for e-mission app <-> e-mission server communication, and authentication/authorization that is used for access to other services. The google auth flow is used for authentication between e-mission app <-> e-mission server. As I discussed before, if we want to provide alternatives to the google auth flow, we need to create a JWT on the phone and so on.
> 
> From our earlier discussion, what you want to do is to authenticate and authorize with facebook for e-mission server <-> facebook communication. That is a fundamentally different requirement, and so it cannot replace the google flow, as we discussed. Maybe this is what you are trying to say, but it wasn't super clear to me from reading it, and I just want to make sure that we are on the same page.
> 

I’m not interested in replacing  the Google flow. I think a settings screen is appropriate. The attached is an example of the kind of thing that would be most useful. 



Then, I imagine the access token would be handed to the server via post request? I imagine this would need to contain the UUID.

shankari:
> I’m not interested in replacing  the Google flow. I think a settings screen is appropriate. The attached is an example of the kind of thing that would be most useful.

I don't see an attachment. Maybe you can add the image directly in github.

> Hiding is easy. 
I don't say that it is not. But since you are the second client of the platform, I want to understand what the actual requirements are in greater detail before figuring out how to implement the requirement.

> Yes, but still optional.
> Does it need to be hidden from studies/clients who don't want to use social media, or is it OK to have it there, but not show any results based on it?

These are two slightly different things, and I am not sure which one you are targeting.
1. In the first one, we _always_ show the login screens for services in the onboarding flow, but with a "skip" button that allows people like me to skip the login. Similarly, we have _always_ have a "services" menu item with all the services that we support listed - the user can choose whether or not to log in to them.
2. In the second one, we show the login screens for services only in the onboarding flow for certain users. For example, if we are running an experiment in which only some users are allowed to use social pressure and other users are not, we would not even want to allow other users to see the facebook login screen. Similarly, the "services" screen should not show the facebook screen, or should have it be disabled.

The second one is obviously more flexible, but is also more complicated. What is the actual requirement?

sfwatergit:
Honestly, I don't think we need to have any of the services in the login flow. The pattern I'm used to seeing (which I imagine will be most useful for users) is that additional login services are available for users who wish to do so in a settings menu. Let's keep Google auth flow and just have an additional settings screen (for a few screens, we should do tabs; if more, then a matrix as on the Apple or Ubuntu System Settings dialog) where users can choose from a menu of accounts to add, as in the Apple accounts menu (see photo, attached). If a study is really contingent on having users pass through this, it can be inserted following the Google sign-in as a fragment. Else, if a study wants to disallow this as part of the flow, we remove the tab. I'm fine programming this native or I can figure it out in angular/ionic.

screenshot_6_2_15__2_57_pm

CFC_WebApp/api/cfc_webapp.py
```diff
@@ -352,6 +354,10 @@ def javascriptCallback(clientName, method):
   client.callJavascriptCallback(client_key, method, request.params)
   return {'status': 'ok'}
 
+@get('/fbregister')
```
shankari: This is not really used anywhere in the current implementation.
Remove until it is actually needed.

CFC_WebApp/clients/socialmedia/result_template.tpl
```diff
@@ -0,0 +1,98 @@
...
+  // code below.
+  function checkLoginState() {
+    FB.getLoginStatus(function(response) {
+      statusChangeCallback(response);
+    });
+  }
+
+  window.fbAsyncInit = function() {
+  FB.init({
+    appId      : '456261131199824',
```
shankari: Shouldn't this be part of the keys as well?

CFC_WebApp/clients/socialmedia/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "socialmedia",
+  "start_date": "2014/10/13",
+  "end_date": "2015/06/01",
+  "key": "dc23ad22-bc78-4fd6-9adf-ca2b04505916",
```
shankari: You generally don't want the key to be checked in. It should be generated on each server.

CFC_WebApp/clients/socialmedia/result_template.tpl
sfwatergit: Oops. Yes. Thanks! I’ll pass it in on as a template variable.
Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/clients/socialmedia/settings.json
```diff
@@ -0,0 +1,11 @@
...
+  "client_settings": {
+    "result_url": "https://e-mission.eecs.berkeley.edu/socialmedia",
```
shankari: You do not want this. Setting this is intended to display https://e-mission.eecs.berkeley.edu/socialmedia in the result screen instead of 
https://e-mission.eecs.berkeley.edu/compare.

And you haven't even defined the https://e-mission.eecs.berkeley.edu/socialmedia endpoint.

I might have disabled the client settings code in the apps - that is pretty much dead code at this point. But at any rate, just leave this as "compare".

CFC_WebApp/clients/socialmedia/settings.json
sfwatergit: This was an arbitrary key I was using to figure out what was going on with the clients. I’ll change it back to whatever the dummy key was.

Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/clients/socialmedia/settings.json
```diff
@@ -0,0 +1,11 @@
...
+    "fb_app_secret": "051bb7402ba81dbbd3a7d11030567d07",
```
shankari: Some more keys!!

CFC_WebApp/clients/socialmedia/settings.json.sample
```diff
@@ -0,0 +1,4 @@
+{
+    "fb_app_secret":"ask sid",
```
shankari: json.sample doesn't seem to be similar to json. I presume you want to delete .json, but then if somebody takes .json.sample and copies it over, I don't think that it will work.

sfwatergit: Yes.

I later integrated all of this into the socialmedia client’s settings. So, this file is an artifact of that.

But for future purposes, what is the best place to put these?

Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/api/cfc_webapp.py
sfwatergit: fbregister actually does call the function in socialmedia client to register the user. This facebookLogin() function may be found in socialmedia.py. It renders the template and passes in the client key. Before you pointed out that Client has an update function, I was using the manual update script to populate the client db. Anyway, the subsequent call on successful login redirects the user to a page which calls ‘getFacebookUser.get_user_from_cookies(cookies)’, which then calls the FacebookAPI to get the user’s e-mail. Of course, this could all be simplified by one call if we specify the Facebook UID or e-mail as the foreign key to the uuid_db().  

On a general note: Yes, the current implementation is a rather crude, inefficient, and opinionated one, but was meant to get the conversation going, figure out the moving parts of your system, throw out some ideas, and get folks towards consensus on what a good implementation would look like. I don’t have a full vision of how this should integrate, but the QA/QC is definitely appreciated.

Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/clients/socialmedia/settings.json
sfwatergit: Darn. Thought I’d put these in the .gitignore. 
Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/api/cfc_webapp.py
shankari: My point is that I don't see anything in your code that makes a REST API call to:
https://<host.name>/fbregister
and until it does, there is no point in adding the API endpoint here.

sfwatergit: OK. Removing.

Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/clients/socialmedia/settings.json.sample
shankari: I think that it is fine to leave it as settings.json.sample. My main comments are:

1. Maybe we want to have separate settings for each external service (e.g. `facebook_settings.json.sample`). I am not too insistent on this and am willing to accept your current implementation (`settings.json.sample` with service specific prefixes, e.g. `fb_app_secret`).
1. The format of `settings.json.sample` should be the same as the format of `settings.json`. Otherwise, people will replace the keys and then spend a bunch of time trying to figure out why they are getting parse errors. In your case, for example, in `settings.json`, `fb_app_secret` is a child of `client_settings` while in `settings.json.sample`, it is a top level element.

shankari:
> On a general note: Yes, the current implementation is a rather crude, inefficient, and opinionated one, but was meant to get the conversation going, figure out the moving parts of your system, throw out some ideas, and get folks towards consensus on what a good implementation would look like. I don’t have a full vision of how this should integrate, but the QA/QC is definitely appreciated. 

It is perfectly fine to start off with a strawman implementation and then discuss around it to understand requirements and architecture. I have tried to make my comments follow good critique guidelines and be kind, specific and helpful. In response to your previous questions about code reviews, I hope this also provides you with an example of what code reviews look like and why lots of organizations like them...

CFC_WebApp/clients/socialmedia/socialmedia.py
```diff
@@ -0,0 +1,55 @@
...
+from get_database import get_client_db
+
+FB_BASE_DIR = dirname(dirname(__file__))
+
+
+def getFacebookUser(cookies):
+    return FacebookUser.get_user_from_cookies(cookies)
+
+
+def facebookLogin():
```
shankari: Hm, I see what you are trying to do here. If you want to display this screen as one of the tabs in the result screen, you need to modify the choice screen to add it. Otherwise, this is code that is never exercised, specially since we are going to remove the API endpoint.

CFC_WebApp/dao/socialmediauser.py
```diff
@@ -0,0 +1,235 @@
...
+from get_database import get_uuid_db
+
+# Seems more appropriate to have these here, but can move to the ../socialmedia.py file or to the main api
+fb_config = open('clients/socialmedia/settings.json')
+fb_data = json.load(fb_config)['client_settings']
+FACEBOOK_APP_ID = fb_data['fb_app_id']
+FACEBOOK_APP_SECRET = fb_data['fb_app_secret']
+
+
+class FacebookUser:
```
shankari: So is the plan that FacebookUser is an aspect of a regular user. Because we might want to have TwitterUser, WhatsappUser etc, all defined for the same user, if the user has logged into multiple external services at the same time.

CFC_WebApp/clients/socialmedia/socialmedia.py
sfwatergit: True… was going to defer messing w/choice until I got Cordova set up.

CFC_WebApp/tests/client_tests/TestSocialMedia.py
```diff
@@ -0,0 +1,20 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+
+class TestSocialMedia(unittest.TestCase):
+    def setUp(self):
+        from tests.common import dropAllCollections
+        dropAllCollections(get_db())
+        user = User.register("oqsaqjk_adeagbosky_1432921786@tfbnw.net")
+        self.uuid = user.uuid
+
```
shankari: Shouldn't you call some of the test methods from in here? Otherwise, how do you know that they are invoked?

CFC_WebApp/clients/socialmedia/settings.json.sample
sfwatergit: Yeah. I think having a service-specific settings file makes sense if you are sensitive to the format of the settings.json file. However, my personal preference is externalizing settings as development environment variables. 

I had some interest in using this for a previous project: http://yamjam.readthedocs.org/en/latest/index.html <http://yamjam.readthedocs.org/en/latest/index.html>, which is not just django-specific. This hoists the configuration variables out of anything under source control while implementing a reasonably simple API.   

It may be worth the investment of time now to have a central config structure and avoid sprinkling config files in various locations and having to manage the risk of inadvertent check-in later. Perhaps there is some sense to this that I am unfamiliar with; however. 

Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/dao/socialmediauser.py
sfwatergit: Yes. That is why this is all in a file called SocialmediaUser. I have not stubbed these out yet, but I did consider that we will want additional services as an option.
Sid Feygin
sid.feygin@gmail.com

CFC_WebApp/tests/client_tests/TestSocialMedia.py
sfwatergit: Will do.
Sid Feygin
sid.feygin@gmail.com

shankari:
At a high level, I am not sure whether you want to continue fixing this PR or whether you want to switch to the phone app right now, given that the current implementation is not visible in the UI since it has not been integrated into the choice view. If this was almost working, it might be worthwhile to clean it up further, but since we have discussed that we want the "services" screen to be on the client, maybe you don't want to put a lot of effort into cleaning this up right now.

> I'm fine programming this native or I can figure it out in angular/ionic.

The rule of thumb that we came up with for this was that you would use ionic unless there was a need for native - just because of the crossplatform nature and the dynamic updates. You can certainly do native with a justification, but then you have to implement it in both android and ios.

If you do choose to go with cordova, there's a nice plugin that supports browser-based authentication with tons of services - you can get the access token and then send it to the e-mission server via HTTP POST, just like we currently do with moves.
http://ngcordova.com/docs/plugins/oauth/

Here's a pre-built example that you can work from.
https://github.com/nraboy/ng-cordova-facebook-example

shankari:
Final high level architectural issue:

Your current implementation has the Facebook integration code in CFC_WebApp. That was possibly driven by your original vision of integrating with a javascript based screen. But if you look at the moves integration, the code is actually in CFC_DataCollector.

The reason is that our communication with external services is not driven by user interaction. We don't just want to retrieve the graph and display it to the user, we want to retrieve and store the graph, and then perform analysis on it (I assume, correct my understanding if incorrect). So our retrieval does not happen in the context of a bottle/django/webserver process, but in the context of a regularly scheduled process like a cronjob or other scheduled process.

Again, I think that the moves integration is the template that you want to follow architecturally.

sfwatergit:
I wasn’t planning on running calls to the Graph api as a cronjob. It is in actuality a user-driven process, since the sign-on of a person’s friend will update our internal representation of the graph of connected users, but, beyond this, I anticipate that most information will be pulled immediately or through user interaction. 

The alternative implies that we will build and maintain an internal subgraph and write an api that may be queried from the point of view of individual users. This will get woefully complicated to keep straight, and it is really easy to just let Facebook handle that part. 

I do not think that Facebook <—> Moves is a one-to-one mapping in terms of how users will interact with it and its role in the app. The Facebook state is just much more dynamic, the app_token must be refreshed on a regular basis, and I foresee the user interaction will be largely driven by the activity of other users in addition to regular number crunching via cron-jobs server-side.

However, every time the user wants to see how they are doing compared to their friends, this should be a call to the Graph API to ensure the connections are up-to-date. Running this asynchronously from requests does not make sense to me. 

The cron jobs will cache the user’s and build the global graph in a form that can be exported to, say, NetworkX, but will not necessarily host this in the form of an API. We will also want to discuss how this object will be stored, how it will be anonymized, and so forth. 

So, before I take a deep compare and contrast w/ moves api (which seems to be going away?), it may be worth having a chat via Skype/BlueJeans and develop some graphical artifacts to guide forward progress on this.


Sid Feygin
sid.feygin@gmail.com

shankari:
> it may be worth having a chat via Skype/BlueJeans and develop some graphical artifacts to guide forward progress on this

Agreed. Actually, I am on campus next Monday, so that would be a good time to meet in person and draw diagrams on whiteboards and so on. You could work on exploring cordova in the meanwhile. I think that it would be really useful for me to have a clearer picture of your storyboards around this, and to have some time to think through them before we meet.

Some more clarifications and areas of confusion - I hope they can ensure that we are on the same page:

> before I take a deep compare and contrast w/ moves api (which seems to be going away?), 

the moves api is going away because we have a replacement for it - not because the architecture doesn't work. We could actually keep it around as an example if we want to, although we've have to think of a use case for the data.

> the app_token must be refreshed on a regular basis

this needs to happen for moves too, and for all oauth based systems. It can happen in the background without user interaction by using the refresh token. It's just that on moves, the access token is good for 6 months.

> I foresee the user interaction will be largely driven by the activity of other users in addition to regular number crunching via cron-jobs server-side

I think that this comes down to requirements again. What is the activity of the other users? What is the number crunching that you envision? Maybe you can write your vision down in a document and give me some time to think through it before we meet. I think that it is important to start with what we need rather than what we can do - in other words, start with the requirements instead of the architecture. 

> The cron jobs will cache the user’s and build the global graph in a form that can be exported to, say, NetworkX, but will not necessarily host this in the form of an API.

You need to update this cache somehow whether you export it as an API or not. Do you need user interaction to update the cache? Why? I agree that it doesn't make sense to rehost as a raw API (and might violate Facebooks T&C), but we don't really do that with moves either. What do you plan to do with the cache? Do you plan to listen to other user activity as you imply above? Will you use that activity to prompt the user? Why should the user use this app to interact with other users - wouldn't she just want to interact with them directly through facebook?

shankari:
Closing this pull request and waiting for one that follows the architecture that we defined.

PR #134, opened 2015-06-03
changed end dates in the setting.json; gave us another year

joshzarrabi:


shankari:
Can you remove `commontrips/result_template.html`? That is generated every time, and so checking it in will result in constant conflicts as it gets updated by different people in different ways. As we discussed during the retreat, the code shouldn't be generating the result template in the first place because of race conditions across multiple users, but as long as we are doing it, we shouldn't be checking it in...

shankari:
Otherwise :+1:

PR #135, opened 2015-06-05
First Azure stuff; added support for online SQL DB

joshzarrabi:


CFC_WebApp/clients/leaderboard/front/client.leaderboard.display.js
```diff
@@ -8,6 +8,10 @@ angular.module('e-mission-leaderboard', ['ionic'])
      * but not tested
      */
     $scope.showscores = function() {
+      var client = new WindowsAzure.MobileServiceClient(
```
shankari: Where are you using the client? Is this required?

shankari:
High level comments:

1. The SQL part seems pretty trivial, which is good!
2. The javascript part is... javascript. A similar option will certainly work for cordova, which is javascript based, after all, but won't provide background sync, as we have discussed extensively earlier.

joshzarrabi:
Oops!
I forgot the js stuff is there.
Sorry!

CFC_WebApp/clients/commontrips/result_template.html
```diff
@@ -0,0 +1,31 @@
+<html>
```
shankari: Please remove this file - the result template is overwritten every time.

CFC_WebApp/clients/leaderboard/SQL.py
```diff
@@ -0,0 +1,23 @@
+import pyodbc
+
+CONNECTION = pyodbc.connect("DRIVER={SQL Server};SERVER=tcp:xalxu8zk56.database.windows.net,1433;DATABASE=e-mission_db;Uid=joshz@xalxu8zk56;PWD=Nathan123")
```
shankari: Are we retaining this file?
If so, shouldn't it support both Azure and Couchbase with an abstraction layer on top?

CFC_WebApp/clients/leaderboard/leaderboard.py
```diff
@@ -8,12 +8,28 @@
...
+#import SQL
+#from couchbase.bucket import Bucket
+
+
+## The couchbase python library only compiles on Linux Machines so be careful 
+## pip install couchbase
+
+## Cloud abstraction choice
+cloud = { 
+          "AZURE" : False,
```
shankari: Why are all of these False?

CFC_WebApp/clients/leaderboard/leaderboard.py
```diff
@@ -145,8 +167,12 @@ def getResult(user_uuid):
   for user_uuid_dict in get_uuid_db().find({}, {'uuid': 1, '_id': 0}):
     (currPrevScore, currCurrScore) = getStoredScore(User.fromUUID(user_uuid_dict['uuid']))
     otherCurrScoreList.append(currCurrScore)
+    if cloud['AZURE']:
```
shankari: Why is this "if" out here? Why isn't it part of the abstraction layer?

CFC_WebApp/clients/leaderboard/leaderboard.py
```diff
@@ -8,12 +8,28 @@
 import math
 from get_database import get_uuid_db
 from main import userclient
+#import SQL
+#from couchbase.bucket import Bucket
```
shankari: why is this commented out? How are you able to import bucket in that case?

CFC_WebApp/clients/leaderboard/result_template.html
```diff
@@ -29,9 +29,13 @@
         <body onLoad='showscores()'></body>
     </div>
     <script>
+        var client = new WindowsAzure.MobileServiceClient(
```
shankari: So result_template is currently loaded from the server and has the results encoded in it. We want to switch to a model in which we read the data from the local database that has been synced there earlier. I don't see that change in here. I hope that is in the phone code (I'm going to review that next), but in that case, you should revert the changes to this code since we don't want to change it....

shankari:
At a high level, I still don't see how we authenticate anything. When you save the data, you just add a bucket with the UUID -> score mapping. Where does couchbase retrieve the UUID from the JWT and use it to look up the score?

PR #137, opened 2015-06-11
fixed gmap_display

nephraim:


CFC_WebApp/main/gmap_display.py
```diff
@@ -148,7 +148,7 @@ def drawSection(section, option, gmap, Color = 'default'):
                 # coordinates are in GeoJSON format, ie lng, lat
                 gmap.addpoint(coordinate[1], coordinate[0], color)
         if option == PATH or option == ALL:
-            gmap.addpath(path, color, _id)
```
shankari: This was a change that was made by @jeffdh5 in order to display the ID as a tooltip for the point, which would help us visualize the data better. Do you agree?

It is not clear to me why having the third parameter does not work. gmap_display has the line:

     import pygmaps_modified as pygmaps

and I see a 

     def addpath(self,path,color = '#FF0000', id = ''):

in pygmaps_modified....

nephraim: I think I figured it out. In utils/plot_trip.py, it was importing pygmaps instead of pygmaps_modified, so the modified function was never being called. I'm fixing it.

shankari: awesome.

PR #139, opened 2015-06-13
Improvements to visualizing trips related to ground truthing smoothing

shankari:
The details are in the individual commits. @joshzarrabi @nephraim can you take a look?

### DIFF
--- a/CFC_DataCollector/get_database.py
+++ b/CFC_DataCollector/get_database.py
@@ -40,6 +40,12 @@ def get_routeCluster_db():
     routeCluster=current_db.Stage_routeCluster
     return routeCluster
 
+def get_groundClusters_db():
+    current_db=MongoClient().Stage_database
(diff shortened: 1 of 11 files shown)


shankari:
@nephraim, @joshzarrabi  I plan to merge this tomorrow morning...

joshzarrabi:
Great job!
Just a few things I noticed, mostly style... 

I'm not sure if its good style to not include an __init__ statement for the SmoothingEvalResult class in gps_smoothing.py, maybe you should initialize the variables there to make it more explicit, but if you think it is fine the way it is then it probably is. (couldn't find anything about this in PEP)

On lines 54-57 in useful_queries.py you choose arbitrary numbers for your original bounds which would work, but its probably better to use the python implementations of infinity and -infinity, which are float('inf') and float('-inf'). 

Also we should clean up the imports at the top of gmap_display.py

PR #140, opened 2015-06-18
Support more sophisticated smoothing

shankari:
That is not handled by the simple boundary or adaptive conditions that we tested earlier


shankari:
@joshzarrabi, @nephraim, this is probably too detailed into the smoothing code for you, but feel free to take a look if you have the time.

PR #141, opened 2015-06-28
Add some more code for exploratory analysis

shankari:
In particular, try to figure out the kinds of trajectories that might need
smoothing, using both manual rules, and machine learning and plot their before
and after values.

Also do some code cleanup:
- move the script to generate ground truth for smoothing into the new directory
- add a README
- refactor to draw from section json list in addition to a section id list, so
  that we don't have to keep moving back and forth
-


shankari:
@nephraim, this is the analysis that I did to generate the graphs in the smoothing_report.

shankari:
I plan to merge this sometime tomorrow

PR #142, opened 2015-06-29
Begin unifying/refactoring the python module structure

shankari:
As a first step, unify all the get_database code into one so that we don't have
to copy it around to multiple files all the time.


shankari:
@joshzarrabi @nephraim @sdsingh the great cleanup begins
in small, small steps

PR #143, opened 2015-06-30
Initial implementation of the user cache

shankari:
This includes:
- definition of the abstract superclass for the usercache
- a start on the implementation of the "builtin" usercache, which uses a
  collection in mongodb for the storage
- unit tests for the implemented code


shankari:
@joshzarrabi @nephraim @farietta @scisley, a start on the implementation on the server side

PR #146, opened 2015-07-01
Added support for our own trip generation

joshzarrabi:
Added support for our own trip generation. 
Make trips by running create_fake_trips() in trip_gen.py


CFC_DataCollector/trip_generator/input.json
```diff
@@ -0,0 +1,15 @@
...
+		"2703 Hallmark Dr Belmont Ca",
+		"ATT Park"
+
+	],
+	"ending centroids" :
+	[
+		"UC Berkeley",
+		"Morgan Hill"
+	],
+	"number of trips" : 100
```
shankari: Suggestions for optional enhancements:

1. Allow people to specify the number of trips of each mode
1. Allow people to specify the total number of trips, and the probability distribution of each mode
1. Allow people to specify a set of tour models (bonus: will help us define what a tour model looks like)

@nephraim, can you think of other options that you might find useful?

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -0,0 +1,136 @@
...
+    def get_trips_from_a_to_b(self):
+        mode_tuple = ("CAR", "WALK", "BICYCLE", "TRANSIT")
+        curr_time = datetime.datetime.now()
+        curr_month = curr_time.month
+        curr_year = curr_time.year
+        curr_day = curr_time.day
+        curr_hour = curr_time.hour
+        curr_minute = curr_time.minute
+        for t in self.a_to_b:
+            mode = mode_tuple[random.randint(0, len(mode_tuple) - 1)] ## Unsophisticated mode choice
```
shankari: Enhancement: Allow people to specify the random seed so that we can get arbitrary, but reproducible trips.

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -0,0 +1,136 @@
...
+                otp_trip = OTP(t[0], t[1], mode, write_day(curr_month, curr_day, curr_year), write_time(curr_hour, curr_minute), True)
+                alt_trip = otp_trip.turn_into_trip(0, 0, 0)   ## ids dont matter here 
+                alt_trip.save_to_db()
+            except PathNotFoundException:
+                print "In the sea, skipping"
+                self.amount_missed += 1
+
+def geocode_address(address):
+    if address.cord is None:
+        business_geocoder = Geocoder()
```
shankari: Switch to OSM?

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -0,0 +1,136 @@
...
+    u = random.random()
+    v = random.random()
+    w = radius_in_degrees * math.sqrt(u)
+    t = 2 * math.pi * v
+    x = w * math.cos(t)
+    y = w * math.sin(t)
+    x = float(x) / float(math.cos(y_0))   # To account for Earth something 
+    return Coordinate(x + x_0, y + y_0)
+
+def test_get_random_point():
```
shankari: move this to a separate unit test file so that we know if it breaks.
Look at the examples - they are really easy to write.

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -0,0 +1,136 @@
...
+    a = Address("2703 Hallmark Dr Belmont CA")
+    print "1 kilometer: %s" % get_one_random_point_in_radius(a, 1)
+    print "2 kms : %s" % get_one_random_point_in_radius(a, 2)
```
shankari: Ah, you are only printing, not checking the results? Pretty easy to do check the results automatically as well if you can specify a random seed.

CFC_DataCollector/trip_generator/trip_gen.py
joshzarrabi: Its making very few calls for now so limits shouldn't be a problem, but I can.

base/get_database.py
```diff
@@ -133,3 +133,8 @@ def get_usercache_db():
     current_db = MongoClient().Stage_database
     UserCache = current_db.Stage_usercache
     return UserCache
+
+def get_fake_trips_db():
```
shankari: For the unit tests, I am not sure whether these should be stored into the Fake_Trips_db. The rest of the code will assume that they are in the section Db like real sections. Might be a good idea to allow users to specify which DB they want the fake trips to go into.

shankari:
@joshzarrabi, looks great! I have several suggestions for enhancements, but I'm mainly writing them down for the record. I am going to merge this as is.

@nephraim, you can start using this beginning now. Please use this pull request, or a new issue that you file to provide suggestions on functionality that you would like to see.

CFC_DataCollector/trip_generator/trip_gen.py
joshzarrabi: Okay, good idea

CFC_DataCollector/trip_generator/trip_gen.py
joshzarrabi: Yeah, this was mostly just a sanity check thing, I probably should have removed it.

base/get_database.py
joshzarrabi: Why not, I think the fake trips should be separated from everything else...

CFC_DataCollector/trip_generator/input.json
joshzarrabi: Okay, I can start working on these now.

nephraim: Maybe you could allow people to specify the probability of taking a trip to or from specific places?

joshzarrabi: Okay!

I'll check it out.

Josh

CFC_DataCollector/trip_generator/trip_gen.py
shankari: Yeah, it's just to set an example from the beginning. One thing that I've realized is that most of the people working on this project don't really spend a lot of time thinking about alternatives - they just copy and paste what we already have. And if we start out with google maps, we will end up with lots of google maps calls sprinkled all through the code.

joshzarrabi: Okay,

I'm working on writing tests right now, so I'll include that with that pull
request, (either tonight or tomorrow)

CFC_DataCollector/trip_generator/trip_gen.py
shankari: no, no, don't remove it. move it to a unit test instead.
this is what we use unit tests for - the advantage is that then the test sticks around and can be run again and again to ensure that somebody else didn't screw things up.

joshzarrabi: I'm changing it so it actually runs a test and putting it in a test file.

base/get_database.py
shankari: I think that it is fine for the fake trips to be separated. If you do it this way, though, scripts written to work on real data won't work on fake data.

Let me give you a concrete use case. The recommendation pipeline tests ensure that the recommendation pipeline reads sections and processes them correctly. But the recommendation pipeline reads sections from the section DB.

So if you store fake trips in the fakeDB and then run the recommendation pipeline, it won't do anything.

There are a couple of other ways of doing this:

1. cheap and easy: use a unique uuid prefix or suffix to indicate that the records belong to a fake person.
1. more complicated but more correct: create a factory method that returns different values of get_db() based on the context/config. Then, during normal operation, have it return Staging_database and during test operation, have it return Test_database or Fake_database. This way, we can run tests on the same database as our real data without any conflicts.

But in both cases, you will be writing into the trip_db and section_db collections.

PR #147, opened 2015-07-02
Fixes from production

shankari:


shankari:
@farietta, here's the fix to your recommender screen.

PR #151, opened 2015-07-08
Fake trips and clustering

nephraim:
For the fake trip generation, I updated the code to include labels so that I can compare my the clusters that my algorithm outputs to the correct clusters. 

For the clustering algorithm, featurization.py tests for how many clusters there should be, creates the clusters, and compares them to the ground truth. By default, it runs the algorithm on the fake trips database, but you can also pass in your own data to be clustered, the range of numbers of clusters to try, and the clustering algorithm you want to use. I am working on extending it to work better on real data and to better predict the number of clusters.


shankari:
ok to test

shankari:
jenkins, please test this change

shankari:
jenkins, test this change please

shankari:
jenkins, test this please

CFC_DataCollector/recommender/otp.py
```diff
@@ -130,7 +130,7 @@ def turn_into_trip(self, _id, user_id, trip_id, is_fake=None):
 
         mode_list = list(mode_list)
         if is_fake:
-            return Fake_Trip(_id, user_id, trip_id, sections, final_start_time, final_end_time, final_start_loc, final_end_loc, 0, cost, mode_list)
+            return Fake_Trip(_id, user_id, trip_id, sections, start_time, end_time, final_start_loc, final_end_loc)
```
shankari: why do you want to use start_time and end_time, but final_start_loc and final_end_loc? Please add a comment explaining your decision for future reference.

CFC_DataCollector/trip_analysis/kmedoid.py
```diff
@@ -0,0 +1,122 @@
+import random
+import sys
+
+#code based on K_medoid_2.py in CFC_WebApp/main
```
shankari: If so, can you delete the one in main? It really shouldn't be in main because it is not really part of the webapp.

shankari: Also, can you summarize the improvements that you made to it?

shankari: Also, can you add a unit test to ensure that it continues working as expected even after you leave?

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -124,13 +151,3 @@ def create_fake_trips():
     my_creator.get_trips_from_a_to_b()
     return my_creator
 
-def sanity_check():
```
shankari: Move this to a unit test? Either @nephraim or @joshzarrabi should do this.

CFC_DataCollector/trip_generator/input.json
```diff
@@ -1,15 +0,0 @@
-{
-	"radius" : "1",
```
shankari: Why delete this? If you delete it, people will have to reverse engineer the format by looking at the code, a process that which 99% of the people in the world hate. Having the input file here helps a lot. If you are concerned about privacy or inadvertent use. use dummy values and rename it to input.json.sample if you like.

CFC_DataCollector/trip_analysis/kmedoid.py
nephraim: If we delete the one in main, then cluster_pipeline won't run. Let me first integrate it so that mine works for any kind of data, and then we can delete the other one. 

I did not make improvements to the actual k-medoids algorithm, but I changed the file so that it works with data that you pass in instead of user data, I added a distance calculator function that works with my featurization scheme, and I added a method that constructs the distance matrix within the file. I will make a unit test, add comments, and improve this file to work in all cases.

CFC_DataCollector/trip_generator/input.json
nephraim: I thought that we weren't supposed to upload json files. I will add it back.

CFC_DataCollector/recommender/otp.py
nephraim: @joshzarrabi this was your change

joshzarrabi: Hi, that's a mistake, @nephraim  you should change it back...

CFC_DataCollector/trip_analysis/kmedoid.py
shankari: I would actually split out the functionality:
1. one file for k-medoid
2. one file for featurization
3. one file for distance matrix calculation

Or maybe you could combine 2. and 3.

Couldn't you refactor the existing code so that it works with an iterator, and either pass in a cursor or a list of sections depending on the call site? Copy/pasting is never a good idea...

CFC_DataCollector/trip_generator/input.json
shankari: We weren't supposed to upload json files which included secret keys :) It is fine to upload samples of json files to indicate the format, and in fact, I have several existing one (keys.json.sample, or config.json.sample).

shankari:
I plan to merge once you have re-introduced the json file and reverted the change in trips.py. You can submit future cleanup in a separate pull request.

CFC_DataCollector/trip_analysis/kmedoid.py
nephraim: Yeah, I can do that.

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -96,13 +120,16 @@ def get_one_random_point_in_radius(address, radius):
     t = 2 * math.pi * v
     x = w * math.cos(t)
     y = w * math.sin(t)
+    print "type of y is %s" % type(y_0)
+    print "x = %s" % x
     x = float(x) / float(math.cos(y_0))   # To account for Earth something
```
shankari: Earth something?

CFC_DataCollector/trip_generator/trip_gen.py
```diff
@@ -96,13 +120,16 @@ def get_one_random_point_in_radius(address, radius):
     t = 2 * math.pi * v
     x = w * math.cos(t)
     y = w * math.sin(t)
+    print "type of y is %s" % type(y_0)
+    print "x = %s" % x
     x = float(x) / float(math.cos(y_0))   # To account for Earth something 
-    return Coordinate(x + x_0, y + y_0)
+    to_return = Coordinate(y + y_0, x + x_0)
+    print to_return
+    if abs(to_return.get_lat()) < 30 or abs(to_return.get_lon() < 30):
```
shankari: Why do you need this? Add a comment in the file describing why?
What is so magical about 30, specially for longitude.

shankari:
Added some more comments, but they can be addressed later.
Merging now as promised.
Congratulations, @nephraim!

PR #152, opened 2015-07-10
Tour model

joshzarrabi:
Added tour model code, added unit tests for it.
Its a basic data structure that can represent the tour model type activities of a user of the app.

### DIFF
--- a/CFC_DataCollector/tests/tripGenTests/TestTourModel.py
+++ b/CFC_DataCollector/tests/tripGenTests/TestTourModel.py
@@ -1,83 +1,92 @@
+import sys
+sys.path.append("./../../trip_generator")
 import unittest
-from trip_generator.tour_model import Location, TourModel
+from tour_model import Location, TourModel
 
-class TestTourModel(unittest.Testcase):
+class TestTourModel(unittest.TestCase):
 
-	def test_something(self):
-	    """ Basic Sanity Check """
-	    home = Location("home")
-	    work = Location("work")
-	    coffee = Location("coffee")
-	    home.is_home = True
-	    home.add_to_successors(work)
-	    work.add_to_successors(coffee)
-	    tm = TourModel(home, "test")
-	    tm_from_work = tm.get_tour_model_from(work)
-	    should_be = [work, coffee]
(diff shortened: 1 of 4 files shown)


CFC_DataCollector/trip_generator/tour_model.py
```diff
@@ -0,0 +1,229 @@
...
+        currNode.times_allowed_two -= 1
+        temp = _make_path(came_from, came_from["%s%s" % (currNode, currNode.count_down+1)], starting_point)
+        return temp + [currNode]
+    else:
+        return [starting_point]
+
+
+
+
+def test_something():
```
shankari: Aren't these already in the Test class? Shouldn't they be removed from here?

CFC_DataCollector/trip_generator/tour_model.py
```diff
@@ -0,0 +1,229 @@
...
+    # return sucessor
+    return sampleFromCounter(place.successors_counter)
+
+def choose_random_successor(successors):
+    num_successors = len(successors)
+    index = random.randint(0, num_successors - 1)
+    return successors[index]
+
+
+def _dfs_search(TM, starting_point):
```
shankari: Aren't there existing graph libraries for python that you could use? I think that Naomi has some pointers...

CFC_DataCollector/tests/tripGenTests/TestTourModel.py
```diff
@@ -0,0 +1,92 @@
+import sys
+sys.path.append("./../../trip_generator")
+import unittest
+from tour_model import Location, TourModel
+
+class TestTourModel(unittest.TestCase):
+
+    def test_something(self):
+        """ Basic Sanity Check """
+        home = Location("home")
```
shankari: How do you specify the times_allowed and the counts for the locations you are creating here?

shankari:
I will merge once you:
- make the automatic jenkins tests pass (currently failing)
- remove the unit tests from the main file
- move the "complicated tour" test to the test file
- add documentation for the functions, in particular around the use of the Count class.

shankari:
Merging, please address documentation nits is subsequent checkin.

PR #158, opened 2015-07-18
Tour model updates 

joshzarrabi:
New tour model, take a look and see what you think. 

I tried to make it as simple as possible, which may have made it more complex to use...
Thanks!


shankari:
I am going to merge this now, but I think that you and I should spend some time talking about the code. Prof. Culler likes to work at the idea level, but I like to work at the code level as well...

joshzarrabi:
Okay!

I know this could probably be better, and I'm open to changing it, but this
is what I have right now.

Thanks!

CFC_DataCollector/trip_generator/tour_model.py
```diff
@@ -1,89 +1,104 @@
...
-
-    def increment_weight_of_successor(self, successor):
-        self.successors_counter[successor] += 1
-
-    def add_n_to_weight_of_successor(self, successor, n):
-        self.successors_counter[sucessor] += n
-
-    def set_weight_of_successor(self, sucessor, weight):
-        self.successors_counter[sucessor] = weight
+    def __init__(self, name, hour, day):
```
shankari: I am not sure whether you want to have the location have the time or the edge have the time. If you think about this as a graph, you have classes for the nodes, but not the edges. But in our discussion, we had assumed that we would have decorations on the edges (in other words, the time that you go from work to lunch is different from the time you go from work to home)

CFC_DataCollector/trip_generator/tour_model.py
```diff
@@ -1,89 +1,104 @@
...
+        self.hour = hour ## An int 0-23 representing the hour 
+        self.name = name ## The name of the place, important for equality
```
shankari: I think that you generally want to use an id instead of a name for equality testing. What happens if the label is changed?

CFC_DataCollector/tests/tripGenTests/TestTourModel.py
```diff
@@ -1,29 +1,68 @@
...
+        work_1_tues = Location("work_1", 8, 1)
+        coffee_tues = Location("coffee", 10, 1)
+        lunch_tues = Location("lunch", 12, 1)
+        work_2_tues = Location("work", 13, 1)
+        home_end_tues = Location("home", 18, 1) 
+        home_start_tues.add_successors({work_1_tues : 1})
+        work_1_tues.add_successors({coffee_tues : 2})
+        coffee_tues.add_successors({lunch_tues : 3})
+        lunch_tues.add_successors({work_2_tues : 4}) ## Completes the cycle
+        work_2_tues.add_successors({home_end_tues : 100})
```
shankari: Why is the last successor with 100 while the others are in numeric order?

PR #159, opened 2015-07-20
Restructure the server code to make logical sense

shankari:
This is not a very general cleanup - just a restructuring of the server code to a version that makes logical sense. Most of it is just moving files around, although I'm doing a little cleanup from time to time if things are too egregiously bad.

The planned structure is in the wiki:
https://github.com/e-mission/e-mission-server/wiki/Module-structure

Important decisions that have been deferred from this change are:
- Before this, we used clients as a catchall for configuring behavior (thresholds, etc), calculating results and displaying them. We need to split those out properly. I have largely moved the client code into the new directory untouched
- Some of the prior work from Mogeng is not properly structured into functions and invocation script, so the rules for "bin" haven't been properly observed. I have a list of these written down and will be cleaning them up as I have time during the weeks to come


shankari:
@joshzarrabi @nephraim @scisley, here's what the change looks like right now. Now I just need to fix all the imports so that the test cases pass, and I'm done for now.

Everybody should do a mega restructure like this at least once in their lives :)

screen shot 2015-07-19 at 6 12 25 pm

shankari:
@Mogeng, I know you don't have time to work on this any more, but given that you did so much work earlier, I thought that you might like to see what the new structure looks like...

e-mission-server-module-structure

shankari:
@sdsingh, @shanthistyle, I am not sure whether you have pending changes in the server code, but in case you do, you should know that I am doing a significant cleanup this weekend.

shankari:
@sfwatergit, forgot to add you to the list of people notified

Mogeng:
That looks great! Definitely the work you guys have done is fantastic and out of my expertise…Congrats on these amazing progress!  I’d like to know the new professes you have made and I am willing to provide my phone as a test phone if needed.

Thanks for letting me know!
Mogeng

shankari:
@joshzarrabi @nephraim @scisley changes complete.
Tests pass.
Merging now.

PR #162, opened 2015-07-31
Tour model

nephraim:
This pull requests includes all the code for clustering data and preparing it for the tour model. 
It also includes updates to the fake trips so that they are written to the trip database and they can be used for testing. 

Changes to the clustering pipeline include using the trip wrapper class, creating a separate file for working with ground truth, and adding a file to turn the clusters into a representation for the tour model. I have also written test files for all the files used in the cluster pipeline. 

I am also including Josh's tour model files but they are subject to change.


shankari:
ok to test

PR #165, opened 2015-08-07
clustering pipeline

nephraim:
In this pull request I've changed the clustering pipeline, which includes similarity.py, featurization.py, representatives.py, and cluster_pipeline.py. I also changed Profile.py to use the new clustering code. I've also written test cases for the clustering files. 

In addition, I am adding a file cluster_groundtruth.py that can be used to evaluate ground truth.

### DIFF
--- a/emission/analysis/modelling/tour_model/featurization.py
+++ b/emission/analysis/modelling/tour_model/featurization.py
@@ -1,53 +1,74 @@
 # Standard imports
+import matplotlib
+matplotlib.use('Agg')
 import matplotlib.pyplot as plt
 import math
 import numpy
 from sklearn.cluster import KMeans
-from kmedoid import kmedoids
 from sklearn import metrics
-from sklearn.metrics.cluster import homogeneity_score, completeness_score
-import sys, os
+import sys
+
+# our imports
+from emission.core.wrapper.trip import Trip, Coordinate
+from kmedoid import kmedoids
 
 """
 This class is used for featurizing data, clustering the data, and evaluating the clustering. 
 
 The input parameters of an instance of this class are:
-- data: Pass in a list of trips, where each entry is a dictionary containing trip_start_location and trip_end_location.
-- colors (optional): a list of the ground truth clusters for the data, in the form of a list of integers where different integers correspond to different clusters. If there is no ground truth, colors defaults to None. 
+- data: Pass in a list of trip objects that have a trip_start_location and a trip_end_location
 
 This class is run by cluster_pipeline.py
 """
 class featurization:
 
-    def __init__(self, data, colors=None):
+    def __init__(self, data):
         self.data = data
+        if not self.data:
+            self.data = []
         self.calculate_points()
-        self.colors = colors
+        self.labels = []
+        self.clusters = None
 
     #calculate the points to use in the featurization. 
     def calculate_points(self):
         self.points = []
-        for i in range(len(self.data)):
-            start = self.data[i]['trip_start_location']
-            end = self.data[i]['trip_end_location']
-            self.points.append([start[0], start[1], end[0], end[1]])
-
+        if not self.data:
+            return
+        for trip in self.data:
+            start = trip.trip_start_location
+            end = trip.trip_end_location
+            if not (start and end):
+                raise AttributeError('each trip must have valid start and end locations')
+            self.points.append([start.lon, start.lat, end.lon, end.lat])
 
     #cluster the data. input options:
-    # - min_clusters (optional): the minimum number of clusters to test for. Must be at least 2. Default to 2.
-    # - max_clusters (optional): the maximum number of clusters to test for. Default to the number of points divided by 2. 
     # - name (optional): the clustering algorithm to use. Options are 'kmeans' or 'kmedoids'. Default is kmeans.
-    # - initial (optional): the way you want to initialize the means in kmeans. Options are 'random', 'k-means++', or an ndarray. Default is k-means++. 
-    def cluster(self, name='kmeans', min_clusters=2, max_clusters=None, initial='k-means++'):
+    # - min_clusters (optional): the minimum number of clusters to test for. Must be at least 2. Default to 2.
+    # - max_clusters (optional): the maximum number of clusters to test for. Default to the number of points. 
+    def cluster(self, name='kmeans', min_clusters=2, max_clusters=None):
         if min_clusters < 2:
-            print 'Must have at least 2 clusters'
             min_clusters = 2
+        if min_clusters > len(self.points):
+            sys.stderr.write('Maximum number of clusters is the number of data points.\n')
+            min_clusters = len(self.points)-1
+        if max_clusters == None:
+            max_clusters = len(self.points)-1
+        if max_clusters < 2:
+            sys.stderr.write('Must have at least 2 clusters\n')
+            max_clusters = 2
+        if max_clusters >= len(self.points):
+            max_clusters = len(self.points)-1
+        if max_clusters < min_clusters:
+            raise ValueError('Please provide a valid range of cluster sizes')
         if name != 'kmeans' and name != 'kmedoids':
             print 'Invalid clustering algorithm name. Defaulting to k-means'
             name='kmeans'
-        if max_clusters == None:
-            max_clusters = len(self.data)/2.0
-            
+        if not self.data:
+            self.sil = None
+            self.clusters = 0
+            self.labels = []
+            return []
         max = -2
         num = 0
         labely = []
@@ -55,6 +76,7 @@ class featurization:
         if name == 'kmedoids':
             for i in range(r):
                 num_clusters = i + min_clusters
+                print 'testing ' + str(num_clusters) + ' clusters'
                 cl = kmedoids(self.points, num_clusters)
                 self.labels = [0] * len(self.data)
                 cluster = -1
@@ -68,11 +90,13 @@ class featurization:
                     num = num_clusters
                     labely = self.labels
         elif name == 'kmeans':
+            import warnings
             for i in range(r):
                 num_clusters = i + min_clusters
-                cl = KMeans(num_clusters, random_state=8, init=initial)
+                cl = KMeans(num_clusters, random_state=8)
                 cl.fit(self.points)
                 self.labels = cl.labels_
+                warnings.filterwarnings("ignore")                
                 sil = metrics.silhouette_score(numpy.array(self.points), self.labels)
                 if sil > max:
                     max = sil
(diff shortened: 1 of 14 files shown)


shankari:
ok to test

shankari:
ok to test

emission/analysis/modelling/tour_model/featurization.py
```diff
@@ -1,150 +1,137 @@
...
-            end_lat = self.data[i]['trip_end_location'][1]
-            end_lon = self.data[i]['trip_end_location'][0]
-            path = [(start_lat, start_lon), (end_lat, end_lon)]
-            mymap2.addpath(path, matcol.rgb2hex(colormap(float(self.labels[i])/self.clusters)))
-        mymap2.draw('./mylabels.html')
-
-
+# Standard imports
+import matplotlib
+matplotlib.use('Agg')
```
shankari: Nit: Do you need this? Are you plotting anything in here?

emission/analysis/modelling/tour_model/featurization.py
```diff
@@ -1,150 +1,137 @@
...
+            sys.stderr.write('No clusters to analyze\n')
+            return
+        if not self.labels:
+            print 'Please cluster before analyzing clusters.'
+            return
+        print 'number of clusters is ' + str(self.clusters)
+        print 'silhouette score is ' + str(self.sil) 
+
+    #map the clusters
+    def map_clusters(self):
```
shankari: map_clusters should be moved out to to a file under emission.analysis.plotting, where it would presumably map the clusters that have been saved to a database? Or can be displayed inline in ipython notebook.

emission/analysis/modelling/tour_model/representatives.py
```diff
@@ -0,0 +1,185 @@
...
+- labels: A list of integers that define the clusters on the data.
+"""
+
+class representatives:
+
+    def __init__(self, data, labels):
+        self.data = data
+        if not self.data:
+            self.data = []
+        self.labels = labels
```
shankari: what are the labels and how do we get them?

emission/analysis/modelling/tour_model/similarity.py
```diff
@@ -1,170 +1,177 @@
...
-        d = R * c
-        if option==True:
-            print d
-        if d <= self.radius:
-            return True
-        return False
+# Standard imports
+import math
+import matplotlib
+matplotlib.use('Agg')
```
shankari: Similarly, I don't think that we should need to include matplotlib here.

emission/analysis/modelling/tour_model/similarity.py
```diff
@@ -1,170 +1,177 @@
...
+            self.bins.pop()
+        newdata = []
+        for bin in self.bins:
+            for b in bin:
+                d = self.data[b]
+                newdata.append(self.data[b])
+        self.newdata = newdata
+
+    #calculate the cut-off point in the histogram
+    def elbow_distance(self):
```
shankari: Can you provide a short writeup of the motivation here? Not as comprehensive as the paper, but a quick idea on what you are doing.

emission/analysis/modelling/tour_model/featurization.py
shankari: I don't think that you have the time to do this, but you should add a TODO comment about it for later.

emission/core/wrapper/trip.py
```diff
@@ -88,7 +114,15 @@ def get_distance(self):
         return cm.calDistance(self.trip_start_location, self.trip_end_location, True)
 
     def save_to_db(self):
-        pass
+        db = edb.get_trip_db()
+        for section in self.sections:
+            section.save_to_db()
+        db.insert({"user_id": self.user_id, "trip_id": self.trip_id, "sections": range(len(self.sections)), "trip_start_datetime": self.start_time,
```
shankari: wait! why are we inserting trips into the database from here? The only place where we should be inserting trips into the database is while processing the data from the phone. I don't see any restructuring of the code that processes data from the phone to use this.... Is this to store representatives? If so, you should use the canonical trips.

emission/core/wrapper/trip.py
```diff
@@ -162,9 +196,13 @@ def _get_datetime(cls, json_segment, coord_key):
 
     def save_to_db(self):
         db = edb.get_section_db()
-        db.update({"_id": self._id},
-                      {"$set": {"distance" : self.distance, "mode" : self.mode, "confirmed_mode" : self.confirmed_mode}},
-                       upsert=False, multi=False)
+        # db.update({"_id": self._id},
+        #               {"$set": {"distance" : self.distance, "mode" : self.mode, "confirmed_mode" : self.confirmed_mode}},
+        #                upsert=False, multi=False)
+        db.insert({"user_id" : self.user_id, "trip_id" : self.trip_id, "distance" : self.distance, "type" : self.section_type,
```
shankari: ditto.

emission/core/wrapper/trip.py
```diff
@@ -284,6 +322,72 @@ def _save_perturbed(self, perturbeds):
...
+
+
+    def save_to_db(self):
+        db = edb.get_fake_trips_db()
+        db.insert({"_id": self._id, "user_id": self.user_id, "trip_id": self.trip_id, "sections": range(len(self.sections)), "trip_start_datetime": self.start_time,
+                "trip_end_datetime": self.end_time, "trip_start_location": self.trip_start_location.coordinate_list(), 
+                "trip_end_location": self.trip_end_location.coordinate_list(), "mode_list": self.mode_list})
+
+    @classmethod
+    def trip_from_json(cls, jsn):
```
shankari: Isn't this very similar to the method in Trip? Why do you have to copy and paste in here? Can't you just call the method in the superclass? Copy-pasting without a strong reason is a set of errors waiting to happen...

emission/simulation/trip_gen.py
```diff
@@ -25,30 +27,33 @@ class Creator:
...
-        self.starting_addresses = []
-        self.ending_addresses = []
-        self.labels = []
+        self.starting_addresses = [ ]
+        self.ending_addresses = [ ]
+        self.labels = [ ]
 
     def get_starting_ending_points(self):
-        city_file = open("CFC_DataCollector/trip_generator/input.json", "r") ## User (Naomi) specifies locations and radius they want
+        city_file = open("emission/simulation/input.json", "r") ## User (Naomi) specifies locations and radius they want
```
shankari: I think that this should be part of the input to the generator instead of a hardcoded file.

emission/tests/analysisTests/TestClusterPipeline.py
```diff
@@ -0,0 +1,82 @@
...
+        trips = db.find()
+        uuids = set()
+        for t in trips:
+            uuids.add(t['user_id'])
+        if len(uuids) == 0:
+            uuids.add(None)
+        self.uuids = uuids
+
+        data = cp.read_data(size=10)
+        self.assertTrue(len(data) == 10 or len(data) == 0)
```
shankari: when would len(data) == 0?

emission/tests/analysisTests/TestClusterPipeline.py
```diff
@@ -0,0 +1,82 @@
...
+        uuid = 'baduuid'
+        data = cp.read_data(uuid=uuid)
+        self.assertTrue(not data)
+        uuid = self.uuids.pop()
+        data = cp.read_data(uuid = uuid)
+        self.assertTrue(len(data) <= db.find({'user_id' : uuid}))
```
shankari: db.find() returns a cursor. I think that you want db.find().count() instead.

emission/tests/analysisTests/TestClusterPipeline.py
```diff
@@ -0,0 +1,82 @@
...
+        data = cp.read_data()
+        self.assertTrue(0 <= len(data) <= db.find().count())
+        self.assertTrue(len(data) == sum)
+        for d in data:
+            self.assertTrue(d.trip_start_location and d.trip_end_location and d.start_time)
+
+    def testRemoveNoise(self):
+        data = cp.read_data(size=100)
+        newdata, bins = cp.remove_noise(None, 200)
+        self.assertTrue(len(newdata) == len(bins) == 0)
```
shankari: nit: a small note "Checking that we don't crash if no data is passed in" would be useful here.

emission/tests/analysisTests/TestClusterPipeline.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertTrue(clusters == 0 or 10 <= clusters <= 15)
+        self.assertTrue(len(labels) == len(newdata))
+        self.assertTrue(cmp(newdata, data) == 0)
+        data, bins = cp.remove_noise(data, 200)
+        clusters, labels, newdata = cp.cluster(data, 20)
+        self.assertTrue(clusters == 0 or 20 <= clusters <= 30)
+
+    def testClusterToTourModel(self):
+        data = cp.cluster_to_tour_model(None, None)
+        self.assertTrue(not data)
```
shankari: This is a negative test case. Don't you need a positive test case as well?

emission/tests/analysisTests/TestFeaturization.py
```diff
@@ -0,0 +1,92 @@
...
+    def testCluster(self):
+        feat = featurization.featurization(self.data)
+        feat.cluster(min_clusters=2, max_clusters=10)
+        self.assertTrue(len(feat.labels) == len(feat.points))
+        self.assertTrue(feat.clusters == len(set(feat.labels)))
+        a = feat.cluster(name='kmeans', min_clusters=5, max_clusters=20)
+        self.assertTrue(len(feat.labels) == len(feat.points))
+        self.assertTrue(feat.clusters == len(set(feat.labels)))
+        b = feat.cluster(name='nonname', min_clusters=5, max_clusters=20)
+        self.assertTrue(a == b)
```
shankari: How can a == b if you use name = 'nonname'?

shankari:
Looks good overall - detailed comments above. I am ok with deferring all changes except the ones related to save_to_db() for trip and section. I'd like to understand that rationale and be very careful before introducing other points where we write to the trip and section databases.

emission/analysis/modelling/tour_model/featurization.py
nephraim: No, I'll take it out.

emission/analysis/modelling/tour_model/featurization.py
nephraim: Okay, I'll add that for later.

emission/analysis/modelling/tour_model/representatives.py
nephraim: the labels are a list of integers that define the clusters on the list of trips. If you run this from cluster pipeline, the labels are created there.

emission/analysis/modelling/tour_model/similarity.py
nephraim: okay, I'll add that

emission/core/wrapper/trip.py
nephraim: This is one of @joshzarrabi's changes.

emission/core/wrapper/trip.py
nephraim: Also @joshzarrabi's change.

emission/core/wrapper/trip.py
nephraim: Also @joshzarrabi

emission/simulation/trip_gen.py
nephraim: I agree. We were thinking of doing that but we decided to leave it this way for now.

emission/tests/analysisTests/TestClusterPipeline.py
nephraim: just in case someone runs this test without running the whole file and the database is empty.

emission/tests/analysisTests/TestClusterPipeline.py
nephraim: Yes. I'll change that.

emission/tests/analysisTests/TestClusterPipeline.py
nephraim: Okay I'll add that.

emission/tests/analysisTests/TestClusterPipeline.py
shankari: Add a comment explaining it?

emission/tests/analysisTests/TestClusterPipeline.py
nephraim: The positive is in the following lines. I can add a few more tests here, but the functionality of each part within this method is tested in TestRepresentatives

shankari:
For the @joshzarrabi pieces, can you remove them from your pull request? I can then merge your changes and he and I can argue about it later.

emission/tests/analysisTests/TestFeaturization.py
nephraim: The method defaults to k-means if an invalid name is given, and a message prints to let the user know.

emission/tests/analysisTests/TestClusterPipeline.py
nephraim: sure

shankari:
Also, in general, for the parts where I asked questions, could you add your response as a comment in the source code? It will help future readers and contributors a lot.

Thanks!

nephraim:
Sure - do you want me to just fix the small things first, like the comments? Also, the parts that Josh did all  have to do with fake trips. If I don't include them, then I can't generate fake trips from my test cases, which means that most likely the tests will run on an empty database. They will pass, but they won't be testing the true functionallity of the code, only that it doesn't crash on an empty database.

shankari:
I think that you should fix the comments in one commit and remove the code from Trip and Section in the second commit. Once those are done, I can merge your changes.

I am not sure that those are still relevant to fake trips - there is a new FakeTrip class that inherits from Trip and seems to incorporate all the Fake trip functionality.

See if removing them actually causes your tests to run with an empty database.

nephraim:
Okay, I'll try it. Last I checked we weren't using the fake trip class but maybe Josh changed it.

emission/analysis/modelling/tour_model/featurization.py
nephraim: Nevermind, I'm using it in map_clusters to get the color map. I can move the import to inside the function if that would be better.

shankari:
And even if these are still relevant, go ahead and remove them anyway. Josh can replace with the FakeTrip class later.

emission/analysis/modelling/tour_model/featurization.py
shankari: I think you can leave it there, but you don't need the .use('Agg'), right?

nephraim: To pass the tests on the server it needs to be there.

emission/core/wrapper/trip.py
nephraim: I think we do this to save the fake trips to the trip database and save the corresponding sections to the section database so that we can use them for testing.

shankari:
To be clear, once you remove the added code from Trip.save_to_db and Section.save_to_db, then I will merge the changes.

nephraim:
I removed it, but when I run the tests, mine run on no data since there is nothing in the trip database. The fake trips that are generated aren't saved to any database without that code. Even when I run it using the fake trip database, there's nothing there. I'm fine as is since the tests pass either way, and then in the future if the tests get fixed so that they don't delete the database my tests will still work.

shankari:
that's fine. Just push the removals and I will talk to Josh about fixing the saves.

PR #170, opened 2015-08-29
Create a base class for all subsequent wrappers

shankari:
And a test for it.


shankari:
jenkins, test this please

PR #189, opened 2015-09-28
Add saving data to the usercache to the pipeline

shankari:
Turn on saving the trips from the past few days in geojson format into the usercache.
This exposed a couple of bugs in the pipeline code, which in turn exposed a bug in the timeseries.
Fixed all bugs, and added the new step to the intake script.

Also added plotting support, and added an ipython notebook that reads the data from the usercache and displays it. This was used to validate the results.


shankari:
jenkins, test this please

shankari:
jenkins, please test this

shankari:
retest this please

PR #190, opened 2015-10-01
Fix the paths in the server to be correct

shankari:


shankari:
The tests are failing because the OTP server hasn't been changed to point to the new one.
Note that this change is to the webapp layer, so it is not captured by the tests anyway.
Merging it so that I can make room for other changes.

PR #192, opened 2015-10-04
Start saving results into the usercache

shankari:


shankari:
jenkins, test this please

PR #199, opened 2015-10-09
a python wrapper for the geocoder i just set up

joshzarrabi:
Here is a wrapper for the API you can call
Be careful though ... its not *nearly* as good as google maps, and is very nit-picky with the exact string it wants for each address. As a failsafe it will call google maps, and print "defaulting"

Anyway, here it is; free unlimited geocoding for the Bay Area and Denver-Boulder

### DIFF
--- a/emission/core/our_geocoder.py
+++ b/emission/core/our_geocoder.py
@@ -1,5 +1,5 @@
 import urllib, urllib2
-from emission.core.wrapper.trip import Coordinate
+from emission.core.wrapper.trip_old import Coordinate
 import json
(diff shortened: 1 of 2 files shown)


joshzarrabi:
I'm not sure why the tests are failing; I only changed one file...

shankari:
Test are failing because of me - i.e. I did a big push to get end-to-end working on android and didn't fix all the tests. Will do so by this weekend.

shankari:
> Be careful though ... its not nearly as good as google maps, and is very nit-picky with the exact string it wants for each address.

I primarily need it for reverse geocoding right now, so this is fine.

PR #206, opened 2015-11-06
Fix all errors in copying values from the usercache

shankari:
In particular:
- Support entries that wrap transitions
- Add support for iOS formatting
- Handle various corner cases with iOS activity detection


shankari:
jenkins, test this please

PR #208, opened 2015-11-06
Sometimes the document is in the error database because of operations…

shankari:
… after the formatting

This means that it has already been saved to the timeseries database, which
means that when we try to fix the error database, the insert will fail with a
DuplicateKeyError, which will again put it into the error database and so on,
ad infinitum.

We want to keep the timeseries database as read-only, so we ignore the entry if
it was already in the timeseries database.


shankari:
jenkins, test this please

PR #212, opened 2015-11-19
added giles integration for stats archiving, added tests

jeffdh5:
Basic integration works now with Giles for client_stats, server_stats, and results_stats.

However, there are some issues to be resolved:
1) When I use the Giles API for inserting new readings to a stream, it replaces the reading rather than adding a new reading. 
2) I have to convert timestamps to integers in order to insert them into Giles; there is some issue in Go with unmarshalling floats. 
I will chat with Gabe soon to figure out how to solve these problems.

### DIFF
--- a/emission/net/api/stats.py
+++ b/emission/net/api/stats.py
@@ -3,7 +3,7 @@ import logging
 import time
 
 # Our imports
-from emission.core.get_database import get_client_stats_db, get_server_stats_db, get_result_stats_db
+from emission.core.get_database import get_client_stats_db, get_server_stats_db, get_result_stats_db, get_client_stats_db_backup, get_server_stats_db_backup, get_result_stats_db_backup
 
 STAT_TRIP_MGR_PCT_SHOWN = "tripManager.pctShown"
 STAT_TRIP_MGR_TRIPS_FOR_DAY = "tripManager.tripsForDay"
@@ -43,13 +43,18 @@ def setClientMeasurements(user, reportedVals):
 # metadata format is 
 def storeClientEntry(user, key, ts, reading, metadata):
   logging.debug("storing client entry for user %s, key %s at timestamp %s" % (user, key, ts))
-  currEntry = createEntry(user, key, ts, reading)
-  # Add the os and app versions from the metadata dict
-  currEntry.update(metadata)
-  response = get_client_stats_db().insert(currEntry)
+  response = None
   
-  return response != None
+  try:
+    currEntry = createEntry(user, key, ts, reading)
+    # Add the os and app versions from the metadata dict
+    currEntry.update(metadata)
+    response = get_client_stats_db().insert(currEntry)
 
+  except Exception as e:
+    logging.debug("failed to store client entry for user %s, key %s at timestamp %s" % (user, key, ts))
+    get_client_stats_db_backup().insert(currEntry)
+  return response != None
 
 # server measurements will call this directly since there's not much point in
 # batching in a different location and then making a call here since it runs on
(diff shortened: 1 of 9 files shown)


emission/tests/netTests/TestStats.py
```diff
@@ -6,10 +6,88 @@
 
 # Our imports
 from emission.core.get_database import get_client_stats_db, get_server_stats_db, get_result_stats_db
+from emission.net.int_service.giles import archiver
 from emission.net.api import stats
 
 logging.basicConfig(level=logging.DEBUG)
 
+class TestArchiver(unittest.TestCase):
```
shankari: Please put this into its own class. It's not a simple two-liner, it's a substantial piece of code, so it makes sense to pull it out rather than sticking it into a file that doesn't even have the same name.

shankari:
jenkins, test this please

shankari:
jenkins, test this please

emission/tests/netTests/TestStats.py
```diff
@@ -6,10 +6,88 @@
...
+  def setUp(self):
+    self.path = '/archiver_test_path'
+    self.archiver = archiver.StatArchiver(self.path)
+
+  def tearDown(self):
+    self.archiver.remove()
+
+  # @TODO: Rewrite this test, it doesn't really test much
```
shankari: How would you rewrite the test?

emission/tests/netTests/TestStats.py
```diff
@@ -35,30 +113,54 @@ def testCreateEntry(self):
...
+    self.assertEquals(type(savedEntries), list)
+    self.assertEquals(len(savedEntries), 2)
+
+
+    # @TODO: Improve this test to be more informative
+    for savedEntry in savedEntries:
+      #print(savedEntry)
+      self.assertEquals(savedEntry['Metadata']['client_app_version'], '2.0.1')
+      self.assertEquals(savedEntry['Metadata']['client_os_version'], '4.3')
+      #self.assertAlmostEqual(savedEntry['Metadata']['reported_ts'], time.time(), places = 0)
```
shankari: Why has all this been commented out?

emission/tests/netTests/TestStats.py
```diff
@@ -35,30 +113,54 @@ def testCreateEntry(self):
...
-    self.assertEqual(get_client_stats_db().find({'ts': currTime}).count(), 1)
+    self.assertEqual(get_client_stats_db().query_tags(), None)
+    
+    success = stats.storeClientEntry("testuser", "testfield", currTime, 0.002, {"metadata_key": "metadata_val"})
+    savedEntries = get_client_stats_db().query_tags()
+    self.assertEquals(type(savedEntries), list)
+    self.assertEquals(len(savedEntries), 1)
+    entry = savedEntries[0]
+    self.assertEquals(entry["Metadata"]["metadata_key"], "metadata_val")
+    #self.assertEqual(get_client_stats_db().find({'user': 'testuser'}).count(), 1)
```
shankari: Can't we query by user any more?

emission/tests/netTests/TestStats.py
```diff
@@ -35,30 +113,54 @@ def testCreateEntry(self):
...
+    success = stats.storeServerEntry("testuser", "GET foo", currTime, 0.002)
+    self.assertEquals(success, True)
+    savedEntries = get_server_stats_db().query_tags()
+    self.assertEquals(type(savedEntries), list)
+    self.assertEquals(len(savedEntries), 1)
+    entry = savedEntries[0]
+    self.assertEquals(entry["Metadata"]["user"], "testuser")
+    #self.assertEqual(get_server_stats_db().find().count(), 1)
+    #self.assertEqual(get_server_stats_db().find({'user': 'testuser'}).count(), 1)
+    #self.assertEqual(get_server_stats_db().find({'ts': currTime}).count(), 1)
```
shankari: Or by time for that matter?

emission/tests/netTests/TestStats.py
jeffdh5: I'm still working on figuring out how to make the queries. But I am under the impression that the plotter application gives us these queries for free.

emission/tests/netTests/TestStats.py
jeffdh5: I mainly wrote that as a temporary mental note. I would add tests that it works with non-integer timestamps.

emission/tests/netTests/TestStats.py
jeffdh5: see above

emission/tests/netTests/TestStats.py
jeffdh5: see above

emission/net/api/stats.py
```diff
@@ -46,22 +46,33 @@ def storeClientEntry(user, key, ts, reading, metadata):
   currEntry = createEntry(user, key, ts, reading)
   # Add the os and app versions from the metadata dict
   currEntry.update(metadata)
-  get_client_stats_db().insert(currEntry)
+  response = get_client_stats_db().insert(currEntry)
```
shankari: Add some sanity checks here. Having the instrumentation server go down or be misconfigured should not affect the main operation of the server. Adding a try/catch block over the entire method should be good enough, but you can also look at alternative strategies on the internet.

Please add a negative test case for this as well.

shankari: Thanks to Omid for pointing this out!

shankari: Also, if there is an exception, maybe continue logging to the old method (in mongodb) so that we can repair when we are done.

jeffdh5: What is a negative test case? test case for when this fails?

shankari: Yes. You make it fail, maybe by giving a bad/invalid input, or by
reconfiguring the URL within the test, and then ensuring that nothing is
raised. Then you can restore the original URL.

shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #215, opened 2015-11-30
Three minor fixes; two in production

shankari:


shankari:
jenkins, test this please

PR #218, opened 2015-12-15
Change the expected zip codes

shankari:
I don't know where the original ground truth for the zip codes came from, but
it is pretty clear that both of them match to 94704.

http://www.melissadata.com/lookups/latlngzip4.asp?lat=37.871758&lng=-122.259769
http://www.melissadata.com/lookups/latlngzip4.asp?lat=37.8697&lng=-122.2584


shankari:
jenkins, test this please

PR #220, opened 2015-12-18
Check in a bunch of pending fixes found while doing the full-fledged data collection

shankari:
Next up: changes made on the server during deployment...


shankari:
jenkins, test this please

PR #222, opened 2015-12-31
Latest changes

jeffdh5:


emission/net/api/stats.py
```diff
@@ -59,10 +59,16 @@ def storeClientEntry(user, key, ts, reading, metadata):
...
   try:
     response = get_client_stats_db().insert(currEntry)
+    if response == None:
+      get_client_stats_db_backup().insert(currEntry)
   except Exception as e:
     logging.debug("failed to store client entry for user %s, key %s at timestamp %s" % (user, key, ts))
     logging.debug("exception was: %s" % (e))
     get_client_stats_db_backup().insert(currEntry)
+
+  if success == None:
```
shankari: where is the variable "success" defined? this should just fail immediately, right? Is there a test case for this?

emission/net/int_service/giles/archiver.py
```diff
@@ -34,16 +36,20 @@ def __init__(self, collection):
...
+        
+        try:
+            stat = str(entry['stat'])
+            user_uuid = str(entry['user'])
+            # TODO: Support more precise timestamps
+            # Giles has some problem unmarshalling floats
+            #client_ts = int(entry['ts'])
+            client_ts = int(entry['ts'])
+            reading = float(entry['reading'])
+        except Exception as e:
```
shankari: Can you explain the rationale behind the try/catch here? If something is wrong in the formatting, we don't want it to fail silently here, right?

emission/net/int_service/giles/timeout.py
```diff
@@ -0,0 +1,24 @@
+# Code comes from here: http://stackoverflow.com/questions/2281850/timeout-function-if-it-takes-too-long-to-finish
+
+from functools import wraps
+import errno
+import os
+import signal
+
+def timeout(seconds=10, error_message=os.strerror(errno.ETIME)):
```
shankari: I am not sure we need to do this. Isn't there a built-in timeout in the requests library? Timing out network connections is a fairly common process...

emission/net/int_service/giles/archiver.py
jeffdh5: That's true. I added a try/catch in the archiver so that I could write the negative test case. That way the archiver itself has ways to deal with failure independently of how it's being called.

shankari:
jenkins, test this please

shankari:
jenkins, test this please

shankari:
> "Timezone": "America/Los_Angeles",
This is incorrect. The timezone is not always "America/Los_Angeles".

It is certainly not the case for server stats - the server is running in the cloud and is configured to UTC. It is not guaranteed to be the timezone for the client, since, as you know now, people can travel with their phones to places that are not in the pacific time zone. We currently don't include the timezone as part of the metadata, so you can either change the client code to include it, or you can just default it to UTC as well.

shankari:
jenkins, test this please

shankari:
obsoleted by reworked stats code

PR #224, opened 2016-01-03
Support retrieving the timeline for a particular day

shankari:
Since we no longer send all the data to the phone app, we need to add a method
that can return the timeline for a particular day. We can use this method to
read the data for days that are not in the cache.

This involved some minor restructuring of the geojson code generation code and
the timeline code to support reading values by datetime instead of by
timestamp.

We might want to consider supporting only datetime going forward because it
fixes the issue with missing trips that are in a different timezone and so
don't fall within the correct timestamp range.

https://github.com/e-mission/e-mission-server/blob/master/emission/net/usercache/builtin_usercache_handler.py#L189


shankari:
This fixes the first issue on the server side.
The second issue will be fixed once all the clients have been updated.

PR #227, opened 2016-01-26
Wrapper for common trips class

joshzarrabi:
Here is my wrapper class for common trips
Basically what I had before but with better documentation and names


shankari:
jenkins, test this please

shankari:
glad you made a new PR - makes it much easier to review overall.
Will take a look at this once I get in (and once the tests pass).

emission/core/wrapper/common_place.py
```diff
@@ -0,0 +1,12 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class CommonPlace(ecwb.WrapperBase):
+    props = {"coords" : ecwb.WrapperBase.Access.WORM, # geojson coordinates that represent the place
+             "address" : ecwb.WrapperBase.Access.WORM, # address that represents the place
+    }
```
shankari: Don't you need a list of place IDs that are associated with the common place?
Also, you should have probabilities of the arrival time, departure time, duration, etc, right?

emission/core/wrapper/tour_model.py
```diff
@@ -0,0 +1,9 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class TourModel(obecwb.WrapperBase):
+
+    props = {"user_id" : ecwb.WrapperBase.Access.WORM} # user_id of the E-Missions user the graph represnts
```
shankari: Doesn't the tour model need any other information? like the list of common trips and common places? How am I supposed to retrieve the list of common places and trips from the tour model? Or if the intention is to just retrieve them directly, then why do we need a tour model wrapper?

Adding the decorators will make it much more clear how we plan on stitching together the objects.

emission/core/wrapper/common_trip.py
```diff
@@ -0,0 +1,14 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class CommonTrip(ecwb.WrapperBase):
+    props = {"start_loc" : ecwb.WrapperBase.Access.WORM, # Lat/lng start point of trip
+             "end_loc" : ecwb.WrapperBase.Access.WORM, # Lat/lng end point of trip
```
shankari: geojson is a tad more than lat/lng

emission/core/wrapper/common_trip.py
```diff
@@ -0,0 +1,14 @@
...
+             "trips" : ecwb.WrapperBase.Access.WORM, # List of trip_ids that are associated with this common trip
+             "probabilites" : ecwb.WrapperBase.Access.WORM  # a matrix that represents the probabilites for edge
```
shankari: How do I access the probability distribution for a particular field (e.g. probability distribution for start_time)? Some documentation on that would be good.

joshzarrabi:
I added some decorations and tests, I know its not much but it took me a little bit to figure out the system/being sick. If this is on the right track I'll do much more tomorrow/Friday

emission/core/wrapper/common_place.py
```diff
@@ -0,0 +1,17 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class CommonPlace(ecwb.WrapperBase):
+    props = {"coords" : ecwb.WrapperBase.Access.WORM, # geojson coordinates that represent the place
+             "edges" : ecwb.WrapperBase.Access.WORM, # set of CommonTrips that include this location
+             "successors" : ecwb.WrapperBase.Access.WORM, # set of CommonPlaces connected to this location
```
shankari: Just so that I understand, we don't technically need the successors, right? We can just iterate through the edges and find the destination node for each of them. So this is a convenience function that is duplicated. Correct?

joshzarrabi: Yeah that's right

emission/core/wrapper/common_trip.py
```diff
@@ -0,0 +1,14 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class CommonTrip(ecwb.WrapperBase):
+    props = {"start_loc" : ecwb.WrapperBase.Access.WORM, # _id of common_place of starting point
+             "end_loc" : ecwb.WrapperBase.Access.WORM, # _id of common_place of ending point
+             "trips" : ecwb.WrapperBase.Access.WORM, # List of trip_ids that are associated with this common trip
```
shankari: Why isn't there a similar list of places associated with the CommonPlace?

joshzarrabi: We didn't have it originally because there is no direct analog in mongo

emission/core/wrapper/common_trip.py
```diff
@@ -0,0 +1,14 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class CommonTrip(ecwb.WrapperBase):
+    props = {"start_loc" : ecwb.WrapperBase.Access.WORM, # _id of common_place of starting point
```
shankari: Do we assume that every common trip will be between two common places? Not an unreasonable assumption, but think through whether that is always true.

emission/storage/decorations/common_place_queries.py
```diff
@@ -0,0 +1,83 @@
...
+    #     "successors" : common_place.successors,
+    #     "user_id" : common_place.user_id
+    #     })
+
+def get_common_place_from_db(_id):
+    db = edb.get_common_place_db()
+    json_obj = db.find_one({"_id" : _id})
+    return make_common_place_from_json(json_obj)
+
+def make_common_place_from_json(json_obj):
```
shankari: why do you even need this function? Your json_obj is already a dict, right? You should be able to just convert it directly to a CommonPlace.

emission/core/wrapper/common_trip.py
shankari: I am not sure what you mean by this. Do you mean that are no place objects? There definitely are, that's what the Place wrapper class is about.

joshzarrabi: Yeah sorry I meant Naomi's code didn't include that when she packaged the info, I can add that

emission/storage/decorations/common_place_queries.py
joshzarrabi: Okay

emission/core/wrapper/common_place.py
```diff
@@ -0,0 +1,16 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class CommonPlace(ecwb.WrapperBase):
+    props = {"coords" : ecwb.WrapperBase.Access.WORM, # geojson coordinates that represent the place
+             "successors" : ecwb.WrapperBase.Access.WORM, # set of CommonPlaces connected to this location
+             "user_id" : ecwb.WrapperBase.Access.WORM,
+             "common_place_id" : ecwb.WrapperBase.Access.WORM
```
shankari: why not just use _id? The good thing about _id is that mongodb will autopopulate it. How were you planning on generating a unique ID that would be better than mongo's version?

emission/core/wrapper/common_trip.py
shankari: In trip and place, I've used XXX_loc for the geojson. So if we followed that convention, then this would be start_point or start_common_place, and start_loc would be an additional field that would store the geojson representation of the point to support mongo geojson queries.

emission/core/wrapper/common_place.py
joshzarrabi: I was trying to make it in a way I could recreate it

emission/core/wrapper/common_trip.py
```diff
@@ -0,0 +1,20 @@
...
+             "end_loc" : ecwb.WrapperBase.Access.WORM, # _id of common_place of ending point
+             "trips" : ecwb.WrapperBase.Access.WORM, # List of trip_ids that are associated with this common trip
+             "probabilites" : ecwb.WrapperBase.Access.WORM,  # a matrix that represents the probabilites for edge
```
shankari: I'd really like to understand how this matrix is structured, and how I can retrieve the probability of an edge, etc.

emission/core/wrapper/common_place.py
shankari: Can you explain further? When would you want to/need to recreate the ID?

emission/core/wrapper/tour_model.py
```diff
@@ -0,0 +1,14 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class TourModel(ecwb.WrapperBase):
```
shankari: I am not sure if TourModel needs to be a wrapper class in its current form. Maybe you need to put the common places and common trips into it?

emission/simulation/trip_gen.py
```diff
@@ -77,28 +77,35 @@ def get_trips_from_a_to_b(self):
...
                 print "server error"
+                pass   
+            except:
                 pass
 
 def save_trip_to_db(trip):
+    print "saving trip to db"
+    print trip.user_id
     db = edb.get_trip_db()
+    print "RHRHRH"
```
shankari: Remove before final merge

emission/storage/decorations/common_place_queries.py
```diff
@@ -0,0 +1,78 @@
...
+def make_new_common_place(user_id, coords, successors):
+    props = {
+        "user_id" : user_id,
+        "coords" : coords,
+        "common_place_id" : "%s%s" % (user_id, coords["coordinates"]),
+        "successors" : successors
+    }
+    return ecwcp.CommonPlace(props)
+
+def make_common_place(props):
```
shankari: what is the difference between this and `make_common_place_from_json`?

shankari:
Looking much better overall.
I haven't reviewed the decorators yet, will do so tomorrow night.

joshzarrabi:
Okay.
Good night!

shankari:
Just to clarify - this is not yet ready to merge.
See my comments from earlier today.

joshzarrabi:
I know.
It's not passing my own tests

emission/core/wrapper/common_place.py
joshzarrabi: I use it in line 73 of common_trip_queries.py when I'm looking for successors 
        _id = esdctp.make__id(user_id, place_id, suc)
        trip = esdctp.get_common_trip(_id)

but I can just use a mongo query, I don't need a special _id

emission/storage/decorations/common_place_queries.py
joshzarrabi: This is just an easy way to create a new trip without having to write a new props, 
the make_common_place_from_json is for when we get it out of mongo

shankari: I understand what `make_new_common_place` is. My comment is for `make_common_place`.
What is the difference between `make_common_place` and `make_common_place_from_json`?

emission/core/wrapper/common_place.py
shankari: Maybe my questions was not quite clear. I am asking why you need to construct a unique ID instead of using the Mongo generated id? Is there a particular reason for this?

emission/storage/decorations/common_place_queries.py
joshzarrabi: Oh yeah youre right they are the same, we only need one

PR #234, opened 2016-02-24
Fixes that we made on the production server to fix post-deployment issues

shankari:


shankari:
jenkins, test this please

PR #238, opened 2016-03-01
Added clustering to pipeline

joshzarrabi:
Here is the change

### DIFF
--- a/emission/net/usercache/builtin_usercache_handler.py
+++ b/emission/net/usercache/builtin_usercache_handler.py
@@ -169,9 +169,9 @@ class BuiltinUserCacheHandler(enuah.UserCacheHandler):
         """
         tour_model = esdtmpq.get_tour_model(self.user_id)
         uc = enua.UserCache.getUserCache(self.user_id)
-        logging.debug("Adding common trips for user_id %s" % self.user_id)
-        uc.putDocument("common_trips-%s" % tour_model.user_id, tour_model)
-        valid_key_list = ["common_trips-%s" % tour_model.user_id]
+        logging.debug("Adding common trips for day %s" % str(pydt.date.today()))
+        uc.putDocument("common_trips-%s" % str(pydt.date.today()),  tour_model)
+        valid_key_list = ["common_trips-%s" % str(pydt.date.today())]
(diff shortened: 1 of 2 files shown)


emission/net/usercache/builtin_usercache_handler.py
```diff
@@ -160,6 +162,21 @@ def storeViewsToCache(self):
...
 
+    def storeCommonTripsToCache(self):
+        """ 
+        Determine which set of common trips to send to the usercache. 
+        As of now we will run the pipeline on the full set of data and send that up
+        """
+        tour_model = esdtmpq.get_tour_model(self.user_id)
+        uc = enua.UserCache.getUserCache(self.user_id)
+        logging.debug("Adding common trips for user_id %s" % self.user_id)
+        uc.putDocument("common_trips-%s" % tour_model.user_id, tour_model)
```
shankari: Two questions here:
- There is only one %s, but there are two fields. Are you sure that this works? Please add the log file lines that show that this works
- why put the user_id into the common trips given that the user cache is user specific? The phone does not actually know the user's user_id so it would be hard to retrieve this document on the phone.

emission/net/usercache/builtin_usercache_handler.py
```diff
@@ -184,7 +201,7 @@ def get_obsolete_entries(self, uc, valid_key_list):
         # lexicographic ordering, but at the same time, this seems much easier
         # and safer to deal with.
         logging.debug("curr_key_list = %s, valid_key_list = %s" % 
-	    (curr_key_list, valid_key_list))
+        (curr_key_list, valid_key_list))
```
shankari: Revert irrelevant change by restoring indentation.

emission/net/usercache/builtin_usercache_handler.py
shankari: Ah, I see that the key just includes the user_id, the second parameter is the object that you are inserting. But you do have to remove the user_id because as I said, it is not available on the phone.

joshzarrabi: Okay

PR #241, opened 2016-03-10
New common trips

joshzarrabi:
Here are the changes, fixed the issue, made it a bit easier to make many trips, updated the readme.


shankari:
How did you test this? Please copy and paste test script command lines and relevant parts of the output.

shankari:
In particular, I don't think that this was the error that I encountered. You're adding a try/catch around an array access. I got an `AssertionError`. I would like you to test with loading a single day's worth of data, similar to the instructions in my email. Otherwise, this is going to crash the entire pipeline the very first time that we have a new user install the app.

joshzarrabi:
Hi that isn't the relevant change.
The one that makes it work is in similarity.py

shankari:
Can you please add the test results as I  requested earlier?

joshzarrabi:
$ python bin/purge_database_json.py

collections = [u'system.indexes', u'Stage_usercache', u'Stage_pipeline_state', u'Stage_place', u'Stage_common_place', u'Stage_stop', u'
Stage_timeseries', u'Stage_trip_new', u'Stage_section_new', u'Stage_common_trips']
Skipping system collection system.indexes
Dropping collection Stage_usercache
Dropping collection Stage_pipeline_state
Dropping collection Stage_place
Dropping collection Stage_common_place
Dropping collection Stage_stop
Dropping collection Stage_timeseries
Dropping collection Stage_trip_new
Dropping collection Stage_section_new
Dropping collection Stage_common_trips
```

shankari:
Awesome! Thanks for putting in the testing and a useful commit message. It really does help.

Issue #247, opened 2016-03-14
Trips are looked up twice from the database in the new tour model code


shankari:
First, we get all the trips from the database.

```
    if not old:
        logging.debug("not old")
        db = edb.get_trip_new_db()
        trips = db.find({"user_id" : uuid})
```

Then, we convert them to ids

```
return [trip["_id"] for trip in trips]
```

And then we look up the trips again individually in the featurization constructor.

It looks like for the old version, we pass the trips around directly. Why can't we do that again? Why do we need to convert them to ids and then re-read the trips for the IDs?

Currently, we open the database for every read of a single trip, 
```
def get_trip(trip_id):
    """
    Returns the trip for specified trip id.
    :rtype : emission.core.wrapper.Trip
    """
    return ecwt.Trip(edb.get_trip_new_db().find_one({"_id": trip_id}))
```
and opening the database multiple times in quick succession causes a crash.

```
Traceback (most recent call last):
  File "bin/intake_stage.py", line 48, in <module>
    esdtmq.make_tour_model_from_raw_user_data(uuid)
  File "/mnt/e-mission/e-mission-server/emission/storage/decorations/tour_model_queries.py", line 50, in make_tour_model_from_raw_user_data
    list_of_cluster_data = eamtmcp.main(user_id, False)
  File "/mnt/e-mission/e-mission-server/emission/analysis/modelling/tour_model/cluster_pipeline.py", line 106, in main
    n, labels, data = cluster(data, len(bins), old=old)
  File "/mnt/e-mission/e-mission-server/emission/analysis/modelling/tour_model/cluster_pipeline.py", line 83, in cluster
    feat = featurization.featurization(data, old=old)
  File "/mnt/e-mission/e-mission-server/emission/analysis/modelling/tour_model/featurization.py", line 33, in __init__
    self.calculate_points()
  File "/mnt/e-mission/e-mission-server/emission/analysis/modelling/tour_model/featurization.py", line 48, in calculate_points
    trip = esdtq.get_trip(trip)
  File "/mnt/e-mission/e-mission-server/emission/storage/decorations/trip_queries.py", line 43, in get_trip
    return ecwt.Trip(edb.get_trip_new_db().find_one({"_id": trip_id}))
  File "/mnt/e-mission/e-mission-server/emission/core/get_database.py", line 183, in get_trip_new_db
    current_db = MongoClient().Stage_database
[...]

PR #248, opened 2016-03-14
Pass around trip objects in the new version of the cluster pipeline a…

shankari:
…s well

In the old version of the cluster pipeline, we used to read the trip objects
and pass them around. In the new version of the cluster pipeline, this was
changed so that we passed around the IDs instead. This meant that we had to go
back to the database to read the trips over and over as we can see from this
pull request.

This was a particular concern while reading many trips in quick succession in
the featurization step. Since we do not have an index on the trip collection,
this caused the database to be very slow and basically stop responding.

So I went back to passing trip objects around. Now it runs fine, even for
people with a large number of trips, like me and Tom.

This is a fix for https://github.com/e-mission/e-mission-server/issues/247

### DIFF
--- a/emission/analysis/modelling/tour_model/cluster_pipeline.py
+++ b/emission/analysis/modelling/tour_model/cluster_pipeline.py
@@ -63,7 +63,7 @@ def read_data(uuid=None, size=None, old=True):
                 if len(data) == size:
                     break
         return data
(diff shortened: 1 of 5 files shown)


shankari:
@joshzarrabi fixed the trip re-read the issue myself since it wasn't working on production without the change.

PR #249, opened 2016-03-15
Comment out all instances to the crontab in the recommender pipeline

shankari:
- we are not actually using the recommender pipeline
- we plan to replace the crontab with a scheduling pipeline contributed by NREL
- using crontab is messing up other jenkins builds

So let's just comment out all uses of the crontab, and all associated tests for now


shaneknapp:
lgtm

shankari:
jenkins, test this please

Issue #256, opened 2016-04-02
Heatmaps are impossibly slow


shankari:
Even heatmaps that don't return a lot of data, such as "all routes within cal" return take noticeable time to run. For example, the query to get the routes in cal takes over 3 minutes.

```
2016-04-02 20:05:07,093:DEBUG:140049096890112:START GET /result/heatmap/pop.route/cal
2016-04-02 20:08:17,305:DEBUG:140049096890112:END GET /result/heatmap/pop.route/cal  190.212227106
```

The log for the call shown above, highlighting the areas which take a lot of time, is:

```
2016-04-02 20:05:07,093:DEBUG:140049096890112:START GET /result/heatmap/pop.route/cal
2016-04-02 20:05:07,093:DEBUG:140049096890112:Filtering values for range 1458975600000 -> 1459580400000
2016-04-02 20:05:07,094:INFO:140049096890112:About to query for 2016-03-26 07:00:00 -> 2016-04-02 07:00:00 in [[-122.267443, 37.864693], [-122.250985, 37.880687]]
2016-04-02 20:05:14,004:DEBUG:140049096890112:result cursor has 656924 entries
2016-04-02 20:08:04,744:DEBUG:140049096890112:result list has 656924 entries
2016-04-02 20:08:04,746:DEBUG:140049096890112:Converted datetime range 2016-03-26 07:00:00 -> 2016-04-02 07:00:00 to timestamp range 1459001305.45 -> 1459623320.87
2016-04-02 20:08:04,876:DEBUG:140049096890112:len(places) = 257, len(trips) = 214
...
2016-04-02 20:08:17,305:DEBUG:140049096890112:END GET /result/heatmap/pop.route/cal  190.212227106
```

shankari:
I think that an immediate fix is pretty obvious. I can do it in ~ 2 mins, or you can figure it out and do it.
If you don't have the fix in by tomorrow morning, I'm just going to make the change...

shankari:
Here are the logs for the walking trips, which took roughly  5 mins.

```
2016-04-02 20:26:32,840:DEBUG:140048976553728:START GET /result/heatmap/pop.route/commute/walking
2016-04-02 20:26:32,841:DEBUG:140048976553728:Filtering values for range 1458975600000 -> 1459580400000
2016-04-02 20:26:32,841:INFO:140048976553728:About to query for 2016-03-26 07:00:00 -> 2016-04-02 07:00:00
[...]

PR #257, opened 2016-04-04
fixed major timing issue

joshzarrabi:
Stopped converting cursor to list and cut out a large time burden
This is much faster.
According to my time measurements the geojson conversion is still slow, I'll look into that now but its hard to tell on my data set, this might be enough

### DIFF
--- a/emission/storage/decorations/timeline.py
+++ b/emission/storage/decorations/timeline.py
@@ -49,6 +49,8 @@ def get_timeline(user_id, start_ts, end_ts):
 
     return Timeline(places, trips)
 
+
+
 def get_aggregate_timeline_from_dt(start_dt, end_dt, box=None):
     import emission.core.get_database as edb
     import emission.core.wrapper.entry as ecwe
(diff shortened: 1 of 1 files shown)


shankari:
Testing?
In particular, I don't think that you can use the cursor directly like that, although I could be mistaken.

joshzarrabi:
It works

shankari:
I checked in big new datasets into the `data-collection-eval` repo. Use git lfs to check them out.
https://git-lfs.github.com/

shankari:
I'd like some more detail than "it works".
what did you do to test it? what is the impact on the time?
some copy-pasted traces would be good.

joshzarrabi:
74910 2016-04-04 14:35:31,144:DEBUG:START GET /result/heatmap/pop.route/commute/walking                                                                              
74911 2016-04-04 14:35:31,144:DEBUG:Filtering values for range 1325404800000 -> 1459666800000
74912 2016-04-04 14:35:31,146:INFO:About to query for 2012-01-01 00:00:00 -> 2016-04-03 00:00:00
74913 2016-04-04 14:35:31,438:DEBUG:result cursor has 10275 entries
74914 2016-04-04 14:35:31,496:DEBUG:Converted datetime range 2012-01-01 00:00:00 -> 2016-04-03 00:00:00 to timestamp range 1448640163.72 -> 1449037798.53
74915 2016-04-04 14:35:31,595:DEBUG:len(places) = 107, len(trips) = 107

shankari:
This is a fix for https://github.com/e-mission/e-mission-server/issues/256

emission/storage/decorations/timeline.py
```diff
@@ -61,12 +63,12 @@ def get_aggregate_timeline_from_dt(start_dt, end_dt, box=None):
         logging.info("About to query for %s -> %s in %s" % (start_dt, end_dt, box))
     result_cursor = edb.get_timeseries_db().find({"data.local_dt": {"$gte": start_dt, "$lte": end_dt}}).sort("metadata.write_ts")
     logging.debug("result cursor has %d entries" % result_cursor.count())
-    result_list = list(result_cursor)
-    logging.debug("result list has %d entries" % len(result_list))
```
shankari: You removed this line, but then we don't have an idea of how long it takes to access the cursor (e.g. to get the cursor count). Please restore a similar line in your next commit.

PR #262, opened 2016-04-12
Common trips update

joshzarrabi:
Added a list of places to the common trips object.


shankari:
It looks like this includes two changes that have already been merged (restored timing information and converted result_counter.count()). Did you pull from upstream?

joshzarrabi:
I know what to do to fix this,

joshzarrabi:
Yeah I pulled too late

PR #263, opened 2016-04-13
Common trips new

joshzarrabi:
Here are the the changes, sorry the old commit is still there but it doesn't look like it's actually showing up on the diff.


emission/analysis/modelling/tour_model/cluster_pipeline.py
```diff
@@ -40,17 +40,15 @@
 #read the data from the database. 
 def read_data(uuid=None, size=None, old=True):
     data = []
-    db = edb.get_trip_db()
+    trip_db = edb.get_trip_db()
     if not old:
-        logging.debug("not old")
```
shankari: Nice cleanup!

emission/analysis/modelling/tour_model/representatives.py
```diff
@@ -142,6 +143,16 @@ def cluster_dict(self):
         for i in range(self.num_clusters):
             a = {'sections' : self.clusters[i]}
             self.tour_dict[i] = a
+        for i in range(self.num_clusters):
+            start_places = []
+            end_places = []
+            for t in self.tour_dict[i]["sections"]:
```
shankari: I'm confused here. The trips have start and end places, not sections. Sections have start and end stops. Or is that just legacy that the dict is called "sections"?

emission/tests/analysisTests/TestClusterPipeline.py
```diff
@@ -1,171 +0,0 @@
-import unittest
-import emission.core.get_database as edb
-import emission.analysis.modelling.tour_model.cluster_pipeline as cp
-import emission.simulation.trip_gen as tg
-import uuid as uu
-
-class ClusterPipelineTests(unittest.TestCase):
```
shankari: Can you comment more on why you had to remove these tests? I know you said that they weren't valid any more, but I'd like to understand why, and to see how much effort it would be to port them over.

shankari:
I have a couple of high level questions:
- One of the big goals was to use the start and end places for binning instead of the first and last points in the trip. This will fix the issue with the first point of the trip being different from the start place and ensure that we don't have spurious common places that are ~ 200m away from the actual common places (e.g. Soda and SDH). I don't see where that change has happened. Can you please let me give me file and line #?
- I see the 'start_places' and 'end_places' arrays being tracked, but I don't see where they are represented in either `common_trips` or `common_places`. I thought that the plan was to have each `common_place` have a list of the places that map to it, just like each `common_trip` has a list of trips that map to it. Can you clarify?

emission/analysis/modelling/tour_model/representatives.py
joshzarrabi: It's legacy

joshzarrabi:
I'll fix everything and push changed today

emission/storage/decorations/common_place_queries.py
```diff
@@ -61,16 +73,25 @@ def create_places(list_of_cluster_data, user_id):
...
+
+        if end_loc_str not in places_dct:
+            places_dct[end_loc_str] = dct["end_places"]
+
     clear_existing_places(user_id)
     logging.debug("After creating map, number of places is %d" % len(places_to_successors))
     for loc_str in places_to_successors.iterkeys():
         start = make_new_common_place(user_id, gj.loads(loc_str))
+        logging.debug("Adding %d places for this place" % len(places_dct[loc_str]))
+        start.places = places_dct[loc_str]
```
joshzarrabi: This is where i add the list of places

shankari:
You haven't yet responded to the question about the obsolete tests and what you are replacing them by.

emission/tests/analysisTests/TestClusterPipeline.py
shankari: Josh, comments on the test removal?

joshzarrabi: Sorry I'll do this by Sunday

shankari: Note that I have made substantial changes to the codebase in terms of
storing trips and places. Please remember to pull and switch to a new
branch before making any more changes.
https://github.com/e-mission/e-mission-server/pull/265

Shankari

joshzarrabi: Okay!

emission/analysis/modelling/tour_model/similarity.py
```diff
@@ -41,10 +41,10 @@ def __init__(self, data, radius, old=True):
                 # print "a is %s" % a
                 t = a
                 try:
-                    start_lon = t.start_loc["coordinates"][0]
-                    start_lat = t.start_loc["coordinates"][1]
-                    end_lon = t.end_loc["coordinates"][0]
-                    end_lat = t.end_loc["coordinates"][1]
+                    start_lon = t.start_place.location["coordinates"][0]
```
shankari: Don't you need to make a similar change in representatives?

                    points[0].append(c.data.start_loc["coordinates"][1])
                    points[1].append(c.data.start_loc["coordinates"][0])
                    points[2].append(c.data.end_loc["coordinates"][1])
                    points[3].append(c.data.end_loc["coordinates"][0])

shankari: I added some logging statements, and it is pretty clear that the one in representatives is the one being called. Which makes sense because t.start_place is an object ID, so I would expect t.start_place.location to fail.

How did you test this to confirm that it works?

joshzarrabi: I did not sorry... I'm writing tests now...

PR #266, opened 2016-04-23
Add support for a new stage that cleans and resamples the data for display

shankari:
In particular, it:
* creates a completely parallel "cleaned" hiearchy of trips, places, sections and stops, which
  * skips spurious, zero section trips
  * merges the places on both sides of the trips
   (note that with our current implementation, these had to happen after the section detection, so in a separate pass)
* makes the points uniform between platforms and sampling schemes
  * removes outliers from the points
  * resamples them at a fixed (30 second) interval
* adds a bunch of useful decorations
  * adds distances and speeds to the points
  * adds the mode and the section to each point
  * sums up the distances to a section distance
  * sums up the section distances to a trip distance
  * populates display names, osm type and osm ID for each place


shankari:
The tests are failing because the version of scipy jenkins server has not been updated.

I am using the scipy interpolate function with fill_value = 'interpolate'.
https://github.com/e-mission/e-mission-server/pull/266/files#diff-a288fba96dafbe078c864c654382686dR317

```
lat_fn = spi.interp1d(x=loc_df.ts, y=loc_df.latitude, bounds_error=False,
                      fill_value='extrapolate')
```

However, that is only available in scipy version 0.17 or later
http://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d

The tests are passing for me locally, and I want to move on to changing the UI, so I'm going to merge this pull request anyway.

PR #267, opened 2016-04-25
Ionic-based version of the webapp

shankari:
This works end-to-end and shows the heatmap, retrieved from a HTTP POST call.

We need to figure out a packaging strategy to share code between the phone and the server, but this is a good first step...


shankari:
Failures are due to the 'extrapolate' issue again.
So I'm going ahead and merging.

PR #280, opened 2016-06-10
Js testing

adambrown13:
Adding some config files that I missed last commit. In order to run all the test, while in the webapp directory, run the command "karma start my.conf.js"


shankari:
Is this really the correct approach to take?
If a lot of the files are related to the karma module, does it just make sense to update the README to specify installation via npm instead?

adambrown13:
The my.conf.js can be a bit tricky, but the node modules definitely can be built with the npm!

shankari:
config should definitely be checked in. I generally don't like to check in unmodified binary code - much easier to pull it from somewhere as part of the build process.

shankari:
Just to confirm, I am waiting for this to be resolved before I merge.

adambrown13:
I took out all of the node_modules except for angular_mocks. The rest of the node_modules compile on the npm installation. If I should add the npm installation to the read instructions, I will

shankari:
> If I should add the npm installation to the read instructions, I will

Yes, you should.

shankari:
Why do you need a new `webapp/www/lib/app.js`? There's already an existing `webapp/www/js/app.js`...

shankari:
Looks much better now.

shankari:
For the record, this is what I mean about unmodified binary code.
https://help.github.com/articles/what-is-my-disk-quota/#external-dependencies

shankari:
If I run 

```
$ npm install karma --save-dev
```
then `karma` is not in my path. Need to update the README to reflect that.

Also, when I run `./node_modules/karma/bin/karma start my.conf.js`, and navigate to 
http://localhost:8080/
I get the error

```
21 06 2016 12:31:00.326:INFO [Firefox 47.0.0 (Mac OS X 10.11.0)]: Connected on socket /#28Ai1-RCtiaPUKltAAAA with id manual-8592
Firefox 47.0.0 (Mac OS X 10.11.0) LOG: 'WARNING: Tried to load angular more than once.'

Firefox 47.0.0 (Mac OS X 10.11.0) Testing Home controller sample slides are correctly added in for the Aggregate Slides FAILED
        TypeError: angular.element.cleanData is not a function in /private/tmp/e-mission-server/webapp/node_modules/angular-mocks/angular-mocks.js (line 2915)
        module.$$cleanup@/private/tmp/e-mission-server/webapp/node_modules/angular-mocks/angular-mocks.js:2915:7
        module.$$afterEach@/private/tmp/e-mission-server/webapp/node_modules/angular-mocks/angular-mocks.js:2885:7
```

shankari:
Also, I need to run `karma` from the `webapp` directory, I think, not the root directory.
Might want to update README to reflect that.

adambrown13:
Is it there a problem when you run 'karma start my.conf.js' or just when './node_modules/karma/bin/karma start my.conf.js' is run. And I updated the readme to reflect the path, I'll push right now

shankari:
Hm. I am not able to run `karma start my.conf.js` because `karma` is not in my PATH.
I can add it to the PATH and try, but I'll wait until you update the README so that I can follow the exact instructions and ensure that the README is up to date.

adambrown13:
It appears to be a problem between 2 different configurations of angular mock. I can't find why angular mocks are being built differently, since the all of the configuration settings are the same. As a temporary fix, I added the angular_mock.js file that correctly runs. I'm not happy with it, so I'll dive into why that's the problem, but it runs now

shankari:
> It appears to be a problem between 2 different configurations of angular mock. 

You should be able to specify a particular version of angular mock during installation.
Check out the `npm` documentation. You can also create a `package.json` and then change the instructions to say `npm install`. See my posts on the habitica mobile code.

shankari:
Bunch of README fixes.
- Since you have checked in `angular_mocks`, remove the `npm install angular_mocks`, otherwise, it overrides and tests still fail
- `/node_modules/karma/bin/karma` should be `./node_modules/karma/bin/karma`

With those changes, the tests pass, so once you have fixed those minor issues, I can merge.

```
Firefox 47.0.0 (Mac OS X 10.11.0): Executed 1 of 2 SUCCESS (0 secs / 0.018 secs
Firefox 47.0.0 (Mac OS X 10.11.0): Executed 2 of 2 SUCCESS (0 secs / 0.026 secsFirefox 47.0.0 (Mac OS X 10.11.0): Executed 2 of 2 SUCCESS (0.005 secs / 0.026 secs)
```

shankari:
Filed https://github.com/e-mission/e-mission-server/issues/289 to track the cleanup

adambrown13:
Committed but didn't push earlier. ReadMe updated

PR #284, opened 2016-06-11
Updates on README

yw374cornell:
Please find below screenshots of the errors I got when trying to set up the server using Python 3.5:

1. The error I got after running pip to install the Python dependencies. According to the discussion [here](https://github.com/skoczen/will/issues/148), it seems like the wsgiref library does not support Python 3 yet.   
screen shot 2016-06-07 at 2 52 13 pm

2. The error I got after running the server. 
screen shot 2016-06-07 at 2 32 34 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ actively running when you attempt to start the server.
         % md c:\data\db\  
 or
 
-        on Mac (give the user account running mongod read and write permissions for the data directory)
+        on Mac (the user account running mongod must have read and write permissions for the data directory)
         $ mkdir -p /data/db
         $ chmod 777 /data/db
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -166,11 +170,19 @@ of the notebooks.
 1. If a python execution fails to import a module, make sure to add current
 directory to your PYTHONPATH.
 
-1. If starting the server gives a CONNECTION\_ERROR, make sure MongoDB is
+2. If starting the server gives a CONNECTION\_ERROR, make sure MongoDB is
```
shankari: For the record, you shouldn't need to do this because markdown should automatically generate numbers. For example, in the list below, all my numbers started with 1. You can try to add a github comment and check for yourself. But it doesn't hurt anything, so I won't hold up the merge for this.

1. Yawen
1. Wang
1. is
1. awesome

yw374cornell:
That's neat, thank you for the tip! I will go learn more about the Markdown syntax when I get time :)

> 0. Good 
> 1. night
> 3. !

README.md
yw374cornell: That's neat, thank you for the tip! I will go learn more about the Markdown syntax when I get time :)

> 0. Good
> 3. night
> 1. !

Issue #281, opened 2016-06-10
Move away from br-fullpage to show the full page


shankari:
`br-fullpage` was an angular-based fullpage javascript library.
It _was_ at https://github.com/BjornRombaut/br-fullpage.git/, but it looks like it has been removed.
we either need to switch to a different fullpage library, or to a different set of style decisions altogether.

shankari:
@sunil07t, adding this to your pile for now.

shankari:
`br-fullpage` didn't work properly on firefox anyway.
Maybe a native angular/ionic solution will work better?

shankari:
@sunil07t, can you revert my workaround by removing all references to br-fullpage and the checked in libraries?

Then, you can close this issue.

sunil07t:
Sorry, I did not include closes #281 in my commit. Is there another way that I can close this issue?

shankari:
You can just close it from the UI after I merge the pull request.

shankari:
Fixed by @sunil07t

PR #285, opened 2016-06-13
Updated the UI of the homepage

sunil07t:
Changed the fonts and colors of the text on the home page. Added new tab called Trip Planning that points to OTP server and changed the layout of each section to a card.


shankari:
High level comment for future changes: you should commit your changes to a branch, and then do the pull request from the branch. That way, if your pull request is completely hosed, you can just delete the branch and try again. Also allows you to start work on a new pull request while waiting for review on the old one :)

shankari:
Overall, looks good. I'm glad you were able to convert this to cards. Since you are making that change anyway, can you remove the dependency on the fullpage package.

https://github.com/e-mission/e-mission-server/issues/281

That is the second task I was going to assign to you, and switching to cards is basically the second option - " to a different set of style decisions altogether."

Detailed comments inline.

webapp/www/index.html
```diff
@@ -46,6 +46,7 @@
       <ion-nav-buttons side="primary">
         <a class="button button-block button-calm icon-left ion-home" ui-sref="home">Home</a>
         <a class="button button-block button-calm icon-left ion-ios-analytics" ui-sref="tab.heatmap">Analysis</a>
+        <a class="button button-block button-calm icon-left ion-map" href="http://54.162.140.55:8080">Trip Planning</a>
```
shankari: This is not strictly correct. Note that for the heatmap and home tabs, the tab header is visible at the top, so if you are in heatmap, you can go back to home. But with this option, if you click on trip planner, you go to the trip planner page but you can't go to (say) heatmap without clicking the back button.

You really do need to create a new tab for the trip planner, but it should be fairly straightforward to have it load the OTP server page as the HTML.

webapp/www/templates/home.html
```diff
@@ -45,43 +45,43 @@
 <fullpage class="instant-size" page-class="page">
```
shankari: As I said in the high level comment, since this is cards anyway, maybe this can be a list instead of fullpage?
That should help with the scrolling on firefox and also resolve #281.
Or maybe the solution is to use a different fullpage library?

webapp/www/templates/home.html
```diff
@@ -45,43 +45,43 @@
...
-            <p class="intro" style="color:brown;">The project consists of a smartphone application that functions as a personal travel diary, and an aggregate data analysis platform that enables interactive exploration of the collected data.</p>
+            <h1 class="introH">E-Mission</h1>
+            <p class="intro firstPar">E-Mission is a research project at UC Berkeley that allows all of us to share our travel patterns and experiences with our community and with urban planners.</p>
+            <p class="intro firstPar">Our goal is to enable personal and structural shifts to sustainable transportation.</p>
+            <p class="intro firstPar" >The project consists of a smartphone application that functions as a personal travel diary, and an aggregate data analysis platform that enables interactive exploration of the collected data.</p>
             </div>
             <div class="col-md-2">
-               <img ng-src="img/personal/compare_with_common.png" style="opacity: 0.5" height="480em">
+               <img ng-src="img/personal/compare_with_common.png" style="opacity: 0.7; float:right" height="480vw">
```
shankari: If the images are not going to overlap the text, we can remove the opacity...

webapp/www/templates/home.html
```diff
@@ -45,43 +45,43 @@
...
-          <h1>Personal</h1>
-          <div class="row" style="height: 480em">
+          <div class="row card" style="padding:1%;">
             <div class="col-md-4">
                 <uib-carousel id="personalCarousel" active="active" interval="myInterval" no-wrap="noWrapSlides">
                 <uib-slide ng-repeat="slide in personalSlides track by slide.id" index="slide.id">
                 <img ng-src="{{slide.image}}" style="margin:auto;">
-                <div class="carousel-caption" style="color:brown">
-                  <p style="color:brown">{{slide.text}}</p>
+                <div class="carousel-caption">
```
shankari: At least for me, the carousel arrows don't show up any more - instead, they show E079. Is this a font issue? See screenshots below.

screen shot 2016-06-13 at 10 56 24 pm

screen shot 2016-06-13 at 10 56 57 pm

webapp/www/templates/home.html
shankari: Maybe just convert this from fullpage to list and show Juliana/Yawen/Adam for design feedback and then decide whether we want a different fullpage library as a next step?

webapp/www/templates/home.html
```diff
@@ -94,64 +94,60 @@
...
+                <uib-carousel active="active" interval="myInterval" no-wrap="noWrapSlides">
                     <uib-slide ng-repeat="slide in aggregateSlides track by slide.id" index="slide.id">
-                      <img ng-src="{{slide.image}}" style="align:center;margin:auto;">
-                         <div class="carousel-caption" style="color:brown">
-                           <p style="color:brown">{{slide.text}}</p>
-                         </div>
-                     </uib-slide>
-                    </uib-carousel>
-                 </div>
+                        <img ng-src="{{slide.image}}" style="align:center;margin:auto;height:90%;width:90%">
```
shankari: Ditto.

shankari:
That's it. Looking forward to the fixed code!

webapp/www/templates/home.html
shankari: On Chrome, they are boxes. Certainly looks like a font issue to me.
screen shot 2016-06-13 at 11 29 56 pm

sunil07t:
Should I make a new branch and do a pull request for these fixes or just continue with master branch for this pull request? I will start working on the fixes right now.

shankari:
Continue with master for now. That was feedback for the next change.

webapp/www/templates/home.html
sunil07t: I removed the fullpage and set the layout to cards but a bug that I am trying to fix is linking nav bar to sections of the homepage.

webapp/www/templates/home.html
sunil07t: This seems to be working now after I removed fullpage.

webapp/www/templates/home.html
shankari: I am not sure we need the second nav bar - don't you think it looks a bit
weird? I just added it because the full page template required it.

You could just try removing it and see how it looks

Shankari

sunil07t: Yes it does look weird, I will remove it.

shankari:
Tested it out on a local copy - looks great!
Merging now and will deploy soon.

PR #286, opened 2016-06-16
Fix some embedded links by changing them to the new URL format

shankari:
And remove the fullpage link to the last card by converting it to a mailto: link

### DIFF
--- a/webapp/www/templates/home.html
+++ b/webapp/www/templates/home.html
@@ -47,7 +47,7 @@
                                 <li> are you using different transportation modes?</li>
                                 </ul>
                             </p>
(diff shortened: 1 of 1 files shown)


shankari:
@sunil07t, I just made some more minor fixes.
Let me know when you are doing with screenshots so that I can deploy!

sunil07t:
I took the screenshots

shankari:
jenkins, test this please

shankari:
The failure is not related to this change, which does not touch any python code. The underlying issue needs to be investigated and fixed, but doesn't need to hold up other changes.

shankari:
I will investigate and fix the underlying issue once I get some breathing room from the theory paper.

PR #287, opened 2016-06-21
removed all references to br-fullpage and uninstalled the br-fullpage…

sunil07t:
… library to resolve issue #281


shankari:
Did you try displaying the page to ensure that nothing is broken?
I love deleting obsolete code and making everything nice :)

sunil07t:
I checked it on my firefox browser it works fine but now when I check it in my chrome browser the page is broken. My chrome browser is trying to get br-fullpage, even though there is no command for it, this my just be the chrome browser issue on my computer. Could you please double check?

shankari:
You may need to do shift-reload, but OK, I'll check.

PR #299, opened 2016-06-28
Clean up current implementation of the common trips

shankari:
It is useless to try and debug the code unless it is actually understandable.
Therefore, we remove the old, obsolete code to improve code clarity.
This reveals that the tests had not been fixed to work with the new code and so were testing currently unused code.
Ended up fixing the tests as well.

Check out commit messages for more details.

### DIFF
--- a/emission/tests/analysisTests/TestSimilarity.py
+++ b/emission/tests/analysisTests/TestSimilarity.py
@@ -1,29 +1,36 @@
+import logging
 import unittest
-import emission.core.get_database as edb
-import emission.analysis.modelling.tour_model.similarity as similarity
-import emission.simulation.trip_gen as tg
-import math
-from emission.core.wrapper.trip_old import Trip, Coordinate
-import emission.analysis.modelling.tour_model.cluster_pipeline as cp
+import uuid
+import time
 import datetime 
 import os, os.path
 
+import emission.tests.analysisTests.tourModelTests.common as etatc
+
+import emission.core.get_database as edb
+
+import emission.analysis.modelling.tour_model.similarity as similarity
+import emission.analysis.modelling.tour_model.cluster_pipeline as cp
+
+import emission.storage.timeseries.abstract_timeseries as esta
+
 class SimilarityTests(unittest.TestCase):
 
-    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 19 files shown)


shankari:
@sunil07t, you can review this now

emission/user_model_josh/utility_model.py
```diff
@@ -127,7 +126,7 @@ def get_top_choice_places(self, start_place, end_place):
         return self.get_top_choices_lat_lng(start, end)
 
     def get_all_trips(self, start, end, curr_time=None):
-        c = googlemaps.client.Client(GOOGLE_MAPS_KEY)
+        c = gmaps.client.Client(GOOGLE_MAPS_KEY)
```
sunil07t: Line 507 has googlemaps instead of gmaps, does that need to be gmaps?

shankari:
@sunil07t, good catch! Fixed.
I guess Josh didn't write any test cases for those functions...

shankari:
common_trips_working_again

emission/tests/analysisTests/TestSimilarity.py
```diff
@@ -86,10 +94,10 @@ def testDeleteBins(self):
             self.assertTrue(b == sim.num)
 
     def testElbowDistance(self):
-        start = Coordinate(47,-122)
-        end = Coordinate(47,-123)
-        now = datetime.datetime.now()
-        t = Trip(None, None, None, None, now, now, start, end)
+        start = [-122,47]
+        end = [-133,47]
```
sunil07t: is the x coordinate a typo or did you change it compared to old test? (-123 instead of -133)

shankari: Typo. Will fix along with anything else that you catch.
Also, note that I didn't write the original tests - they were written by a SUPERB student from the last year. So you too can write tests :)

sunil07t:
That is it! Everything else is clear :+1:

emission/tests/analysisTests/TestSimilarity.py
sunil07t: I can try to write the test codes after Goals tab :)

PR #300, opened 2016-06-28
Support the new google play motion activity format

shankari:
As part of the change to slim down the apk for android,
https://github.com/e-mission/e-mission-phone/pull/46,
https://github.com/e-mission/e-mission-data-collection/pull/116
we switched from google play version from 8.1.0 to 8.3.0, which is the version
that has separate jar files, at least in my install.

But this changed the motion activity format - the field names are now `zzaKM`
and `zzaKN` instead of `zzaEg` and `zzaEh`. So we change the formatter on the
server to handle this use case as well.

Note that we should really fix
https://github.com/e-mission/e-mission-data-collection/issues/80
to stop running into this in the future

### DIFF
--- a/emission/net/usercache/formatters/android/motion_activity.py
+++ b/emission/net/usercache/formatters/android/motion_activity.py
@@ -18,13 +18,18 @@ def format(entry):
     data = ad.AttrDict()
     if 'agb' in entry.data:
         data.type = ecwa.MotionTypes(entry.data.agb).value
-    else:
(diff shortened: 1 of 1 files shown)


shankari:
I have already made this change directly on the server so that it is not broken.
But we should really push it as part of the next server release as well so that we don't have uncommitted code floating around.

shankari:
@sunil07t in case you want more reviewing practice....

sunil07t:
:+1:

PR #304, opened 2016-07-07
Changes made in server for generating maps

yw374cornell:


shankari:
I'll merge and test this, but just FYI:
- this is still from master
- it includes a change that we definitely don't want to commit (hardcoding the a path with your name)

I'll pull and try to experiment with this on the weekend - I want to get the metrics calculation in place for now

yw374cornell:
oh right.. That was a temporary workaround before I got the python path set up correctly. I should have changed it back. Thanks for resolving that for me!

yw374cornell:
actually, I have reverted changes made on master and pulled from upstream if you looked at the latest commits. I created a separate branch that contains the changes made in plotting files and I could create another pull request from that branch.

shankari:
yeah I noticed that there were no files listed :)
would be great to send a PR from the other branch when free

PR #305, opened 2016-07-07
Habitica new user registration

juemura:
Missing: 
- update/delete user
- wrapper for other http requests
- Improve unit testing

### DIFF
--- a/conf/net/keys.json
+++ b/conf/net/keys.json
@@ -10,5 +10,10 @@
     "client_secret": "Get from https://dev.moves-app.com",
     "https_redirect_url": "Value registered at https://dev.moves-app.com",
     "ios_redirect_url": "Value registered at https://dev.moves-app.com"
+  },
+  "habitica" : {
+    "url": "http://54.159.38.241:3000",
+    "api_id": "Get from http://54.159.38.241:3000",
+    "api_token": "Get from http://54.159.38.241:3000"
   }
 }
--- a/conf/net/keys.json
+++ b/conf/net/keys.json
@@ -12,8 +12,6 @@
     "ios_redirect_url": "Value registered at https://dev.moves-app.com"
   },
   "habitica" : {
-    "url": "http://54.159.38.241:3000",
-    "api_id": "Get from http://54.159.38.241:3000",
-    "api_token": "Get from http://54.159.38.241:3000"
+    "url": "http://54.159.38.241:3000"
   }
 }
--- a/emission/net/ext_service/habitica/register.py
+++ b/emission/net/ext_service/habitica/register.py
@@ -1,0 +1,24 @@
+# Standard imports
+import json
+import requests
+import logging
+
+# Our imports
+from sdk import Habitica
+from emission.core.get_database import get_habitica_db
+
+key_file = open('conf/net/keys.json')
+key_data = json.load(key_file)
+url = key_data["habitica"]["url"]
+
+def habiticaRegister(username, email, password, our_uuid):
+
+  register_url = url + '/api/v3/user/auth/local/register'
+  user_request = {'username': username,'email': email,'password': password,'confirmPassword': password}
+  u = requests.post(register_url, json=user_request)
+  user_dict = json.loads(u.text)
+  logging.debug("parsed json from habitica = %s" % user_dict)
+
+  user_access_table = get_habitica_db()
+  user_access_table.insert({'user_id': our_uuid, 'habitica_username': username, 'habitica_id': user_dict['data']['_id'], 'habitica_token': user_dict['data']['apiToken']})
+
--- a/emission/net/ext_service/habitica/register.py
+++ b/emission/net/ext_service/habitica/register.py
@@ -4,7 +4,6 @@ import requests
 import logging
 
 # Our imports
-from sdk import Habitica
 from emission.core.get_database import get_habitica_db
 
 key_file = open('conf/net/keys.json')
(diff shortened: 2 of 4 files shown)


shankari:
ok to test

shankari:
High level comment: in the future, please submit pull reviews from branches
Makes it easier to clean up bad mistakes and revert unnecessary changes

conf/net/keys.json
```diff
@@ -10,5 +10,10 @@
     "client_secret": "Get from https://dev.moves-app.com",
     "https_redirect_url": "Value registered at https://dev.moves-app.com",
     "ios_redirect_url": "Value registered at https://dev.moves-app.com"
+  },
+  "habitica" : {
+    "url": "http://54.159.38.241:3000",
+    "api_id": "Get from http://54.159.38.241:3000",
```
shankari: `api_id` and `api_token` are not used, so should be removed to avoid confusion

emission/net/ext_service/habitica/register.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Our imports
+from sdk import Habitica
+from emission.core.get_database import get_habitica_db
+
+key_file = open('conf/net/keys.json')
+key_data = json.load(key_file)
+url = key_data["habitica"]["url"]
+
+def habiticaRegister(username, email, password, our_uuid):
```
shankari: @sunil07t and @juemura need to decide login flow. will the user specify a password or will we autogenerate on? Everybody hates coming up with new passwords every time and requiring a separate password means that the integration is not as seamless as we would like.

On the other hand, if we want to use the standard habitica, maybe we want to allow existing users to login to their existing account?

emission/tests/netTests/TestHabiticaRegister.py
```diff
@@ -0,0 +1,39 @@
...
+  def setUp(self):
+    print "Test setup called"
+    self.testUserUUID = uuid.uuid4()
+    self.sampleAuthMessage1 = {'username': '99999999999995543', 'email': '99999999999995543@test.com', 'password': 'aa!A15551', 'our_uuid': self.testUserUUID}
+
+  # One part of this is the communication with habitica. That is an integration test, not a unit test,
+  # so we don't test that here. Instead, we test the save and retrieval of the datastructure 
+  # that we get back from habitica. Some sample auth messages are shown above
+
+  # TODO: See if we can use Mock to create a mock habitica server that can allow
```
shankari: yes, this is a good suggestion

emission/tests/netTests/TestHabiticaRegister.py
```diff
@@ -0,0 +1,39 @@
...
+import emission.net.ext_service.habitica.register as reg
+import emission.core.get_database as edb
+
+class TestHabiticaRegister(unittest.TestCase):
+  def setUp(self):
+    print "Test setup called"
+    self.testUserUUID = uuid.uuid4()
+    self.sampleAuthMessage1 = {'username': '99999999999995543', 'email': '99999999999995543@test.com', 'password': 'aa!A15551', 'our_uuid': self.testUserUUID}
+
+  # One part of this is the communication with habitica. That is an integration test, not a unit test,
```
shankari: This is not accurate - this does actually test the communication with habitica since the register code makes a call to the server.

shankari:
I will merge this as soon as you fix the `conf` file and the tests pass.
Other comments are basically nits

shankari:
jenkins, test this please

shankari:
Also, you need to include `sdk` in `requirements.txt` otherwise the import fails

```
ERROR: netTests.TestHabiticaRegister (unittest.loader.ModuleImportFailure)
ImportError: Failed to import test module: netTests.TestHabiticaRegister
  File "/home/anaconda/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/home/anaconda/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "<https://amplab.cs.berkeley.edu/jenkins/job/e-mission-server-prb/ws/emission/tests/netTests/TestHabiticaRegister.py",> line 9, in <module>
    import emission.net.ext_service.habitica.register as reg
  File "emission/net/ext_service/habitica/register.py", line 7, in <module>
    from sdk import Habitica
ImportError: No module named sdk
```

Alternatively, since we are not using the sdk now, and it is not clear how much the sdk adds given the planned passthrough, we can just remove the sdk for now add it later when we really need it

shankari:
Test failed because:

```
ERROR: testAddNewUser (netTests.TestHabiticaRegister.TestHabiticaRegister)
  File "<https://amplab.cs.berkeley.edu/jenkins/job/e-mission-server-prb/ws/emission/tests/netTests/TestHabiticaRegister.py",> line 22, in testAddNewUser
    sampleAuthMessage1Ad.password, sampleAuthMessage1Ad.our_uuid)
  File "emission/net/ext_service/habitica/register.py", line 26, in habiticaRegister
    'habitica_password': password, 'habitica_id': user_dict['data']['_id'],
KeyError: 'data'

```

I think that we hit this yesterday - you get a 401 when you try to reuse the user email.
```
{u'message': u'Email address is already used in an account.', u'success': False, u'error': u'NotAuthorized'}
```

I guess you fixed it by changing the email address at that time. That doesn't work as more than a stopgap because the unit tests are run over and over again. If a particular parameter needs to be unique, it needs to be freshly generated each time.

Randomly generate username and user email in the test instead of hardcoding them?

juemura:
Sorry, there was a silly mistake. Fixed now. I added code to randomly generate a string, but maybe you already have something like that and I can just import it? If not, this should work.

shankari:
> I added code to randomly generate a string, but maybe you already have something like that and I can just import it?

I don't think that I do. The only random generation that I do is the UUID generation. This looks good - we could move it into `common` if we need to use it other places as well. But for now, this is fine.

shankari:
Congratulations! First pull request merged?

PR #306, opened 2016-07-07
Add a POST API method for habiticaRegister

shankari:
This is an exemplar to show how to add an externally invocable POST API method
to the e-mission server.

For further explanation and examples, please consult the documentation for bottle.
http://bottlepy.org/docs/dev/index.html

@juemura, @sunil07t, @yw374cornell: this might be helpful to you as you
consider adding other methods.

This method would be invoked by sending a HTTP POST

`$http.post()`
or
`requests.post()`

with data {'username': ..., 'email': ....}

While calling this from the emission app, you should use a helper function
defined in `CommHelper` so that the JWT is inserted correctly into the HTTP
POST message.

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -443,7 +443,7 @@ def habiticaRegister():
                 (request.json, request))
   user_uuid = getUUID(request)
   assert(user_uuid is not None)
-  username = request.json['username']
+  username = request.json['regConfig']['username']
   # This is the second place we use the email, since we need to pass
   # it to habitica to complete registration. I'm not even refactoring
   # this into a method - hopefully this makes it less likely to be reused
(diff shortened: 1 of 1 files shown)


shankari:
I felt this was faster than yet another round of explaining with text snippets or diagrams.
THIS HAS NOT BEEN TESTED.
But I expect it to work with minor bug fixes.

shankari:
jenkins, test this please

shankari:
jenkins, test this please

emission/net/api/cfc_webapp.py
```diff
@@ -435,6 +436,21 @@ def movesCallback():
...
   state = request.json['state']
   return auth.movesCallback(code, state, user_uuid)
+
+@post('/habiticaRegister')
+def habiticaRegister():
+  logging.debug("habitica registration request %s from user = %s" %
+                (request.json, request))
+  user_uuid = getUUID(request)
+  assert(user_uuid is not None)
+  user = User.fromUUID(user_uuid)
```
juemura: why do you create a user here? I don't see where it's used.

shankari: Not used yet, but will be used soon to get the stored email address.

shankari:
Changed the expected format slightly to match the authenticated API call from the phone and to be a better example for future changes.
phone service: https://github.com/e-mission/e-mission-phone/pull/63
calls comm function: https://github.com/e-mission/cordova-server-communication/pull/12

PR #307, opened 2016-07-08
Implement generic habitica proxy and expose it via the API

shankari:
This allows us to invoke habitica methods after mapping from our UUID to the
habitica UUID without duplicating the entire habitica API, which is pretty
large, TBH.

The new method is implemented in `proxy.py` (which is renamed from
`register.py` and has been tested with `TestHabiticaRegister.py`).

We've also exposed this proxy via an API method so that we can reuse it on the
phone as well.

The API is a POST method which expects JSON with the following format

```
    {'callOpts': {'method': ...,
                  'method_url': ....
                  'method_args': ....}
```

Note that although the proxy has been tested, the API has not been tested.
But it is fairly straightforward, so should work with minor bug fixes.


shankari:
@juemura, @sunil07t, this is what I meant by the proxy.
Better to call it a proxy than a wrapper because wrapper can be used for both sdk and proxy.
proxy is more precise.

juemura:
This makes so much more sense. Thank you!

emission/tests/netTests/TestHabiticaRegister.py
```diff
@@ -19,18 +19,66 @@ def setUp(self):
...
+      proxy.habiticaRegister(sampleAuthMessage1Ad.username, sampleAuthMessage1Ad.email,
+                             sampleAuthMessage1Ad.password, sampleAuthMessage1Ad.our_uuid)
+      ret_profile = proxy.habiticaProxy(self.testUserUUID, "GET",
+                                        "/api/v3/user", None)
+      ret_json = ret_profile.json()
+      logging.debug("Retrieved profile with keys %s" % ret_json.keys())
+      logging.debug("profile data keys = %s" % ret_json['data'].keys())
+      # User has just been created, so has no gear
+      self.assertEqual(ret_json['data']['achievements']['ultimateGearSets'],
+                       {'warrior': False, 'rogue': False, 'wizard': False,
```
sunil07t: users are default warrior until round 10.

shankari: well, the test passed, so if the user is a default warrior until round 10, then they must be making a call from the UI to set that

shankari: Note also that this is gear sets. So maybe the user is `class` warrior but has no gear?

shankari: class API: https://habitica.com/apidoc/#api-User-UserChangeClass
gear API?: https://habitica.com/apidoc/#api-User-UserBuyGear

sunil07t: Oh yes I just checked the new user has no gear but is class warrior.

PR #308, opened 2016-07-09
Make sure to return a value from habiticaRegister

shankari:
Without this, there is no value returned from the REST API, and we get an
error on the client.

`"During server call, error Error Domain=NSCocoaErrorDomain Code=3840 \"No value.\" UserInfo={NSDebugDescription=No value.}"`


shankari:
@sunil07t this is the fix to `habiticaRegister`.
You might want to apply the change locally and continue debugging while waiting for the tests to finish running...

PR #309, opened 2016-07-09
Add a new script to clean up a habitica user given user email

shankari:
- Looks up uuid
- uses that to lookup password
- calls delete method

Simple!


shankari:
@sunil07t @juemura you might find this helpful as you go through the testing process.
@juemura this can also be a template for other habitica calls in the future

shankari:
jenkins, test this please

PR #310, opened 2016-07-10
Add support for computing metrics

shankari:
Add support for computing metrics across various time ranges, both in local and in UTC time.
As always, the local time computations were the most complex, but I think that they are finally correct.
The actual metrics computed are represented by simple functions on the sections.

These are very simple metrics.
The next step will be to use calorie calculations, which require user profile information, such as height, weight and age.


shankari:
This appears to be an intermittent failure that is not related to any code.

In particular, if I check out a brand new repo, checkout to the version that just passed, then the test fails.

```
$ git clone https://github.com/shankari/e-mission-server.git test_wierd_failure
$ cd test_wierd_failure/
$ git checkout -b cleanup_data_models
$ git checkout 1a225e004ef79d039c5d8dcd4def4d7988f433a0
$ ./e-mission-py.bash emission/tests/analysisTests/TestRepresentatives.py RepresentativesTests.testClusterDict
$ ======================================================================
FAIL: testClusterDict (__main__.RepresentativesTests)
  File "emission/tests/analysisTests/TestRepresentatives.py", line 139, in testClusterDict
    self.assertTrue(c['start'] != c['end'])
AssertionError: False is not true

Ran 1 test in 0.963s
```

shankari:
In fact, it is failing even on master.

```
$ git checkout master
bash-3.2$ git log | head
commit 7b0331eeae40c7e4b7db6cd98b7b7f2480d8c204
Merge: 9c297c9 f37b608
Author: shankari <shankari@eecs.berkeley.edu>
Date:   Thu Jul 7 21:40:47 2016 -0700

    Merge pull request #307 from shankari/add_api_hook
    
    Implement generic habitica proxy and expose it via the API

$ ./e-mission-py.bash emission/tests/analysisTests/TestRepresentatives.py RepresentativesTests.testClusterDict

FAIL: testClusterDict (__main__.RepresentativesTests)
  File "emission/tests/analysisTests/TestRepresentatives.py", line 139, in testClusterDict
    self.assertTrue(c['start'] != c['end'])
AssertionError: False is not true

```

shankari:
Given this, I am not going to try to investigate further or fix now.
I'll retry later to see if the issue auto-resolved.
Otherwise, since we don't use the representatives anyway and are probably going to rewrite that code, we will just move this test to the ignored tests section.

shankari:
Tests just passed on Yawen's PR.
Let's try this again.

shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #313, opened 2016-07-14
Fixed the script for pulling and loading public data

yw374cornell:


shankari:
ok to test

shankari:
jenkins, test this please

emission/net/api/cfc_webapp.py
```diff
@@ -383,17 +384,14 @@ def getCarbonCompare():
...
 
-#Pulling public data from the server  
-@get('/getData')
-def getData():
+# Pulling public data from the server  
+@get('/eval/publicData/timeseries')
+def getPublicData():
   from_date = request.query.from_date
   to_date = request.query.to_date
+  from_ts = arrow.get(from_date).timestamp
```
shankari: Instead of passing in the date to `getPublicData` and YYYY-MM-DD, why not just pass it directly as a UTC timestamp. Unambiguous. Clear.

request_public_data.py
```diff
@@ -23,16 +25,19 @@
 if args.verbose:
 	logging.basicConfig(level=logging.DEBUG)
 
-r = requests.get("http://" + server_url + "/getData?from_date=" + from_date + "&to_date=" + to_date)
+from_date_UTC = arrow.get(from_date).replace(tzinfo='local').to('utc').format('YYYY-MM-DD HH:mm:ss')
```
shankari: I am not sure that this actually parses the data correctly. IIRC, you need to do

`arrow.get(year=xx, month=xx, day=xx, tzinfo='local')`
then you can just do `.timestamp` on that.

I'd suggest some poking around with arrow in an ipython notebook

shankari: Concretely, I think that `strptime` will work.

shankari: because you can specify the `tzinfo` at creation time.

yw374cornell: yea, I didn't use strptime because I thought one of the advantage of using arrow is that it parses a string to a date/arrow object directly. However, if you use arrow.get() to parse a string then you cannot specify the `tzinfo` at creation time (it converts to UTC by default), but you can use the replace function to set `tzinfo`.
I have tested this out in ipython notebook, and this does work: 
#### input:
```python
from_date = '2016-07-12'
to_date = '2016-07-13'

from_date_UTC = arrow.get(from_date).replace(tzinfo='local').to('utc').format('YYYY-MM-DD HH:mm:ss')
to_date_UTC = arrow.get(to_date).replace(tzinfo='local').to('utc').format('YYYY-MM-DD HH:mm:ss')
print from_date_UTC
print to_date_UTC

from_ts = arrow.get(from_date_UTC).timestamp
to_ts = arrow.get(to_date_UTC).timestamp
print from_ts
print to_ts

from_dt = arrow.get(from_ts)
to_dt = arrow.get(to_ts)
print from_dt
print to_dt
```   

#### output:  
```
2016-07-12 07:00:00
2016-07-13 07:00:00
1468306800
1468393200
2016-07-12T07:00:00+00:00
2016-07-13T07:00:00+00:00
```

emission/net/api/cfc_webapp.py
yw374cornell: absolutely! I started with doing all time conversion in the API and then migrated some to the script. I should have just done the timestamp conversion in the script. Will make that change soon and then create a pull request

request_public_data.py
shankari: so then why is the data retrieved in UTC anyway?

shankari: Maybe print the parsed timestamps on the server side as well and compare...

yw374cornell: yea you are right... The conversion from `local time` to `UTC` is unnecessary. I thought you had to convert to `UTC` to get the correct timestamp for some reason, but that's wrong

yw374cornell:
The script seems to work on the local server. Here is how I tested it  

1. loaded data from `2016-7-11` to `2016-7-13` to the local server 
2. pulled data from `2016-7-12` to `2016-7-13` from the remote server: `write_fmt_time` for the first pulled entry is `2016-07-11T17:00:00`
3. pulled data from `2016-7-12` to `2016-7-13` from the local server: `write_fmt_time` for the first pulled entry is `2016-07-12T00:00:18`

PR #314, opened 2016-07-15
Habitica autocheck

juemura:
It's missing the tests, I'll do them today.
I had some questions, which are all indicated by "FIX!"

### DIFF
--- a/emission/net/ext_service/habitica/proxy.py
+++ b/emission/net/ext_service/habitica/proxy.py
@@ -26,14 +26,17 @@ def habiticaRegister(username, email, password, our_uuid):
   #Since we are randomly generating the password, we store it in case users 
   #want to access their Habitica account from the browser
   #Need to create a way from them to retrieve username/password
+  #metrics_data is used to calculate points based on km biked/walked
+  #last_timestamp is the last time the user got points, and bike/walk_count are the leftover km
   habitica_user_table = edb.get_habitica_db()
   habitica_user_table.insert({'user_id': our_uuid, 
-    'last_timestamp': arrow.utcnow().timestamp,
+    'metrics_data': {'last_timestamp': arrow.utcnow().timestamp, 'bike_count': 0, 'walk_count': 0},
     'habitica_username': username, 
     'habitica_password': password, 
     'habitica_id': user_dict['data']['_id'], 
     'habitica_token': user_dict['data']['apiToken']})
   return user_dict
+  response = proxy.habiticaProxy(uid, 'DELETE', "/api/v3/user", {'password': password})
 
 def habiticaProxy(user_uuid, method, method_url, method_args):
   logging.debug("For user %s, about to proxy %s method %s with args %s" %
--- a/emission/net/ext_service/habitica/proxy.py
+++ b/emission/net/ext_service/habitica/proxy.py
@@ -2,6 +2,7 @@
 import json
 import requests
 import logging
+import arrow
 
 # Our imports
 import emission.core.get_database as edb
@@ -26,8 +27,11 @@ def habiticaRegister(username, email, password, our_uuid):
   #want to access their Habitica account from the browser
   #Need to create a way from them to retrieve username/password
   habitica_user_table = edb.get_habitica_db()
-  habitica_user_table.insert({'user_id': our_uuid, 'habitica_username': username, 
-    'habitica_password': password, 'habitica_id': user_dict['data']['_id'], 
+  habitica_user_table.insert({'user_id': our_uuid, 
+    'last_timestamp': arrow.utcnow().timestamp,
+    'habitica_username': username, 
+    'habitica_password': password, 
+    'habitica_id': user_dict['data']['_id'], 
     'habitica_token': user_dict['data']['apiToken']})
   return user_dict
 
--- a/emission/net/ext_service/habitica/sync_habitica.py
+++ b/emission/net/ext_service/habitica/sync_habitica.py
@@ -1,0 +1,70 @@
+# Standard imports
+import json
+import requests
+import logging
+import arrow
+import attrdict as ad
+
+# Our imports
+import emission.core.get_database as edb
+import emission.net.ext_service.habitica.proxy as proxy
+import emission.analysis.result.metrics.simple_metrics as earmts
+import emission.analysis.result.metrics.time_grouping as earmt
+
+
+
+def create_habit(user_id, new_habit):
+  method_uri = "/api/v3/tasks/user"
+  #how to pass in params? https://habitica.com/apidoc/#api-Task-GetUserTasks
+  #FIX! right now it's returning all types of tasks
+  params = {'type': "habits"}
+  result = proxy.habiticaProxy(user_id, 'GET', method_uri, params)
+  habits = result.json()
+  for habit in habits['data']:
+    if habit['text'] == new_habit['text']:
+      #FIX! this is returning all of the tasks instead of just the one that matches what we want
+      return habit['_id']
+  #if habit not found, create habit
+  response = proxy.habiticaProxy(user_id, 'POST', method_uri, new_habit)
+  habit_created = response.json()
+  return habit_created['data']['_id']
+
+
+
+def reward_active_transportation(self, user_id):
+  #make sure habits exist
+  #bike
+  bike_habit = {'type': "habit", 'text': "Bike", 'up': True, 'down': False, 'priority': 2}
+  bike_habit_id = create_habit(user_id, bike_habit)
+  #walk
+  walk_habit = {'type': "habit", 'text': "Walk", 'up': True, 'down': False, 'priority': 2}
+  walk_habit_id = create_habit(user_id, walk_habit)
+
+  #get timestamps
+  user_val = list(edb.get_habitica_db().find({"user_id": user_id}))[0]
+  timestamp_from_db = user_val['last_timestamp']
+  timestamp_now = arrow.utcnow().timestamp
+  
+  #FIX! how do I change the frequency to be just one block?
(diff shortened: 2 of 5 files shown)


shankari:
Awesome! Congratulations on your first change!
I'll merge later today to give you a chance to address some of the TODOs and maybe write the tests. Did you test it E2E with Sunil's phone changes?

emission/net/ext_service/habitica/proxy.py
```diff
@@ -26,8 +27,11 @@ def habiticaRegister(username, email, password, our_uuid):
   #want to access their Habitica account from the browser
   #Need to create a way from them to retrieve username/password
   habitica_user_table = edb.get_habitica_db()
-  habitica_user_table.insert({'user_id': our_uuid, 'habitica_username': username, 
-    'habitica_password': password, 'habitica_id': user_dict['data']['_id'], 
+  habitica_user_table.insert({'user_id': our_uuid, 
+    'last_timestamp': arrow.utcnow().timestamp,
```
shankari: what does last_timestamp store?

emission/net/ext_service/habitica/sync_habitica.py
```diff
@@ -0,0 +1,68 @@
...
+import emission.analysis.result.metrics.simple_metrics as earmts
+import emission.analysis.result.metrics.time_grouping as earmt
+
+
+
+def create_habit(user_id, new_habit):
+  method_uri = "/api/v3/tasks/user"
+  #how to pass in params? https://habitica.com/apidoc/#api-Task-GetUserTasks
+  #FIX! right now it's returning all types of tasks, I want it to return only "habits"
+  params = {'type': "habits"}
```
shankari: Looks like you figured out how to retrieve params...

emission/net/ext_service/habitica/sync_habitica.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Our imports
+import emission.core.get_database as edb
+import emission.net.ext_service.habitica.proxy as proxy
+import emission.analysis.result.metrics.simple_metrics as earmts
+import emission.analysis.result.metrics.time_grouping as earmt
+
+
+
+def create_habit(user_id, new_habit):
```
shankari: This doesn't actually create the habit, it just retrieves it, right?
(you are calling `GetUserTasks`)

Where are the tasks created? Do you expect the user to create them? Are they created directly from the phone?

emission/net/ext_service/habitica/sync_habitica.py
```diff
@@ -0,0 +1,68 @@
...
+  walk_habit = {'type': "habit", 'text': "Walk", 'up': True, 'down': False, 'priority': 2}
+  walk_habit_id = create_habit(user_id, walk_habit)
+
+  #get timestamps
+  user_val = list(edb.get_habitica_db().find({"user_id": user_id}))[0]
+  timestamp_from_db = user_val['last_timestamp']
+  timestamp_now = arrow.utcnow().timestamp
+  
+  #FIX! how do I change the frequency to be just one segment? For now I just left it as yearly
+  summary_ts = earmt.group_by_timestamp(user, timestamp_from_db, timestamp_now, earmt.LocalFreq.YEARLY, earmts.get_distance)
```
shankari: Are you sure this works?
IIRC, for `group_by_timestamp` you pass in the pandas dateoffset strings.

`The frequency as specified in a pandas date_range frequency string.
    We only support frequencies of a day or longer in order to return the data
    in a format that makes sense`

If you want only one segment, year seems like a good way to get it, given that this script runs every hour.

emission/net/ext_service/habitica/sync_habitica.py
```diff
@@ -0,0 +1,68 @@
...
+
+  #FIX! is this getting the distances properly? Is it possible to have a function that calls "create sample data" to test the metrics
+  bike_distance = summary_ts[0].BICYCLING
+  walk_distance = summary_ts[0].ON_FOOT
```
shankari: You can load sample data using the instructions in the README.
I can also give you ~ 6 mos of data from me, and you can use that to experiment with real data.

emission/net/ext_service/habitica/sync_habitica.py
```diff
@@ -0,0 +1,68 @@
...
+
+  #FIX! consider saving the mod (eg bike_distance%3000) to count towards the next reading
+  #FIX! consider creating separate function to "score_habit"
+  #reward user by + habits
+  # Walk: 1 plus for every km
+  for i in range(walk_distance//1000):
+    res = proxy.habiticaProxy(user_id, 'POST', method_uri_walk, None)
```
shankari: does this even work? I don't see `method_uri_walk` defined anywhere?

emission/net/ext_service/habitica/proxy.py
juemura: the timestamp from when the habitica user is created. Then, when the autocheck is first called, it calculates the distance from the beginning, and updates the timestamp. After that, the method always calculates the distance walked/biked since the last time it was scored. This avoids users from getting rewarded multiple times for the same thing, and it bypasses the problem of when to reward (every hour? once a day? etc)

emission/net/ext_service/habitica/sync_habitica.py
juemura: I didn't, this doesn't actually work. It's still returning all types of tasks.

emission/net/ext_service/habitica/sync_habitica.py
juemura: No, this checks if the task has already been created, if so, it returns its id, otherwise, it creates it (see last 3 lines of method, copied below)
"response = proxy.habiticaProxy(user_id, 'POST', method_uri, new_habit)
  habit_created = response.json()
  return habit_created['data']['_id']"

emission/net/ext_service/habitica/proxy.py
shankari: There's a standard technique that I use for this with the other pipeline states.
Look at `emission/core/wrapper/pipelinestate.py`, `emission/storage/pipeline_queries.py` for examples of where it is defined and `emission/analysis/intake/segmentation/trip_segmentation.py` - e.g. `time_query = epq.get_time_range_for_segmentation(user_id)` for where it is used.

shankari: You might want to reuse the same technique so that I can roll back and re-run the pipeline in a unified fashion if there are errors. But we need to think about what that means from a habitica perspective.

emission/net/ext_service/habitica/sync_habitica.py
shankari: that's wierd - what happens if you run it in isolation directly using requests?
What happens when you run it against the standard habitica server?

emission/net/ext_service/habitica/sync_habitica.py
shankari: Ok. The habitica documentation around this method is very rudimentary/unclear.
If you feel up to it, you can look at the source and see how they treat the parameters...

emission/net/ext_service/habitica/sync_habitica.py
shankari: If you wanted to be more correct, you could also change `group_by_timestamp` to skip the grouping if the freq is None.

emission/net/ext_service/habitica/sync_habitica.py
juemura: I didn't try the real habitica server but I did try making direct request calls and that didn't work

juemura:
All done.

emission/net/ext_service/habitica/sync_habitica.py
```diff
@@ -0,0 +1,86 @@
...
+  walk_habit = {'type': "habit", 'text': "Walk", 'up': True, 'down': False, 'priority': 2}
+  walk_habit_id = create_habit(user_id, walk_habit)
+
+  #get timestamps
+  user_val = list(edb.get_habitica_db().find({"user_id": user_id}))[0]['metrics_data']
+  timestamp_from_db = user_val['last_timestamp']
+  timestamp_now = arrow.utcnow().timestamp
+  
+  #Get metrics
+  summary_ts = earmt.group_by_timestamp(user_id, timestamp_from_db, timestamp_now, None, earmts.get_distance)
```
shankari: Interesting to see that `None` works. I know it won't work for the search by local date, but for timestamp, I just use the pandas `Grouper` directly and it looks like it does The Right Thing.

Cool! I'll file an issue to do the right thing for the local date as well.

shankari: Filed https://github.com/e-mission/e-mission-server/issues/320

shankari:
ok to test

emission/net/ext_service/habitica/proxy.py
```diff
@@ -25,11 +26,17 @@ def habiticaRegister(username, email, password, our_uuid):
   #Since we are randomly generating the password, we store it in case users 
   #want to access their Habitica account from the browser
   #Need to create a way from them to retrieve username/password
+  #metrics_data is used to calculate points based on km biked/walked
+  #last_timestamp is the last time the user got points, and bike/walk_count are the leftover km
   habitica_user_table = edb.get_habitica_db()
-  habitica_user_table.insert({'user_id': our_uuid, 'habitica_username': username, 
-    'habitica_password': password, 'habitica_id': user_dict['data']['_id'], 
+  habitica_user_table.insert({'user_id': our_uuid, 
+    'metrics_data': {'last_timestamp': arrow.utcnow().timestamp, 'bike_count': 0, 'walk_count': 0},
```
shankari: Filed #321 to track the issue of whether the habitica game calculation should be progressive or not.

PR #315, opened 2016-07-16
Reuse mongodb connection across multiple calls

shankari:
In the current implementation of `get_database`, which has been unchanged for the past two years at least, we open a connection to the database `MongoClient()` or `MongoClient('localhost')` for each call to get a collection.
    
This is fine if we only have a few modules with database access, but as we build more modules with database access, we open connections over and over again. This is bad from both a performance perspective and a connection exhaustion perspective.
    
http://stackoverflow.com/questions/14495975/why-is-it-recommended-not-to-close-a-mongodb-connection-anywhere-in-node-js-code
    
We have run into connection exhaustion before this in particular, frequently used functions and have fixed them on an ad-hoc basis (e.g. 2032751816484d77d5331eeaa10e8032c3a05864 and 84e2474a11e3bb5cd3d9995ea6f7e878ad7e017b).
    
However, as the number of modules that called the functions increased, and the number of test cases that we ran increased, the cumulative number of connections increased, and we began exhausting connections. For a while, this used to fail on my laptop but pass on the CI server, but as the number of tests increased, it started failing on the CI server as well.
    
We work around this by opening a single connection to the database on module init and then reuse it for all further calls. With this change, only one connection is opened, and no tests have connection errors, even on my laptop.

```
2016-07-15T11:11:29.878-0700 I NETWORK  [initandlisten] connection accepted
from 127.0.0.1:52235 #8934 (3 connections now open)
2016-07-15T11:11:29.881-0700 I NETWORK  [conn8934] end connection 127.0.0.1
:52235 (2 connections now open)
2016-07-15T11:11:29.882-0700 I NETWORK  [initandlisten] connection accepted
from 127.0.0.1:52236 #8935 (3 connections now open)
2016-07-15T11:11:29.886-0700 I NETWORK  [conn8935] end connection 127.0.0.1:52236 (2 connections now open)
...
2016-07-15T11:12:59.484-0700 I NETWORK  [conn8707] end connection 127.0.0.1:50914 (1 connection now open)
2016-07-15T11:13:00.121-0700 I NETWORK  [conn8691] end connection 127.0.0.1:50865 (0 connections now open)
```
    
```
2016-07-15 11:05:41,863:DEBUG:androidUUID = e4bc20f6-af05-48b8-be81-1e9ebf460500, iosUUID
 = 831888b4-19dd-47a4-9217-6fb5da45d55b
2016-07-15 11:05:41,867:DEBUG:for sensor_config, config_fn = <function get_config at 0x10
ebfd398>
...
2016-07-15 11:12:58,729:DEBUG:While retrieving section_end_point from {u'section_start_da
tetime': datetime.datetime(2015, 1, 1, 1, 1, 1), u'user_id': u'test_user', u'section_end_
datetime': datetime.datetime(2015, 1, 1, 2, 1, 1), u'section_id': 3, u'_id': u'foo_1', u'
type': u'move'}, in is False
```

### DIFF
--- a/emission/core/get_database.py
+++ b/emission/core/get_database.py
@@ -4,34 +4,39 @@ import os
 import json
 from emission.net.int_service.giles import archiver
 
+_current_db = MongoClient('localhost').Stage_database
+
+def _get_current_db():
+    return _current_db
+
 def get_mode_db():
-    current_db = MongoClient().Stage_database
-    Modes=current_db.Stage_Modes
+    # #current_db = MongoClient().Stage_database
+    Modes= _get_current_db().Stage_Modes
     return Modes
 
 def get_moves_db():
-    current_db = MongoClient('localhost').Stage_database
-    MovesAuth=current_db.Stage_user_moves_access
+    # #current_db = MongoClient('localhost').Stage_database
+    MovesAuth= _get_current_db().Stage_user_moves_access
     return MovesAuth
 
 def get_habitica_db():
-    current_db = MongoClient('localhost').Stage_database
-    HabiticaAuth=current_db.Stage_user_habitica_access
+    # #current_db = MongoClient('localhost').Stage_database
+    HabiticaAuth= _get_current_db().Stage_user_habitica_access
     return HabiticaAuth
 
 def get_section_db():
-    current_db=MongoClient('localhost').Stage_database
-    Sections=current_db.Stage_Sections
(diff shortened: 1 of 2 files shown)


shankari:
+ couple of bug fixes to the metrics based on phone updates

shankari:
@yw374cornell - this was the fix!

PR #316, opened 2016-07-16
Move all logging to the main function

shankari:
Before this change, we logged the message, but did not log any timestamps, which makes it harder to see how long various operations took, correlate wi th logs on other servers, etc.
    
The logs were also configured individually in each test and some non-test modules, making it hard to change the log format easily. Finally, the logs were occasionally configured globally on module load. This meant that they would override any configurations in main.
    
I went through and unified the logging module configuration.

- modules do not have module level logging configurations. no exceptions.
  (scripts in `bin` are exceptions because they are intended to be more l ike
    scripts)
- modules that are invoked directly by the user (e.g intake, webapp) have
  existing configurations
- tests have a standard logging configuration defined in `emission.tests.common`
- all tests configure their logging in main using the standard logging configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,14 @@ note that the data is accessible by AWS admins by directly looking at the disk.
 In order to avoid this, you want to encrypt the disk. You can do this by:
 - using an encrypted EBS store, but this doesn't appear to allow you to specify
   your own encryption key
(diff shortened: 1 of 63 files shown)


shankari:
Also moved database references outside loops so that we don't have to get the collection reference over and over again (to add on to dc22f46b9016051bed9ff1892d8ee5e1e247373a) and a couple of other minor fixes.

PR #317, opened 2016-07-17
Fixed the issue with incorrect time range for data pulled from the server

yw374cornell:
Changed the API for getting public data to take in timestamps instead of dates.


shankari:
ok to test

PR #324, opened 2016-07-20
Habitica setup

juemura:


emission/net/api/cfc_webapp.py
```diff
@@ -518,6 +520,12 @@ def habiticaProxy():
     return habitproxy.habiticaProxy(user_uuid, method, method_url,
                                     method_args)
 # Data source integration END
+@post('/habiticaSetup')
+def habiticaSetup():
+    logging.debug("habitica setup request %s" % (request))
+    user_uuid = getUUID(request)
+    assert(user_uuid is not None)
+    return habitassist.user_setup(user_uuid)
```
shankari: why not just invoke this as part of `/habiticaRegister`?

emission/net/ext_service/habitica/setup.py
```diff
@@ -0,0 +1,41 @@
...
+  bike_habit_id = hab.create_habit(user_id, bike_habit)
+  walk_habit = {'type': "habit", 'text': "Walk", 'up': True, 'down': False, 'priority': 2}
+  walk_habit_id = hab.create_habit(user_id, walk_habit)
+
+
+def setup_party(user_id):
+  #parties = [{"leader": "Juliana", "group_id": "488cae51-aeee-4004-9fa0-dd4219a3a77e"}, {"leader": "Sunil", "group_id": "751e5f9a-bd2d-4c4c-ba81-6fb89bccdf5d"}, {"leader": "Shankari", "group_id": "93c35a70-f70e-4d6e-ac2b-3e1c81fedf0f"}]
+  parties = ["Juliana", "Sunil", "Shankari"]
+  group = random.randint(0, 2)
+  leader_val = list(edb.get_habitica_db().find({"habitica_username": parties[group]}))[0]
```
shankari: how do you know that there is an entry for these users in the `habitica_db()`? I guess we are committing to register before anybody else joins?

emission/net/api/cfc_webapp.py
shankari: That way, Sunil doesn't have to add another call from the UI....

emission/net/ext_service/habitica/setup.py
juemura: because these users already exist. Just pushed a script to create them in the production db.

shankari:
ok to test

bin/ext_service/create_party_leaders_in_db.py
```diff
@@ -0,0 +1,35 @@
...
+    su_uuid = edb.get_uuid_db().find_one({'user_email': su_email})['uuid']
+    logging.debug("Found Sunil's uuid %s" % su_uuid)
+    edb.get_habitica_db().insert({'habitica_id': "9ca9993a-3d84-434f-97af-3df2bd1a1def", 
+        'habitica_password': u"autogenerate_me", 
+        'habitica_token': "56977c97-0604-48a2-9572-59d7011b6948", 
+        'habitica_username': "Sunil", 
+        'metrics_data': {'bike_count': 0, 'last_timestamp': 1469020093, 'walk_count': 0}, 
+        'user_id': su_uuid})
+
+    sh_email = "shankari@eecs.berkeley.edu"
```
shankari: the email that I use for login in `shankari@berkeley.edu`

bin/ext_service/create_party_leaders_in_db.py
```diff
@@ -0,0 +1,35 @@
+import logging
+import emission.core.get_database as edb
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG)
+
+    ju_email = "ju.uemura@gmail.com"
+    ju_uuid = edb.get_uuid_db().find_one({'user_email': ju_email})['uuid']
+    logging.debug("Found Juliana's uuid %s" % ju_uuid)
+    edb.get_habitica_db().insert({'habitica_id': "e26ef1c3-d45e-464a-998c-729d107ecc42",
```
shankari: if this is what we are doing, don't we need `register` to be idempotent? I am not sure that habitica's register is idempotent. In particular, what happens when I go to the goals screen for the first time? I have never registered the user. Will it prompt me to register? Will that pass?

PR #325, opened 2016-07-20
different vs of habitica setup

juemura:
I guess with this we won't need to script to create the 3 party leaders, it'll just find you in habitica and return that. I need to do a check in setup_party to catch errors, but other than that this should be ok.
lmk what you think.

### DIFF
--- a/emission/net/ext_service/habitica/proxy.py
+++ b/emission/net/ext_service/habitica/proxy.py
@@ -3,9 +3,11 @@ import json
 import requests
 import logging
 import arrow
+import urllib2
 
 # Our imports
 import emission.core.get_database as edb
+import emission.net.ext_service.habitica.setup as assist
 
 key_file = open('conf/net/keys.json')
 key_data = json.load(key_file)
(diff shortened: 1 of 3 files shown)


juemura:
PS: @sunil07t I'm inviting new users to a party during registration and appending the group_id to the user_dict, so you'll get that right away. The same way you get, say, user['data']['username'] (or something like that), you'll get user['group_id']

shankari:
ok to test

emission/net/ext_service/habitica/proxy.py
```diff
@@ -14,29 +16,47 @@
...
-  habitica_user_table = edb.get_habitica_db()
-  habitica_user_table.insert({'user_id': our_uuid, 
-    'metrics_data': {'last_timestamp': arrow.utcnow().timestamp, 'bike_count': 0, 'walk_count': 0},
-    'habitica_username': username, 
-    'habitica_password': password, 
-    'habitica_id': user_dict['data']['_id'], 
-    'habitica_token': user_dict['data']['apiToken']})
+  user_dict = {}
+  #try to login and see if user is already in Habitica
+  try:
```
shankari: where's the `expect` for this `try`?

shankari: nvm I see it below there

emission/net/ext_service/habitica/proxy.py
```diff
@@ -14,29 +16,47 @@
...
+    habitica_user_table.insert({'user_id': our_uuid, 
+      'metrics_data': {'last_timestamp': arrow.utcnow().timestamp, 'bike_count': 0, 'walk_count': 0},
+      'habitica_username': username, 
+      'habitica_password': password, 
+      'habitica_id': user_dict['data']['_id'], 
+      'habitica_token': user_dict['data']['apiToken'],
+      'habitica_group_id': ""})
+  
+  #Do initial setup and append GroupId to the user
+  group_id = assist.user_setup(our_uuid)
```
shankari: this is outside the emission db check, right? Is it idempotent? 
Recall that the user may call `habiticaRegister` multiple times and our goal is to support that.

Right now, I think that if the user calls `habiticaRegister` a second time, she will be randomly re-assigned to a different group.

shankari:
looks good overall except the call to `user_setup()`.
You need to make it be idempotent too (update the user's party only if it is not already set)

shankari:
also fix the tests :)

PR #326, opened 2016-07-21
Habitica setup v4

juemura:
Not sure how you'll feel about this, but I moved everything setup related to proxy, and I kind of prefer this way. It works now, though I want to add additional checks at some point (check if our db is correct when user already exists, and check if excepts are working properly).
The script is optional, if we are the first ones to log in, then it should be alright, and habiticaRegister will register us in the db correctly. But otherwise we can just run the script and not worry about it.


emission/net/ext_service/habitica/proxy.py
```diff
@@ -3,17 +3,87 @@
...
+      
+    #Now save new user (user_dict) to our db
+    #Since we are randomly generating the password, we store it in case users 
+    #want to access their Habitica account from the browser
+    #Need to create a way from them to retrieve username/password
+    #metrics_data is used to calculate points based on km biked/walked
+    #last_timestamp is the last time the user got points, and bike/walk_count are the leftover km
+    habitica_user_table = edb.get_habitica_db()
+    habitica_user_table.insert({'user_id': our_uuid, 
+      'metrics_data': {'last_timestamp': arrow.utcnow().timestamp, 'bike_count': 0, 'walk_count': 0},
```
shankari: wait, isn't this called every time? Even if the correctly working case where the user is in the database and the user is in habitica? won't that always reset the counts?

shankari: shouldn't you create the user only if she wasn't in the emission database?
And you should update if the user was in emission but not in habitica. Otherwise, you will end up with two entries for the same user, right?

shankari:
ok to test

shankari:
Filed https://github.com/e-mission/e-mission-server/issues/327 to enhance tests

PR #328, opened 2016-07-21
Habitica integration

juemura:
hope this time it works!
updated db when creating new habitica user for user existing in our db
added script to create party leaders for test environment
fixed tests


bin/ext_service/create_party_leaders_in_db.py
```diff
@@ -0,0 +1,38 @@
+import logging
+import emission.core.get_database as edb
+
+if __name__ == '__main__':
```
shankari: For maintainability, this should call `create_party_leaders()`.
But if everything else passes, I will let this slide and let you clean up later.

emission/net/ext_service/habitica/proxy.py
```diff
@@ -22,27 +102,14 @@ def habiticaRegister(username, email, password, our_uuid):
...
   return user_dict
-  response = proxy.habiticaProxy(uid, 'DELETE', "/api/v3/user", {'password': password})
+
 
 def habiticaProxy(user_uuid, method, method_url, method_args):
   logging.debug("For user %s, about to proxy %s method %s with args %s" %
                 (user_uuid, method, method_url, method_args))
   user_query = {'user_id': user_uuid}
-  assert(edb.get_habitica_db().find(user_query).count() == 1)
+  assert(edb.get_habitica_db().find(user_query).count() > 0)
```
shankari: so we had instances where there was more than one entry? That sounds bad...

shankari: how do we know we are not reading an old, obsolete value, then?

emission/net/ext_service/habitica/create_party_leaders_script.py
```diff
@@ -0,0 +1,38 @@
+import logging
+import emission.core.get_database as edb
+
+def create_party_leaders():
+    logging.basicConfig(level=logging.DEBUG)
+
+    ju_email = "ju.uemura@gmail.com"
+    ju_uuid = edb.get_uuid_db().find_one({'user_email': ju_email})['uuid']
+    logging.debug("Found Juliana's uuid %s" % ju_uuid)
+    edb.get_habitica_db().insert({'habitica_id': "e26ef1c3-d45e-464a-998c-729d107ecc42",
```
shankari: If you plan to call this multiple times as part of the tests, then you either need to delete these users as part of tearDown, or convert this to an upsert.

emission/net/ext_service/habitica/proxy.py
shankari: I wonder if the error was caused by multiply created leader accounts.

shankari:
ok to test

emission/net/ext_service/habitica/proxy.py
juemura: That will happen if you run the script to create party leaders multiple times, for example. But yes, I need to go through this and treat edge cases more carefully.

PR #335, opened 2016-07-24
Added games tab and changed trip planner to point to emission url

sunil07t:
Game tab points to habitica server.


shankari:
ok to test

shankari:
@sunil07t, where's game.html?

sunil07t:
screen shot 2016-07-23 at 7 17 16 pm

This is how the nav bar looks for with the current commit.

sunil07t:
screen shot 2016-07-23 at 7 20 32 pm

I added nav button to each html and now it looks like this. Should I push this in this PR?

sunil07t:
screen shot 2016-07-23 at 8 26 53 pm
I added nav button to each html and removed the nav button from index.html and this is how it look. Should I push this?

sunil07t:
Sorry, I did not realized that I had not push game.html.

webapp/www/templates/game.html
```diff
@@ -0,0 +1,8 @@
+<ion-view view-title="Game">
+	<ion-content>
+		<ion-item>
+			<iframe src='http://54.159.38.241:3000/static/front' width="100%" height="550" frameborder="0" allowfullscreen>
```
shankari: you want to use https://em-game.eecs.berkeley.edu here. Otherwise, the game won't load because of https <---> http mismatch. That's what happening for the trip planner on the current production.

shankari:
> I added nav button to each html and removed the nav button from index.html and this is how it looks like. Should I push this?

Sure. Anything that makes it look better is good in my book :)

sunil07t:
Although the trip planner url is HTTPS it still has some HTTP content(PNG images) which causes a mixed content warning.
`Mixed Content: The page at 'https://em-planner.eecs.berkeley.edu/' was loaded over HTTPS, but requested an insecure image 'http://c.tile.thunderforest.com/transport/2/1/1.png'. This content should also be served over HTTPS.`
We can probably use the HTTPS version of those PNG. Is it possible to change that from our end?

shankari:
jenkins, test this please

sunil07t:
screen shot 2016-07-24 at 3 00 31 pm
E-mission faviocn on the browser tab.

shankari:
I have already merged this because I want to deploy tonight, but it is not clear to me that copy-pasting the menu in all the html files is required. Why not use `ion-tabs` instead? Or at least put them into a separate file and then use `ng-include` to include them so that they can be changed in one place?

shankari:
otherwise, looks good

sunil07t:
I think ion-tabs does not work on browsers because it shows tabs on the bottom of the page. I will use ng-include in the next PR.

shankari:
You can use `ion-tabs` with tabs on the top as well.
http://ionicframework.com/docs/components/#striped-style-tabs

> Optionally, also add tabs-top to position the tab at the top

PR #336, opened 2016-07-25
Add server side support for storing the "consent" rw document

shankari:
This was a fairly straightforward change.
The main issue was a bug with the metadata formatting workarounds on the server, which exposed a deeper bug https://github.com/e-mission/cordova-usercache/issues/17


shankari:
jenkins, test this please

Issue #330, opened 2016-07-22
Habitica autocheck crash on the server


shankari:
@juemura can you please take a look?

```
2016-07-22 18:18:32,717:INFO:**********UUID f955cff1-8fb8-3b42-8776-6d8874e5b90a: checking active mode trips to autocheck habits**********
2016-07-22 18:18:32,717:DEBUG:For user f955cff1-8fb8-3b42-8776-6d8874e5b90a, about to proxy GET method /api/v3/tasks/user?type=habits with args None
Traceback (most recent call last):
  File "bin/intake_stage.py", line 53, in <module>
    autocheck.auto_complete_tasks(uuid)
  File "/mnt/e-mission/e-mission-server/emission/net/ext_service/habitica/sync_habitica.py", line 66, in auto_complete_tasks
    reward_active_transportation(user_id)
  File "/mnt/e-mission/e-mission-server/emission/net/ext_service/habitica/sync_habitica.py", line 20, in reward_active_transportation
    bike_habit_id = proxy.create_habit(user_id, bike_habit)
  File "/mnt/e-mission/e-mission-server/emission/net/ext_service/habitica/proxy.py", line 170, in create_habit
    result = habiticaProxy(user_id, 'GET', get_habits_uri, None)
  File "/mnt/e-mission/e-mission-server/emission/net/ext_service/habitica/proxy.py", line 117, in habiticaProxy
    assert(edb.get_habitica_db().find(user_query).count() > 0)
AssertionError
```

shankari:
working around this for now by converting the assert to a log.

shankari:
This appears to be because there is no entry for the particular user in the habitica DB.

I've worked around this for now by using

```
-  assert(edb.get_habitica_db().find(user_query).count() > 0)
+  if (edb.get_habitica_db().find(user_query).count() > 0):
+    logging.warning("Found multiple entries for user %s" % user_uuid)
+    for entry in edb.get_habitica_db().find(user_query):
+      logging.warning("entry is %s" % entry)
+  else:
+    raise RuntimeError("no entry found for user")
```

and 

```
def reward_active_transportation(user_id):
   try:
      // all the code
   except RuntimeException:
      logging.debug("User %s information not logged in to habitica, skipping" % user_id)
```

shankari:
Steps to reproduce:
- purge all entries
- reload my data
- re-run intake pipeline

PR #338, opened 2016-07-27
Issue330

juemura:
fixed issue #330 by adding if to only run automatic reward if user exists in db. Changed asserts back to ==1 (rather than >0)


shankari:
ok to test

PR #339, opened 2016-07-27
New public device

yw374cornell:
Added a new public phone: Android5

### DIFF
--- a/request_public_data.py
+++ b/request_public_data.py
@@ -25,13 +25,15 @@ server_url = args.server_url
 if args.verbose:
 	logging.basicConfig(level=logging.DEBUG)
 
-from_ts = arrow.get(from_date).replace(tzinfo='local').timestamp
-to_ts = arrow.get(to_date).replace(tzinfo='local').timestamp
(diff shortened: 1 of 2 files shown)


shankari:
ok to test

shankari:
ok to test

request_public_data.py
```diff
@@ -45,9 +47,9 @@
 
 for index, entry_list in enumerate(phone_list):
 	if index < 4:
-		logging.debug("iphone" + str(index%4+1) + " first entry:")
+		logging.debug("iphone" + str(index+1) + " first entry:")
 	else:
-		logging.debug("android" + str(index%4+1) + " first entry:")
+		logging.debug("android" + str(index-3) + " first entry:")
```
shankari: why is this `index-3` instead of `index+1`?

shankari: merging anyway since this is in the client script and not on the server.

yw374cornell: it's `index-3` because this for loop is iterating through the concatenated list of iphone and android, so the index for android starts at 4

PR #341, opened 2016-08-02
Google Play and App Store badges that links to Emission mobile app

sunil07t:
Also changed View on GitHub button to GitHub Logo.


shankari:
ok to test

Issue #348, opened 2016-08-14
Raw segmentation skips the walking part of a walk + train + walk trip


shankari:
In #345 we fixed the cleaned trip detection. But while looking more closely at it, it looked like there was an issue with raw segmentation as well.

These are the points detected for the section from the train station to Oakland DOT and back.

<img width="538" alt="screen shot 2016-08-14 at 4 07 52 pm" src="https://cloud.githubusercontent.com/assets/2423263/17652788/29159f9c-623b-11e6-9177-ea9af217c11c.png">

It is not clear why we ended the trip on broadway and 11th (the train station) instead of getting the full multi-modal trip.

![simulator screen shot aug 14 2016 4 23 44 pm](https://cloud.githubusercontent.com/assets/2423263/17652808/83b5870a-623b-11e6-9dd9-637e1545dafa.png)

shankari:
After a big gap, we started getting points again.

```
2016-08-11 20:08:10,282:DEBUG:------------------------------2016-08-11T10:13:00.585000-07:00------------------------------
2016-08-11 20:08:10,287:DEBUG:last5MinsDistances = [ 4084.14912455  3451.52647937  1504.04296114  1337.49816784   446.65066228] with length 5
2016-08-11 20:08:10,292:DEBUG:last10PointsDistances = [ 4583.49305378  4549.46266567  4540.59988907  4573.67651691  4084.14912455
2016-08-11 20:08:10,292:DEBUG:len(last10PointsDistances) = 10, len(last5MinsDistances) = 5
2016-08-11 20:08:10,292:DEBUG:last5MinsDistances.max() = 4084.14912455, last10PointsDistance.max() = 4583.49305378
2016-08-11 20:08:10,293:DEBUG:------------------------------2016-08-11T11:19:10.269000-07:00------------------------------
2016-08-11 20:08:10,295:DEBUG:last5MinsDistances = [] with length 0
2016-08-11 20:08:10,300:DEBUG:last10PointsDistances = [ 31277.26212486  31250.32917744  31289.01878554  30906.9231772
2016-08-11 20:08:10,300:DEBUG:len(last10PointsDistances) = 10, len(last5MinsDistances) = 0
2016-08-11 20:08:10,300:DEBUG:Too few points to make a decision, continuing
```

And at this point, we got a cluster of points that are only 5-10 secs apart, not 30 secs. This is probably because I was using the navigation on the phone to see where I was going.

```
[...]

PR #350, opened 2016-08-15
We have enough points to make a decision only when they span at least…

shankari:
… 5 mins

Before this change, we had two checks to determine whether we had enough
points to make a decision

- whether we had at least 10 points
- whether any of them were in the past 5 mins

If the points that we received were 30 secs apart, as we expected, this would
work just fine. But if they were more frequent, as when navigating while
walking, then we detect a trip end way before 5 mins.

In this change, we also add an additional check to ensure that the points from
the last 5 mins query actually span 5 mins. With this check, I think that we
can remove the last 10 points check, but I want to wait until we have a more
principled checker to confirm before making a subtractive change.


shankari:
This fixes #348.

| more sections | at oakland | 
| ----------- | ------------------- |
| simulator screen shot aug 14 2016 2 57 55 pm | simulator screen shot aug 14 2016 4 23 44 pm |

PR #351, opened 2016-08-16
Don't check for continuation when we are starting a new trip

### LINKED ISSUE #302: Short round-trips have the outgoing trip skipped
Kids took my phone, went to the library and came back. Return trip showed up, outgoing trip did not.
Return trip was 12:53 to 12:57.

trip coming back: `57798b0488f66332d48de971`

When we entered the CLEAN_RESAMPLING stage, we had:

```
2016-07-03 22:00:35,252:DEBUG:Returning entry with length 3 result
2016-07-03 22:00:35,253:DEBUG:Considering place 57798b0288f66332d48de961: 2016-07-03T12:12:08.010000-07:00 -> 2016-07-03T12:17:06.872000-07:00
2016-07-03 22:00:35,253:DEBUG:Considering place 57798b0288f66332d48de963: 2016-07-03T12:19:59.140000-07:00 -> 2016-07-03T12:53:17.999000-07:00
2016-07-03 22:00:35,253:DEBUG:Considering place 57798b0288f66332d48de965: 2016-07-03T12:57:55-07:00 -> None

2016-07-03 22:00:35,253:DEBUG:Considering trip 57798b0288f66332d48de960: 2016-07-03T12:10:46.617000-07:00 -> 2016-07-03T12:12:08.010000-07:00
2016-07-03 22:00:35,254:DEBUG:Considering trip 57798b0288f66332d48de964: 2016-07-03T12:53:17.999000-07:00 -> 2016-07-03T12:57:55-07:00
```

So the earlier trips look like they were stripped out as part of the cleaning stage. Why is this?

Started with the the trip leaving at 12:10:46 from [-122.0838776, 37.3904947].
This is already the library.

The matching place before this is:

```
2016-07-03 22:00:35,271:DEBUG:retrieved object Entry({u'data': {u'enter_fmt_time': u'2016-07-01T19:45:33.126000-07:00', u'exit_fmt_time': u'2016-07-03T12:10:46.617000-07:00', u'ending_trip': ObjectId('57773c6c88f663773188f683'), u'starting_trip': ObjectId('57798b0288f66332d48de960'), u'location': {u'type': u'Point', u'coordinates': [-122.086378, 37.3907813]}, u'enter_ts': 1467427533.126, u'duration': 145513.49100017548, u'exit_ts': 1467573046.617}, u'_id': ObjectId('57773c6c88f663773188f684'), u'metadata': {u'write_fmt_time': u'2016-07-01T21:00:44.551724-07:00', u'write_ts': 1467432044.551724, u'time_zone': u'America/Los_Angeles', u'platform': u'server', u'key': u'segmentation/raw_place'}}) and added to id_map
```

So the raw_place was itself broken, the error is not in the cleaning stage.

Here's why the raw_place is broken. The main reason is that the differences between points is too small, probably because people were walking slowly or waiting at the signal or whatever.

```
2016-07-03 22:00:34,154:INFO:Points
AttrDict({u'loc': {u'type': u'Point', u'coordinates': [-122.0860081, 37.3918007]},
u'fmt_time': u'2016-07-03T12:02:49.571000-07:00',
'metadata_write_ts': 1467572573.296,
u'ts': 1467572569.5710001,
'_id': ObjectId('57797daafe930a83ace95b58'),
u'accuracy': 20.533999999999999})
AttrDict({u'loc': {u'type': u'Point', u'coordinates': [-122.08576, 37.3916086]},
u'fmt_time': u'2016-07-03T12:03:11-07:00',
'metadata_write_ts': 1467572590.678,
u'ts': 1467572591.0,
'_id': ObjectId('57797daafe930a83ace95b5a'),
u'accuracy': 10.0})
are within the distance filter and only 1 min apart so part of the same trip
```
But the first point should have been outside the geofence filter because of the value of the geofence radius....

However, it turns out that in this case, we compared the first point to a point that occurred after it!

```
2016-07-03 22:00:34,151:DEBUG:------------------------------2016-07-03T12:02:49.571000-07:00------------------------------
2016-07-03 22:00:34,151:DEBUG:Appending currPoint because the current start point is None
2016-07-03 22:00:34,152:DEBUG:Comparing with prev_point = AttrDict({u'loc': {u'type': u'Point', u'coordinates': [-122.0864586, 37.3910005]},
u'fmt_time': u'2016-07-03T13:00:53-07:00',
'metadata_write_ts': 1467576051.553,
'_id': ObjectId('57797daafe930a83ace95c22')})
```

why is the prev_point for `2016-07-03T12:02:49.571000-07:00` set to `2016-07-03T13:00:53-07:00`?

`2016-07-03T13:00:53-07:00` looks like it is the last point read in this set...

```
2016-07-03 22:00:33,915:DEBUG:check_prior_duplicate called with size = 73, entry = 78
2016-07-03 22:00:33,919:INFO:Found existing filtered location for entry at index = 78, id = 57797daafe930a83ace95c21, ts = 1467576053.0, fmt_time = 2016-07-03T13:00:53-07:00, skipping
2016-07-03 22:00:33,921:INFO:For stage PipelineStages.ACCURACY_FILTERING, last_ts_processed = 2016-07-03T20:00:55.908000
```

I see why, and it's actually a pretty serious and annoying bug. 

Here's the code for comparing against the prev point. Note that for the first point, `just_ended = True`, and `idx = 0`, so we will end up comparing the first point with the last point from that batch. As long as the first and last points are sufficiently far apart, we are fine. But if the first and last points overlap, as they would for a short round trip, then we will skip the first leg of the trip.

```
        just_ended = True
        for idx, row in filtered_points_df.iterrows():
            currPoint = ad.AttrDict(row)
            currPoint.update({"idx": idx})
            logging.debug("-" * 30 + str(currPoint.fmt_time) + "-" * 30)
            if curr_trip_start_point is None:
                logging.debug("Appending currPoint because the current start point is None")
                # segmentation_points.append(currPoint)

            if just_ended:
                prev_point = ad.AttrDict(filtered_points_df.iloc[idx - 1])

```

shankari:
This fixes #302. Sometimes, we get some points beyond the expected end of a
trip. We want to ignore them so we treat them as a continuation of the current
trip. However, if the points occur at the beginning of a trip, we don't want
this to happen.

Note that we have `justEnded = true` at the beginning of a trip, to have a
ingle control flow for the start and restart cases. So we need to have an
additional check for the idx before we detect a continuation.

### DIFF
--- a/emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
+++ b/emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
@@ -74,25 +74,9 @@ class DwellSegmentationTimeFilter(eaist.TripSegmentationMethod):
                 # segmentation_points.append(currPoint)
 
             if just_ended:
-                # Normally, at this point, since the logic here and the
-                # logic on the phone are the same, if we have detected a trip
-                # end, any points after this are part of the new trip.
-                #
-                #
-                # However, in some circumstances, notably in my data from 27th
-                # August, there appears to be a mismatch and we get a couple of
-                # points past the end that we detected here.  So let's look for
-                # points that are within the distance filter, and are at a
-                # delta of 30 secs, and ignore them instead of using them to
-                # start the new trip
-                prev_point = ad.AttrDict(filtered_points_df.iloc[idx - 1])
-                logging.debug("Comparing with prev_point = %s" % prev_point)
-                if pf.calDistance(prev_point, currPoint) < self.distance_threshold and \
-                    currPoint.ts - prev_point.ts <= 60:
-                    logging.info("Points %s and %s are within the distance filter and only 1 min apart so part of the same trip" %
-                                 (prev_point, currPoint))
+                if self.continue_just_ended(idx, currPoint, filtered_points_df):
                     continue
-                # else: 
+                # else:
                 sel_point = currPoint
                 logging.debug("Setting new trip start point %s with idx %s" % (sel_point, sel_point.idx))
                 curr_trip_start_point = sel_point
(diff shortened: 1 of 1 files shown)


shankari:
|  Fixed optometrist trip | Also made trip to Eagle Park much better |
| ---------------------- | -------------------- |
| simulator screen shot aug 15 2016 10 10 52 pm | simulator screen shot aug 15 2016 10 11 05 pm |

PR #355, opened 2016-08-17
Fix sort key for the timeline

### LINKED ISSUE #354: trips are out of order in the timeline
see examples in https://github.com/e-mission/e-mission-server/issues/302#issuecomment-240219847

This may be a recent regression caused by trip segmentation changes.

shankari:
Although we search for timeline entries using the local date, we want to sort by the timestamp since that is fully ordered while the local date is:
- only partially ordered
- split into components

This fixes #354


shankari:
Fixes

| Jun 20 | Jun 21 | Jul 3 | 
| ----- | ---- | --- |
| simulator screen shot aug 16 2016 8 57 39 pm | simulator screen shot aug 16 2016 8 57 51 pm | simulator screen shot aug 16 2016 8 59 59 pm |

Issue #352, opened 2016-08-16
Trips are sometimes incorrectly "joined" even if the gap between sections is close to an hour long


shankari:
We have just started our trip, yay! Now we still don't have enough points to make a decision.

```
2016-08-15 12:10:17,370:DEBUG:------------------------------2016-08-10T09:54:31-07:00------------------------------
2016-08-15 12:10:17,372:DEBUG:last5MinsDistances = [ 52.82764235] with length 1
2016-08-15 12:10:17,374:DEBUG:last10PointsDistances = [ 52.82764235   0.        ] with length 2, shape (2,)
2016-08-15 12:10:17,374:DEBUG:len(last10PointsDistances) = 2, len(last5MinsDistances) = 1
2016-08-15 12:10:17,375:DEBUG:last5MinsTimes.max() = 31.5750000477, time_threshold = 300
2016-08-15 12:10:17,375:DEBUG:Too few points to make a decision, continuing
...
2016-08-15 12:10:17,435:DEBUG:------------------------------2016-08-10T09:58:13-07:00------------------------------
2016-08-15 12:10:17,440:DEBUG:last5MinsDistances = [ 329.2908086   280.89265787  245.2010695   184.71648394  189.02429048
  142.74543776   87.77772771   43.01575703   56.99527453] with length 9
2016-08-15 12:10:17,444:DEBUG:last10PointsDistances = [ 329.2908086   280.89265787  245.2010695   184.71648394  189.02429048
  142.74543776   87.77772771   43.01575703   56.99527453    0.        ] with length 10, shape (10,)
2016-08-15 12:10:17,446:DEBUG:len(last10PointsDistances) = 10, len(last5MinsDistances) = 9
2016-08-15 12:10:17,446:DEBUG:last5MinsTimes.max() = 253.575000048, time_threshold = 300
2016-08-15 12:10:17,447:DEBUG:Too few points to make a decision, continuing
```

Now we finally have enough points

```
2016-08-15 12:10:17,447:DEBUG:------------------------------2016-08-10T09:58:52-07:00------------------------------
2016-08-15 12:10:17,454:DEBUG:last5MinsDistances = [ 368.82243137  323.60485982  289.21766655  230.01780163  235.70421315
  192.31601923  144.97574662   96.48072959   89.58728643   59.109117  ] with length 10
2016-08-15 12:10:17,457:DEBUG:last10PointsDistances = [ 323.60485982  289.21766655  230.01780163  235.70421315  192.31601923
[...]

PR #357, opened 2016-08-17
Handle mismatch between phone and server trip end detection

shankari:
This is a partial fix for #352. If there is a mismatch between phone and
server trip end detection, then we can end up with "joined" trips. In order to
work around that, we detect if there is a large gap between two points and if
there is, we detect that a visit has started.

More discussion is in the bug, but we also check the distance covered in that
gap (to ensure that we don't truncate trips with poor quality) and ensure that
when we detect a trip end in this new way, we also start a new trip at the
same time.


shankari:
| Optometrist trip is now split | Oakland trip is not split | Trip to Radio Shack is complete and is not split |
| ------------ | ------------ | ------------ |
| simulator screen shot aug 17 2016 4 27 47 pm | simulator screen shot aug 17 2016 4 28 52 pm | simulator screen shot aug 17 2016 4 28 11 pm |

PR #366, opened 2016-08-20
Cover the case in which we get a sync right after a phone-server mism…

shankari:
…atch

This is the second fix for #352. More detail is in the bug, but
basically, this is 2(i) from
https://github.com/e-mission/e-mission-server/issues/352#issuecomment-240526292.

When we have finished looking at all points, if we haven't ended a trip, then:
(i) if there is a transition, end the trip (80%)

Also, enhance the test cases to check for this. In particular, we now:
- load the data into two chunks and run the intake pipeline after each. This
  simulates a sync between the chunks
- add a new check for "almost equal". This is needed because when we sync
  between two trips, and use the transition information to end the trip, the
  set of 10 points that are used to determine the end point are very slightly
  different from the case where we end the trip because of the large gap. This
  is because if when we use the transition information, the currPoint is the
  end point, while when we use the gap, the prevPoint is the end point. So if
  the trip ends at point 22, then in the first case, we will look at points 13 to
  22 while in the second case, we will look at points from 14 to 23.
  I made some weak attempts to fix this, e.g. by dropping the last point in the
  second case. But then we still had points 14 to 22, not 13 to 22. Fixing that
  would have increased the complexity without significantly affecting accuracy.
  So I use the new almost_equal method with a fuzz factor instead.
- load data from the prev day as well so that we would have at least some trips
  even when the trip to the optometrist was dropped. If we don't do that, then
  we don't have *any* trips during the first run and we hit this condition
  ```
    elif len(segmentation_points) == 0:
        # no new segments, no need to keep looking at these again
        logging.debug("len(segmentation_points) == 0, early return")
        epq.mark_segmentation_done(user_id, None)
   else:
  ```

With all these conditions in place, the new test fails without the code fix

```
FAIL: testAug10MultiSyncEndDetected (__main__.TestPipelineRealData)
  File "emission/tests/analysisTests/intakeTests/TestPipelineRealData.py", line 272, in testAug10MultiSyncEndDetected
    ad.AttrDict(ground_truth).data, time_fuzz=60, distance_fuzz=100)
  File "emission/tests/analysisTests/intakeTests/TestPipelineRealData.py", line 92, in compare_approx_result
    self.assertEqual(len(result), len(expect))
AssertionError: 6 != 7

```

and passes with the code fix

```
Ran 1 test in 14.797s

OK
```

### DIFF
--- a/emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
+++ b/emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
@@ -2,6 +2,7 @@
 import logging
 import attrdict as ad
 import numpy as np
+import pandas as pd
 import datetime as pydt
 
 # Our imports
@@ -50,6 +51,8 @@ class DwellSegmentationTimeFilter(eaist.TripSegmentationMethod):
         segmentation points.
         """
         filtered_points_df = timeseries.get_data_df("background/filtered_location", time_query)
+        transition_df = timeseries.get_data_df("statemachine/transition", time_query)
+        logging.debug("transition_df = %s" % transition_df)
 
         if len(filtered_points_df) == 0:
             self.last_ts_processed = None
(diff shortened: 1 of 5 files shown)


shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #368, opened 2016-08-21
Add a test case for the fix for 2 (i) on iOS

shankari:
Although this had already been "fixed" after I encountered it back in Feb, I found two issues while comparing a run with the data loaded in one versus two chunks.

The first was #302, but on iOS.
The second was an issue with starting the trip when the previous trip end is
detected, as documented in
https://github.com/e-mission/e-mission-server/issues/352#issuecomment-241240459

Fixed both issues - the tests now pass.

```
2016-08-20 23:00:40,236:DEBUG:140735159468032:--------------------
2016-08-20 23:00:40,236:DEBUG:140735159468032:====================
2016-08-20 23:00:40,242:DEBUG:140735159468032:Clearing related databases
.
Ran 1 test in 10.904s
```

So this has turned into additional fixes for #352, just on iOS


shankari:
jenkins, test this please

shankari:
jenkins, test this please

Issue #288, opened 2016-06-21
Fix time/distance calculations if many points at the beginning are missing


shankari:
Reported by @gtfierro.

Trip to Soda was 0.15 km, 3 mins
Trip from Soda was 1km, 10 mins

Probably because we start distance calculation from the first point in the trip, not the place.
If we want to fix this for the time too (and we should) need to extrapolate leaving time from speed for the rest of the trip.

shankari:
I loaded data for Gabe for Jun 14-17 and Jun 21, and did not find a huge difference between the trip to Soda and from Soda.

Unfortunately, I did not record the date/time of these trips.

So the alternatives are:
- this is a different issue, such as #345. That was the problem with the trip to the train station to pick to Anirudh - we skipped the trip and then joined home to the station. After the fix for #345, we now see the trip to the train station as well
- the issue is from a day that I didn't record

I then looked at iOS, since it does not support a geofence radius and frequently generates points that are fairly far away.

While looking at the example that motivated the prior fixes to the section segmentation on iOS (https://github.com/e-mission/e-mission-server/pull/233), which is on 22 Feb 2016 from phone `c528bcd2-a88b-3e82-be62-ef4f2396967a` (iphone-3), I found several examples for short trips.

| Kaiser to Cognition is 75m? | To bookstore 356m, from bookstore 737m |
|---------------------|--------------------|
| ![simulator screen shot aug 22 2016 12 44 47 pm](https://cloud.githubusercontent.com/assets/2423263/17869054/5547bd76-6866-11e6-9bf6-25aba88d682f.png) | ![simulator screen shot aug 22 2016 12 45 21 pm](https://cloud.githubusercontent.com/assets/2423263/17869053/55443bb0-6866-11e6-903b-5e219be2b83b.png) |

shankari:
So the castro street one is because we only exited the geofence at Yosemite, three quarters of the way there.

```
>>> filtered_data_df[["fmt_time", "ts", "latitude", "longitude"]]

 	fmt_time 	ts 	latitude 	longitude
6 	2016-02-22T10:06:20.206589-08:00 	1456164380.207 	37.387 	-122.084
7 	2016-02-22T10:06:21.940137-08:00 	1456164381.940 	37.387 	-122.083
8 	2016-02-22T10:07:00.000028-08:00 	1456164420.000 	37.386 	-122.083
```

| Points for Kaiser trip | Expected value (418m/3 mins) v/s (75m/40 secs) |
[...]

PR #371, opened 2016-08-22
Add in a test case for short trips that will let us test the extrapol…

shankari:
…ation

To the beginning of the trip


shankari:
This will help us test #288.

PR #373, opened 2016-08-25
Fix the starting point location and time through extrapolation

### LINKED ISSUE #288: Fix time/distance calculations if many points at the beginning are missing

shankari:
This fixes #288. It also removes a long-standing TODO.

Basically, because we use a geofence and are only notified when we leave the
geofence, we don't actually know when the user really started their trip. We
were hoping to mitigate the resulting error by setting a short geofence
radius, but;
- there is no way to set a geofence radius on iOS
- even on android, the geofence radius does not appear to be honored. Although
  we have set the radius to 100, we almost always get the first point 200m out
  from the previous point.

This is particularly pronounced for short trips on iOS where we frequently get
the first point after half the trip is done. See examples of trip to/from
school and to Cognition.

The ipython notebook documents our attempts to fix it. We first tried to build
a model and extrapolate the start point. Unfortunately, that generated weird
results (see notebook). Then we tried a simpler technique in which we used the
median speed to estimate the time to travel the geofence distance.

That seems to work much better (examples in notebook and in github issue).

Related minor changes:
- since we now fix this in the cleaned trip, we don't need to hack around it
  while converting to geojson. Removing the hack. Yay!
- since we generate a new "fake" cleaned point for the extrapolated start,
  several location fields become WORM.
- regenerate ground truth for all the real tests


shankari:
jenkins, test this please

shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #374, opened 2016-08-26
Have the metrics return both the result and the last section processe…

shankari:
…d to generate that result

This allows us to use the metrics to track progress towards goals.
See https://github.com/e-mission/e-mission-server/issues/321#issuecomment-240222776 for more details.

PR #375, opened 2016-08-26
Game

juemura:
fixed habitica autocheck to get timestamp from last trip processed instead of last time that the pipeline ran.


shankari:
@juemura, the check is a bit more complex. If there were no matching trips (e.g at night), then I return `None` to represent that it is unchanged (that's what I do for all the other pipeline states). You need to check for that, otherwise the `None` will be stored in the database.

Using the standard pipeline states would have handled this for you automagically!

shankari:
https://github.com/e-mission/e-mission-server/pull/374/commits/cda4821aeb161efc9cd71b09edaf91b58fff6a8b

shankari:
@juemura, this is the commit that commented out the failing tests.
You will need to restore them after making sure that they pass with your changes

shankari:
ok to test

Issue #334, opened 2016-07-23
Fix regression that removed smoothing


shankari:
It looks like our smoothing algorithm has not been running since the big refactor of the query framework.

In particular, the query for selecting sections to smooth looks like this:

```
2016-07-23 19:35:15,702:DEBUG:curr_query = {'$or': [{'metadata.key': 'segmentation/raw_se
ction'}], 'end_ts': {'$lte': 1469302510.675839}, 'user_id': UUID('f8cdc7cd-8da3-373a-9079
-b9d49d69b947')}, sort_key = end_ts
```
which is obviously not matched with anything, since `end_ts` is in `data`.

shankari:
It looks like smoothing has never run successfully since the last reset, so if we fix this, we don't have to reset the pipeline for it. But we need to reset the cleaning and resampling stage, because otherwise, it won't take the smoothing into effect.

PR #381, opened 2016-09-03
Handle air trips as well

### LINKED ISSUE #380: Add support for air trips
I have resisted classifying modes using techniques other than the motion activity on the phone because I want to do the one technique that will fix them all.

The problem is that we are frequently missing location data on airplane flights, either because the phone is airplane mode, or turned off. Since we do trip segmentation first and section segmentation second, we end up with trips that have very little motion activity data, and the data that is present is often for walking.

So we end up with trips in which we walk 3000km in 6 hours, which throws off the metrics, and the points for the game.

| Trip 1 | Trip 2 | Trip 3 |
| --------------------- | -------------- | ----------- |
| simulator screen shot aug 28 2016 9 28 43 pm | simulator screen shot aug 28 2016 7 32 58 am | simulator screen shot aug 28 2016 9 48 56 pm |

shankari:
Or at least to the point that they are no longer classified as "walk" or "bike".
This fixes #380.


shankari:
Note that in https://github.com/e-mission/e-mission-server/pull/381/commits/4b5c4b21966c7f9e1ea7e82c64a4a4c3b4d8e6ce, I had the following comment:

> Note that the current implementation loads the data as multiple users and runs the pipeline separately - the result is slightly different if all data is loaded for a single user then the first trip on the 4th gets smushed with the last trip on the 27th. Will file a separate issue for that and fix it separately.

I am not sure that it needs to be fixed. It is basically a worst case scenario because our tracking started on another island (so a very long distance) after not quite enough time to detect a trip end. If we had loaded for the 5th alone, it would have still worked. And if we had loaded for the day of the plane trip, we would have detected it as a plane trip.

The only way to work around that is to check for a reboot transition during the trip, and it is unclear how that works on iOS. Given that it is unlikely that people are going to leave their phones for long periods of time when they are not in a plane, we can probably live with this, and just allow users to flag incorrect trips in the rare case that something like this happens.

shankari:
This failure is in the metrics tests.

```
FAIL: testCountLocalDateMetrics (__main__.TestMetrics)
  File "emission/tests/netTests/TestMetrics.py", line 105, in testCountLocalDateMetrics
    user_met_result[0].IN_VEHICLE + 3) # 21s has one bike trip
AssertionError: 6 not greater than or equal to 8

Ran 1 test in 35.789s
```

On closer examination, this is because the aggregate data now has two AIR trips.

```
{'aggregate_metrics': [ModeStatTimeSummary({'ON_FOOT': 10, 'BICYCLING': 5, 'nUsers': 2, 'fmt_time': u'2015-08', 'ts': 1438412400, 'AIR': 2, 'local_dt': LocalDate({'timezone': u'America/Los_Angeles', 'year': 2015, 'month': 8}), 'IN_VEHICLE': 6})], 'user_metrics': [ModeStatTimeSummary({'ON_FOOT': 6, 'BICYCLING': 4, 'nUsers': 1, 'fmt_time': u'2015-08', 'ts': 1438412400, 'local_dt': LocalDate({'timezone': u'America/Los_Angeles', 'year': 2015, 'month': 8}), 'IN_VEHICLE': 5})]}
```

And on closer examination (see below) that is because of giant zigzags in the location collection which cause trips to be mapped to AIR. But they are not actually AIR trips. But because our smoothing is broken (see #334), they are indistinguishable from them.

| Giant zigzags | Otherwise normal trip to Berkeley | Otherwise normal trip from Berkeley |
| -------------  | ---------------------------------  | -----------------------------------   |
| simulator screen shot sep 2 2016 11 06 50 pm | simulator screen shot sep 2 2016 11 07 08 pm | simulator screen shot sep 2 2016 11 07 29 pm |

Let's just fix the test for now. We will have to regenerate all ground truth after the smoothing fix anyway.

PR #386, opened 2016-09-07
Converting gmail account to UUID

wyawen:
**getUUIDFromEmail**("ucb.sdb.iphone.1@gmail.com") returns None, is this not the right way to pass in the argument?

### DIFF
--- a/emission/core/wrapper/publicdevice.py
+++ b/emission/core/wrapper/publicdevice.py
@@ -1,14 +1,13 @@
 import logging
 import emission.core.wrapper.wrapperbase as ecwb
 
-class Location(ecwb.WrapperBase):
-    props = {"is_public": ecwb.WrapperBase.Access.RO,  # 1 if the phone is (or was once) registered as a public device
-             "start_ts": ecwb.WrapperBase.Access.RO, # timestamp of when the phone becomes public 
-             "end_ts": ecwb.WrapperBase.Access.RO,   # timestamp of when the phone is no longer public 
+class Publicdevice(ecwb.WrapperBase):
+    props = {"ts": ecwb.WrapperBase.Access.RO  # timestamp of when the phone becomes public (first entry)
+                                               # or when the phone is no longer public (second entry)
              }     
     enums = {}
     geojson = []
-    nullable = ["start_ts", "end_ts"]
+    nullable = []
     local_dates = []
 
     def _populateDependencies(self):
--- a/emission/core/wrapper/publicdevice.py
+++ b/emission/core/wrapper/publicdevice.py
@@ -1,0 +1,15 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class Location(ecwb.WrapperBase):
+    props = {"is_public": ecwb.WrapperBase.Access.RO,  # 1 if the phone is (or was once) registered as a public device
+             "start_ts": ecwb.WrapperBase.Access.RO, # timestamp of when the phone becomes public 
+             "end_ts": ecwb.WrapperBase.Access.RO,   # timestamp of when the phone is no longer public 
+             }     
+    enums = {}
+    geojson = []
+    nullable = ["start_ts", "end_ts"]
+    local_dates = []
+
+    def _populateDependencies(self):
+        pass
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -389,9 +389,7 @@ def getCarbonCompare():
 def getPublicData():
   ids = request.json['phone_ids']
   uuids = map(lambda id: __getUUIDFromEmail__(id), ids)
-  print ids[0]
-  print uuids[0]
-  print type(uuids[0])
+  
   from_ts = request.query.from_ts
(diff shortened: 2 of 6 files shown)


shankari:
That is the right way, but it returns None because you haven't registered with that email to your local dev server so there is no email -> UUID mapping in your local database :)

To create that mapping, use the script `bin/create_test_user.py`
e.g.

```
./e-mission-py.bash bin/create_test_user.py ucb.sdb.iphone.1@gmail.com
```

yw374cornell:
Hi Shankari, you are right! Running the script you mentioned above solved my problem. So now I have the gmail accounts associated with the public phones listed in `request_public_data.py`. 

Since people's gmail address is basically public, I must add another public API now for registering a phone as a public device to prevent people from freely accessing others' data. Could you show me how to check if a password entered by the user matches the real one? Thanks!

shankari:
@yw374cornell, we never see the user password. And that is as it should be! You don't want me to know your gmail password, do you? Instead we get an OAuth token from google and use it to login. 

The way to mark a phone (effectively an user) as public is to define a `markAsPublic` method in the server, and call it from the phone with the standard authentication technique (the same one that we use for storing data, getting the timeline, etc). Then, you would probably invoke the `markAsPublic` from a control in the profile screen.

yw374cornell:
Hi Shanakri, I added a new wrapper class for keeping track of when a phone becomes public, and also made corresponding changes in the `requestPublicData` API to check if the phone requested is public. 
I'm not sure what is the proper way to update this timeseries in the `registerPublic` and `unregisterPublic` APIs, could you help clarify? Thank you!

yw374cornell:
Hi @shankari maybe you didn't see my last comment?

shankari:
I did, but I've been busy all of today working on scalability to 500 users, and I'm mucking with the real server so I have to be careful not to lose data. Will respond after I finish current set of updates.
https://github.com/e-mission/e-mission-server/issues/407#issuecomment-248469599

yw374cornell:
Sure, thank you!

emission/core/wrapper/publicdevice.py
```diff
@@ -0,0 +1,15 @@
+import logging
+import emission.core.wrapper.wrapperbase as ecwb
+
+class Location(ecwb.WrapperBase):
```
shankari: Location -> Publicdevice

emission/core/wrapper/publicdevice.py
```diff
@@ -0,0 +1,15 @@
...
+    props = {"is_public": ecwb.WrapperBase.Access.RO,  # 1 if the phone is (or was once) registered as a public device
```
shankari: isn't this implied by the metadata.key?

emission/net/api/cfc_webapp.py
```diff
@@ -388,32 +389,59 @@ def getCarbonCompare():
...
+  uuids = map(lambda id: __getUUIDFromEmail__(id), ids)
+  
   from_ts = request.query.from_ts
   to_ts = request.query.to_ts
 
-  time_range = estt.TimeQuery("metadata.write_ts", float(from_ts), float(to_ts))
-  time_query = time_range.get_query()
-  
+  public_queries = map(lambda id: {'user_id': id, "metadata.key":  "eval/public_device"}, uuids)
+  public_ranges = map(list(edb.get_timeseries_db().find(q))["data"], public_queries)
```
shankari: I would recommend using functions from `emission.storage.timeseries` instead of rewriting your own queries. It will be easier to switch to other timeseries (e.g BtrDB) if we have a clearly defined interface that all modules use.

emission/net/api/cfc_webapp.py
```diff
@@ -388,32 +389,59 @@ def getCarbonCompare():
 @get('/eval/publicData/timeseries')
```
shankari: Also, this is getting to be pretty big now. I think that you should move the implementation to a separate file/module, implemented in emission (e.g. in `emission/net/api/eval.py`?) The goal is have the bottle file be a thin shim over real functionality in the api classes. That way we can switch to a different frontend later.

emission/net/api/cfc_webapp.py
```diff
@@ -388,32 +389,59 @@ def getCarbonCompare():
...
-    data_list = map(lambda q: list(edb.get_timeseries_db().find(q).sort("metadata.write_ts")), user_queries)
+    data_list = map(lambda q: list(edb.get_timeseries_db().find(q).sort("metadata.write_ts")) if q!=None else None, user_queries)
    
   return {'phone_data': data_list}
 
 
+# Registering a phone as public 
+@post("/registerPublic") 
+def registerPublic():
+  user_uuid = getUUID(request)
```
shankari: `emission.storage.timeseries` and associted functions are your friends

shankari: in particular, check out `insert` and `bulk_insert`

yw374cornell:
@shankari your feedback was really helpful! Please refer to my last commit to review the changes I made.

shankari:
public data is moved to a separate server, so this change is now obsolete

PR #393, opened 2016-09-15
Add real timeline tests

### LINKED ISSUE #392: FUZZ factor causes untracked time to be missed if intake pipeline is run multiple times
In b2d7a6eead9e14af96d892b2889df5b40625c9df, we added a small fuzz factor to ensure that the previous last processed element is ignored when we query on a next run. In general, the fuzz factor is valid because we will not start a trip or section right after ending the previous one.

But in the case of untracked time, we set the start of the untracked time period to the end of the previous trip. So with the fuzz factor, we end up missing untracked time periods that span two runs of the pipeline.

More detailed analysis below.

shankari:
Add tests for the minor fixes while testing against entire timelines

Test for all commits except https://github.com/e-mission/e-mission-server/pull/390/commits/866a2e4c88fe50423fb5fcde9e42f5df44cb2f45 and https://github.com/e-mission/e-mission-server/pull/390/commits/f1f23a84f4d4a63ed61de5e526bfec690e96eb06 because I think that they came from other people's data.
    
Need to check, get permission and add the tests

Also fix #392 found while running the tests


shankari:
jenkins, test this please

PR #394, opened 2016-09-16
Fix iOS segmentation issues

shankari:
Fix issues found while checking actual timelines under iOS.
One of the fixes (a56addd) was a subtle differences caused by the iOS distance filter versus the android time filter
The other (ea73e00) was a fix that was equally applicable on both platforms, but the problem that caused it hadn't occurred on android before.

Also add tests for the cases whose files were checked in last time


shankari:
jenkins, test this please

PR #400, opened 2016-09-18
Add in a redirect URL

shankari:
This is a workaround to sending custom URL schemes via gmail or social networks
such as facebook and twitter. All such custom URL schemes are stripped out by
the networks. In response, we use a redirect link on our server that redirects
to the custom URL with the same route and parameters.


shankari:
jenkins, test this please

PR #401, opened 2016-09-18
Minor scalability fixes before re-running the pipeline

shankari:
- Handle missing config files more gracefully
- Add an index for the data.ts as well
- Minor scalability fixes
   - Configure the intake logger to rotate after 10 logs.
   - Don't run the intake pipeline on public phones
   - Pull out the tour model (and other long-running modelling tasks) into a separate pipeline


shankari:
jenkins, test this please

PR #403, opened 2016-09-19
Check to see that the requested UUIDs are actually public

shankari:
Even for the current UUID-based getPublicData API call, we want to check that
the requested UUIDs are actually public. We don't guarantee that UUIDs are private, for example, my UUID is strewn across the github repo in a bunch of tests. We are planning to share UUIDs with the bic2cal researchers, for example.

So let's add a simple check to see if the UUIDs are known to be private before returning them. This is required before I deploy https://github.com/e-mission/e-mission-server/commit/8e9f10e9fd8a470cde00a64243dbe9e385fd011c to the real server, which I plan to do in a little bit.


shankari:
@yw374cornell fyi in case you want to review this change

shankari:
I'm going to merge this as soon as tests pass though, no time to wait

PR #405, opened 2016-09-19
Several minor fixes before deploy

shankari:
- Downgrade logs
- ios smoothing issue fixes
- include all config RW documents in the valid set (unsure if this is necessary, but can't hurt)
- Notebook with analysis!


shankari:
The test pass locally.

```
$ ./e-mission-py.bash emission/tests/analysisTests/intakeTests/TestPipelineRealData.py
Ran 25 tests in 522.062s

OK
```

Jenkins is shutting down and I am not sure when it will come back up. So merging my changes so that we can move ahead with the deployment.

shankari:
Failure because I had merged before the run.

PR #406, opened 2016-09-19
Script(s) to delete all internal views (collections in mongodb) and exteral views (habitica)

shankari:
Should be run every time we completely nuke the views and recreate them.


shankari:
Failure because I had already merged since Jenkins was down.

PR #409, opened 2016-09-19
Enhance the habitica reset pipeline to reset all users

shankari:
Also reset the last timestamp since we have deleted all points so we need to
restore all points.
Also restore their points with an optional argument. This is useful when the trips/sections are correct but the habitica set is screwed up.

### DIFF
--- a/bin/ext_service/reset_all_habitica_users.py
+++ b/bin/ext_service/reset_all_habitica_users.py
@@ -4,17 +4,29 @@ import logging
 
 import emission.core.get_database as edb
 import emission.net.ext_service.habitica.proxy as proxy
+import emission.net.ext_service.habitica.sync_habitica as autocheck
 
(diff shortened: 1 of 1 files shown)


shankari:
@juemura anything else I need to fix?
I will re-run this script after the regular pipeline finishes running...

shankari:
jenkins, test this please

shankari:
jenkins, test this please

shankari:
jenkins, test this please

shankari:
On more test, since we haven't touched that code at all. If it fails again, it's a flaky test, and I need to run multiple times locally

shankari:
jenkins, test this please

shankari:
This is a weird, intermittent failure since I don't actually change anything in the `emission` directory. Merging this in order to allow me to make the scalability changes.

PR #413, opened 2016-09-20
Three scalability fixes + small fix from production

### LINKED ISSUE #408: Scalability of the e-mission server
Although we only have ~ 50 users, we have been taking more than an hour to run the intake pipeline. While we can definitely throw more hardware at the problem, it is going to be hard to scale up to 500 users on a single machine, and I am not sure I have the time to move to the cloud.

Let's see why it is taking so long to run the pipeline. We have never even profiled it before.

We can use the built-in profiler to record the traces
https://docs.python.org/2/library/profile.html

```
./e-mission-py.bash -m cProfile -o /tmp/intake.pstats bin/intake_stage.py
```

shankari:
This almost fixes #408.

After these changes, the run time for over a month (July 20 - Sep 6) for a test user has come down to roughly 5 mins.

```
2016-09-20 09:17:58,680:INFO:140735159468032:processing UUID list = [UUID('348d5c25-5f7d-4e56-94f9-457e604a2342')]
2016-09-20 09:22:29,128:INFO:140735159468032:For stage PipelineStages.OUTPUT_GEN, last_ts_processed = 2016-08-27T23:18:22.049340
```

Since this was for roughly 30 days, we can assume that the time to process one day's worth of data for one user is `(5 * 60) / 30 = 10 secs`.

So we can process data for `3600 / 10 = 360` users in one hour.

Note that this is assuming that we process all the data for a  day at a time, but actually, the pipeline runs every hour, so this is a huge underestimate. On the other hand, this does not take the "move to long term" step into account and bunching and other bolus stuff.

So I think that we can assume ~ 300 users/hour without any problems.


shankari:
Also adding a small fix from production that sneaked in here because it was committed on the same branch and then I needed to fix the tests.

shankari:
Just reran for a single day to fix the ground truth for `testAug27TooMuchExtrapolation`.
Time is 20 secs, so well within our 30 sec budget.

```
2016-09-20 12:51:19,315:INFO:140735159468032:processing UUID list = [UUID('4ba23960-c181-4795-b42b-a31f05e4f9ad')]
2016-09-20 12:51:19,556:INFO:140735159468032:**********UUID 4ba23960-c181-4795-b42b-a31f05e4f9ad: moving to long term**********
...
2016-09-20 12:51:36,351:INFO:140735159468032:For stage PipelineStages.OUTPUT_GEN, last_ts_processed is unchanged
```

PR #414, opened 2016-09-20
Make the UUID really non identifying

shankari:
While we have always had the uuid separate from the email, and used the UUID
exclusively through our code, for historical reasons, we have generated the
UUID using the uuid3 scheme from the email address. What this means is that,
while it is not possible in principle to go from the UUID -> email, if you had
the list of UUIDs, and some idea of the email of some people who were on the
list, you could re-generate the UUIDs and match them up.

Now that we are moving to our first deployment on top of e-mission, and are
going to release limited data to people who are not part of the protocol, it is
time to fix this. It does not have any practical implications any more because,
unlike when we first designed the system without an explicit mapping, we now do
have an explicit mapping in the uuid_db.

This has a couple of minor fallout issues.
- All existing UUIDs have been generated using the old scheme, and should be reset
    - EXCEPT that UUIDs for the public data from the test phones are used in a
      bunch of places - e.g.
        - to decide whether to filter
        - to decide whether to return via public interface
        - to strip out of aggregate data
      so they should NOT be reset
- The load_timeline_for_day_and_user script needs to create the email -> UUID
  mapping, since we can no longer rely on the same email to map to the same UUID.

A bonus of this change is that we can now periodically switch the email -> UUID
mapping on an ongoing basis. So if we do have issues with an email -> UUID
mapping getting leaked, we can just reset the UUID on the fly.

Woo!

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,19 +100,17 @@ You may also want to load some test data.
 
 1. Sample timeline data from the data collection eval is available in the [data-collection-eval repo](https://github.com/shankari/data-collection-eval).
 2. You can choose to load either android data `results_dec_2015/ucb.sdb.android.{1,2,3}/timeseries/*` or iOS data `results_dec_2015/ucb.sdb.ios.{1,2,3}/timeseries/*`
-3. Data is loaded using the `bin/debug/load_timeline_for_day_and_user.py`.
-  * Running it with just a timeline file loads the data with the original user - e.g.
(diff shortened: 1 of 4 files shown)


shankari:
Bonus is that now if we load new data, we don't have to log out and log in again from the UI. Also, with some minor improvements, we can load the same timeline over and over again, making it easier to compare user versus aggregate data.

PR #425, opened 2016-09-30
Two housekeeping fixes

shankari:
Before we start the heavy lifting
- Ignore all the stuff installed by bower + the conf fields
- Remove the move filter step because AFAIK all the filters have already moved!


shankari:
jenkins test this please

PR #426, opened 2016-09-30
Implement framework for autochecking tasks

shankari:
Note that the TestHabiticaAutocheck test has NOT been updated properly.
It will FAIL as is.

This sets a framework for defining a small language for autochecked tasks.

The framework will load a mapping function based on the language, and call it
with the user_id, the task (that contains task arguments) and the previous
state.

It returns the new state, which is then persisted back to the database.

Rework the existing walk and bike task into this framework
and change instances that called it to call the new code instead.


shankari:
@juemura @sunil07t here it is!

juemura:
steps I followed to test on emulator:
1. Create new emission/habitica user.
2. From Python Notebook, find the new user's habitica_id, and then retrieve the autocheck task id by running `db.tasks.find({'userId': *habitica_id*}).pretty()` on mongo shell (remote server)
3. On Python Notebook, reset the task timestamp:
`
import emission.net.ext_service.habitica.auto_tasks.task as enehat
import emission.net.ext_service.habitica.executor as autocheck
from uuid import UUID
import arrow

dummy_task = enehat.Task()
dummy_task.task_id = *_id*
modification = {"last_timestamp": arrow.Arrow(2015,5,1).timestamp, "bike_count": 0, "walk_count":0}
autocheck.save_task_state(UUID(*uuid*), dummy_task, modification)
list(edb.get_habitica_db().find({'user_id': UUID(*uuid*)}))
`
4. Load data for new user:
`$ ./e-mission-py.bash bin/debug/load_timeline_for_day_and_user.py *data_file_path* *user_email*
5. Run pipeline for that user:
`$ ./e-mission-py.bash bin/debug/intake_single_user.py -e *user_email*`

PR #427, opened 2016-09-30
level challenge 2 middle tasks auto check

sunil07t:


shankari:
ok to test

emission/net/ext_service/habitica/auto_tasks/level_challenge.py
```diff
@@ -0,0 +1,210 @@
...
+	    for item in summary_ts["result"]
+	    	try:
+	            bike_trips += item.BICYCLING
+	            logging.debug("bike_trips += %s" % item.BICYCLING)
+	        except AttributeError:
+	            logging.debug("no bike")
+
+	    logging.debug("Finished with bike_trips == %s" % bike_trips)
+
+	    method_uri_level_challenge = "/api/v3/tasks/"+ task.task_id + "/score/up"
```
shankari: How does the user know what they are supposed to be doing now for the challenge? Are you going to edit the task description? If so, where are the calls for that?

emission/net/ext_service/habitica/auto_tasks/level_challenge.py
```diff
@@ -0,0 +1,210 @@
...
+
+	    logging.debug("Returning %s" % new_state)
+	    return new_state
+
+	# Level 4 - 
+	if len(task.habitica_task.history) >= distance_to_cover + trips_to_cover + level_1_repetition: ######and complete == False:
+		
+		if len(task.habitica_task.history) == 111 :
+			done_for_day = False
+			ts_from_level_3 = curr_state["last_timestamp"]
```
shankari: I'm not convince this is right, but I'm going to think about it tonight. In particular, I think that this assumes that the pipeline is running continuously. What if it does down for a bit? What if I want to re-run after the gap of a few days? Would be nice to make it easier to support that.

PR #431, opened 2016-10-05
Cleanup changes made directly on production

shankari:
- print the elements deleted in each database
- skip the user with mixed ios and android streams
- add more public phones to the list
- delete obsolete create party leaders script


shankari:
jenkins, test this please

PR #450, opened 2016-11-24
created script to find inactive users

juemura:
This script prints a list of inactive users (users who do not have a usercache call within the last week). 
To do: format output. (Please let me know what format you want)

### DIFF
--- a/bin/find_inactive_users.py
+++ b/bin/find_inactive_users.py
@@ -2,45 +2,67 @@ import logging
 import emission.core.get_database as edb
 import emission.storage.timeseries.abstract_timeseries as esta
 import arrow
+import pandas as pd
 
 
 def find_inactive_users():
     inactive_users = []
     inactive_users_new_consent = ""
     inactive_users_old_consent = ""
+    inactive_users_before_september = ""
+    inactive_users_after_september = ""
     one_week_ago_ts = arrow.utcnow().replace(weeks=-1).timestamp
+    september_first = arrow.get('2016-09-01').timestamp
     for user in edb.get_uuid_db().find():
         db = esta.TimeSeries.get_time_series(user['uuid']).get_data_df("stats/server_api_time", time_query=None)
         new_consent = esta.TimeSeries.get_time_series(user['uuid']).get_data_df("config/consent", time_query=None)
+        signup_date = arrow.get(user['update_ts'])
         if db.empty:
-            inactive_users.append((user['user_email'], user['update_ts']))                                  
+            inactive_users.append((user['user_email'], signup_date.date(), ()))
             if new_consent.empty:
                 inactive_users_new_consent+=str(user['user_email'])+', '
             else:
                 inactive_users_old_consent+=str(user['user_email'])+', '
+            if signup_date.timestamp < september_first:
+                inactive_users_before_september+=str(user['user_email'])+', '
+            else:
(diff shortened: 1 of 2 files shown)


bin/find_inactive_users.py
```diff
@@ -0,0 +1,28 @@
...
+    one_week_ago_ts = arrow.utcnow().replace(weeks=-1).timestamp
+    for user in edb.get_uuid_db().find():
+        db = esta.TimeSeries.get_time_series(user['uuid']).get_data_df("stats/server_api_time", time_query=None)
+        if db.empty:
+            inactive_users.append(user)
+        else:
+            #check last usercache call: 
+            #the user is inactive if there are no calls or if the last one was before one_week_ago_ts
+            last_usercache_call = db[db['name'].str.contains('usercache', case=False)].tail(1)
+            if last_usercache_call.empty:
```
shankari: is empty a field or a function?
i.e. does it need to be `empty()`?

bin/find_inactive_users.py
```diff
@@ -0,0 +1,28 @@
...
+                inactive_users.append(user)
+            elif last_usercache_call.iloc[0]['ts'] < one_week_ago_ts:
```
shankari: can't you use `head(1)` instead of `iloc[0]`?

shankari:
ok to test

bin/find_inactive_users.py
juemura: empty is an attribute of a pandas dataframe, so no need for parenthesis

bin/find_inactive_users.py
juemura: head(1) doesn't work...

juemura:
If the only use is to send emails, I list a comma separated list would be best.

shankari:
comma separated list of _emails_ though, not uuids (can't send email to uuid, no?)

shankari:
are you suggesting that there are other things we should use this for?

juemura:
I'm just saying that if right now we just need the emails, then a comma separated list of emails would be best because you can just copy and paste into bcc. But if you have something else in mind, then that's where I was asking if you had a preference for format.

shankari:
I feel like we are repeating ourselves.
The main goal of the script was to send emails to inactive users.

It would also be nice to have a graph of active users over time so that we can answer the inevitable question on how many users we actually have.
Note that you need this graph for the paper as well.

juemura:
Right. I think I'll write a different script to generate the graphs for the paper. For now, I turned the list into a string to avoid the ' ' around each email. Let me know if this works for you.

shankari:
Do we want to send emails to everybody, or only to people who installed the app since the summer? There were people who had installed the app before summer, including going back to last year.
You might want to filter by people who have consented to the most recent IRB protocol (that have `config/consent` in the timeseries)

You may want to generate a table of emails and when people signed up - the followup email for your friends who installed over the summer might need to go from you instead of from me.

shankari:
to clarify, I suggest that maybe you should have:
- a table of email + date
- one set of emails for people before "end of summer" (that you can send out)
- one set of emails for people this semester (that I will send out)

shankari:
nit: in your next change, move this into a subdirectory of bin
(`bin/dashboard` or `bin/stats` or something)

PR #453, opened 2016-11-30
Add support for displaying the aggregate stress map

shankari:
Similar to
https://github.com/e-mission/e-mission-phone/commit/9f6c1fb804b20d05ce9fd6f15fd612b3264d0799
add support for displaying the stress heatmap in addition to the count heatmap

This is pretty much a direct port of the phone code. But because the server
code is so much behind the phone code, we needed to:
- copy over a bunch of styles
- add a standard logger and the `post-trip-manual` class
- add switching and stress code to the heatmap with minor variations
- update the HTML to include the two tabs + ensure that the map handles clicks
- add everything to index.html

### DIFF
--- a/webapp/www/css/style.css
+++ b/webapp/www/css/style.css
@@ -91,3 +91,70 @@ section {
     background-position: top left, center center;
 }
 
+
+.hvcenter {
+  display: flex;           /* establish flex container */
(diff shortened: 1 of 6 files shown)


shankari:
All changes are to the UI alone, so testing is really redundant.
But might as well wait for it to finish running.

PR #460, opened 2017-01-19
Only mark an error if the difference is significant enough

shankari:
The values are very similar most of the time, but sometimes, the process
appears to be I/O bound and there is a big gap. Let us mark errors only in that
case to avoid error spew.

We could also stop storing 2 values unless there is an error...


shankari:
jenkins, test this please

PR #468, opened 2017-02-25
initial script for app analytics

juemura:
temporary fix for outliers was to replace any reading >1 with 1. I'll work on that more.
I'm still not happy with the graphs, but I want to see how they look with real data. I'll work more on the style and try to find a better option.


shankari:
I am happy to deploy this and return the results, but I think that it doesn't do what you want. Right now, it will only generate results for the last user. I don't see where you are getting and saving per-user information...

bin/analysis/get_app_analytics.py
```diff
@@ -0,0 +1,120 @@
...
+import datetime as dt
+import matplotlib.dates as mdates
+
+
+def get_app_analytics():
+    df = pd.DataFrame()
+    for user in edb.get_uuid_db().find():
+        user_df = esta.TimeSeries.get_time_series(user['uuid']).get_data_df("stats/server_api_time", time_query=None)
+        if not user_df.empty:
+            df = df.append(user_df, ignore_index = True)
```
shankari: At this point, you have one giant dataframe with data combined across all users, right?

e.g. if you have users u1 and u2, df will consist of 

<u1 row1>
<u1 row2>
<u1 row3>
<u1 row4>
....
<u2 row1>
<u2 row2>
....

so if you plot the calls to metrics/timestamp, you won't see how user1 accessed the dashboard versus user2 accessed the dashboard. Instead, you will get a combined plot across both users, right?

bin/analysis/get_app_analytics.py
```diff
@@ -0,0 +1,120 @@
...
+    df.ix[df.reading>1, 'reading'] = 1
+    fig, ax = plt.subplots()
+    ax.xaxis.set_major_locator(mdates.WeekdayLocator())
+    ax.xaxis.set_major_formatter(mdates.DateFormatter("%m/%d/%Y"))
+    plt.ylabel('Response time')
+
+    dashboard_df = df[df.name == "POST_/result/metrics/timestamp"]
+    dashboard_df.plot(x="datetime", y="reading", ax=ax, style='+', legend=None)
+    plt.title('Dashboard')
+    fig.savefig('Dashboard.png')
```
shankari: in a subsequent change, can you please save the figures into a directory (like `app_analytics_results`) so that it is easier to zip them all up and share with you?

PR #469, opened 2017-02-27
Changes after deployment

shankari:
- Final fixes to the external pushes (skip API call if no users, filter out users with no token, use dev or production as appropriate)
- Add the dev command line parameter to last remaining script
- Switch to the new version of the google play library


shankari:
jenkins is down.
screen shot 2017-02-26 at 10 42 42 pm

PR #470, opened 2017-03-02
Fix import logging

aminebouzaghrane:
Changing the import to fix the error to be able to run the pipeline.


shankari:
ok to test

PR #486, opened 2017-04-17
Add the css for the redirect page

shankari:
Missing file!


shankari:
@juemura missing file!

PR #493, opened 2017-06-27
Model stage

jensteff:
I've been having trouble figuring out the exact syntax to run the modeling stage. I have been basing it off of the intake stage, and have been looking into mode.py and trying to map how to run all the functions. model_stage.py is where my focus is presently.

### DIFF
--- a/emission/analysis/classification/inference/mode.py
+++ b/emission/analysis/classification/inference/mode.py
@@ -28,8 +28,10 @@ class ModeInferencePipeline:
                           "start hour", "end hour", "close to bus stop", "close to train stop",
                           "close to airport"]
     self.Sections = edb.get_section_db()
+    self.PredictionDB = edb.get_prediction_db()
 
-  def runPipeline(self):
+  def runPipelineModelStage(self):
+
     allConfirmedTripsQuery = ModeInferencePipeline.getSectionQueryWithGroundTruth({'$ne': ''})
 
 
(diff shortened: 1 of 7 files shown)


emission/analysis/classification/inference/mode.py
```diff
@@ -28,8 +28,10 @@ def __init__(self):
                           "start hour", "end hour", "close to bus stop", "close to train stop",
                           "close to airport"]
     self.Sections = edb.get_section_db()
+    self.PredictionDB = edb.get_prediction_db()
+
+  def runPipelineModelStage(self):
```
shankari: this seems similar to the changes in #494. Are there any differences? otherwise, my comments still apply.

emission/core/wrapper/modeprediction.py
```diff
@@ -0,0 +1,52 @@
...
+	'''
+	NOT_PREDICTED = 0
+	RANDOM_FOREST = 1
+
+
+class Modeprediction(ecwb.WrapperBase):
+	props = {"trip_id": 	ecwb.WrapperBase.Access.WORM, 	# *******the trip that this is part of
+            "section_id": 	ecwb.WrapperBase.Access.WORM, 	#The section id that this prediction corresponds to
+            "algorithm_id": ecwb.WrapperBase.Access.WORM, 	#The algorithm which made this prediction
+            "confirmed_mode": ecwb.WrapperBase.Access.WORM, #The mode inputted by the user (if it exists)
```
shankari: interesting choice. I am not sure that the `confirmed_mode` should be here. Think about it - let's say that there is a trip section that represents your walk from ihouse to cory. We run two algorithms on it (one based on general model, and one based on personal model), and they return `walking` and `running` respectively. I think that our assumption was that this would lead to two prediction objects - one with `predicted_mode: walking` and the other with `predicted_mode: running`.

But the user wouldn't confirm twice, only once, since there is only one underlying section. Or am I missing something here?

emission/core/wrapper/modeprediction.py
```diff
@@ -0,0 +1,52 @@
...
+            "sensed_mode": 	ecwb.WrapperBase.Access.WORM, 	#THe mode that the phones sensors picked up
+            "predicted_mode": ecwb.WrapperBase.Access.WORM, #What we predicted
+            "confidence":	ecwb.WrapperBase.Access.WORM, 	#Our confidence in the prediction 
+
+
+             # "start_ts": ecwb.WrapperBase.Access.WORM, # ****start UTC timestamp (in secs)
```
shankari: remove commented out code before final merge

emission/core/wrapper/modeprediction.py
```diff
@@ -0,0 +1,52 @@
...
+             # "end_stop": ecwb.WrapperBase.Access.WORM,    # _id of place object after this one
+             # "start_loc": ecwb.WrapperBase.Access.WORM,    # location of start point in geojson format
+             # "end_loc": ecwb.WrapperBase.Access.WORM,      # location of end point in geojson format
+             # "duration": ecwb.WrapperBase.Access.WORM,     # duration of the trip in secs
+             # "sensed_mode": ecwb.WrapperBase.Access.WORM,  # ****the sensed mode used for the segmentation
+             # "source": ecwb.WrapperBase.Access.WORM}        # *****the method used to generate this trip
+             }
+    enums = {"sensed_mode": ecwm.MotionTypes,
+    		"predicted_mode": ecwm.MotionTypes,
+    		"confirmed_mode": ecwm.MotionTypes,
```
shankari: I am not at all convinced that the predicted and confirmed modes should be `ecwm.MotionTypes`. In particular, note that `ecwm.MotionTypes` does not have entries for bus, car or train, let alone more exotic modes.

emission/core/wrapper/modeprediction.py
```diff
@@ -0,0 +1,52 @@
...
+    		"algorithm_id": AlgorithmTypes }
+    geojson = ["start_loc", "end_loc"]
+    nullable = ["start_stop", "end_stop", "confirmed_mode", "predicted_mode"]
+    local_dates = ['start_local_dt', 'end_local_dt']
```
shankari: these are not relevant and need to be fixed.

emission/pipeline/model_stage.py
```diff
@@ -4,13 +4,70 @@
...
+			print "dang flabbit failed on error %s" % e
+
+
+def run_mode_inference_pipeline_for_user(uuid):
+
+	MIP = ModeInferencePipeline() #I don't think its this simple
+	
+
+	allConfirmedTripsQuery = ModeInferencePipeline.getSectionQueryWithGroundTruth({'$ne': ''})
+	#These are all the trips that have a confirmed mode. We will be training off of this.
```
shankari: don't query on confirmed mode, query on time range

emission/storage/pipeline_queries.py
```diff
@@ -113,6 +114,19 @@ def mark_clean_resampling_done(user_id, last_section_done):
 def mark_clean_resampling_failed(user_id):
     mark_stage_failed(user_id, ps.PipelineStages.CLEAN_RESAMPLING)
 
+def get_time_range_for_mode_inference(user_id):
```
shankari: you need to call this somewhere

emission/tests/analysisTests/modeinferTests/TestPipeline.py
```diff
@@ -7,7 +7,7 @@
 
 # Our imports
 from emission.core.get_database import get_db, get_mode_db, get_section_db
-import emission.analysis.classification.inference.mode as pipeline
```
shankari: Same as #494?

emission/core/wrapper/modeprediction.py
jensteff: Yes, that is a good point. I think we should put confirmed mode somewhere else.

emission/core/wrapper/modeprediction.py
jensteff: So, should I create a new enum for it?

shankari: For now, I would make it a string. We should probably store the list of modes in a database somewhere - let's design that when we design the mode editing.

emission/core/wrapper/modeprediction.py
shankari: Yeah I think it should be a separate wrapper object. But then maybe it should not just be mode, but confirmed everything. When you confirm a trip, you are confirming start/end time, location, etc as well.

Just leave confirmed_mode out for now or leave it in section. This needs some design.

PR #494, opened 2017-06-27
Pipeline testing

jensteff:
This is mostly finished and is working code. It will change slightly in the future once we implement the model_stage and the things that go with that.

### DIFF
--- a/emission/analysis/classification/inference/mode.py
+++ b/emission/analysis/classification/inference/mode.py
@@ -8,10 +8,17 @@ import numpy as np
 import scipy as sp
 import time
 from datetime import datetime
+import emission.storage.timeseries.abstract_timeseries as esta
+import emission.storage.decorations.analysis_timeseries_queries as esda
+import emission.core.wrapper.entry as ecwe
+import emission.storage.decorations.trip_queries as esdt
 
 # Our imports
 import emission.analysis.section_features as easf
 import emission.core.get_database as edb
+
+
+from uuid import UUID
 
 # We are not going to use the feature matrix for analysis unless we have at
 # least 50 points in the training set. 50 is arbitrary. We could also consider
@@ -28,13 +35,10 @@ class ModeInferencePipeline:
                           "start hour", "end hour", "close to bus stop", "close to train stop",
                           "close to airport"]
     self.Sections = edb.get_section_db()
-    self.PredictionDB = edb.get_prediction_db()
 
-  def runPipelineModelStage(self):
+  def runPipelineModelStage(self, uuid, timerange):
 
     allConfirmedTripsQuery = ModeInferencePipeline.getSectionQueryWithGroundTruth({'$ne': ''})
-
-
 
     (self.modeList, self.confirmedSections) = self.loadTrainingDataStep(allConfirmedTripsQuery)
     logging.debug("confirmedSections.count() = %s" % (self.confirmedSections.count()))
@@ -57,28 +61,14 @@ class ModeInferencePipeline:
     self.selFeatureMatrix = self.cleanedFeatureMatrix[:,self.selFeatureIndices]
     self.model = self.buildModelStep()
     logging.info("buildModelStep DONE")
-    # toPredictTripsQuery = {"$and": [{'type': 'move'},
-    #                                ModeInferencePipeline.getModeQuery(''),
-     #                               {'predicted_mode': None}]]}
+
     Algorithm_to_be_used_to_predict = 0
 
-    #If we want to predict the things who haven't been predicted with the current prediction method, we can
-    #   set that here. If it is zero, that means predict everything that doesn't have a prediction already in the database
-    if Algorithm_to_be_used_to_predict:
-      IDsWithThisAlgorithm = PredictionDB.find().distinct("section_id", { 'algorithm_id': Algorithm_to_be_used_to_predict})
-      toPredictIDsQuery = {'_id': {'$nin' : IDsWithThisAlgorithm}}
 
-    else:
-      AllIDsInPredictionDB = PredictionDB.find().distinct("section_id")
-      toPredictIDsQuery = {'_id' : {'$nin': AllIDsInPredictionDB}}
+    ts = esta.TimeSeries.get_time_series(user_id)
+    toPredictTrips_it = ts.find_entries(['analysis/cleaned_section'], time_query=timerange)
 
-    #toPredictTripsQuery = {"$and": [ModeInferencePipeline.getModeQuery(''), {'$or': ['predicted_mode': None}]}
-    
-    #This returns all the section ids that are not currently in the predictiondb
-    #We're going to need to either query the section database to see which section does not have a prediction in the
-    #   prediction database, or we need to grab all the ids of the predictions whose algorithm_id is zero.
-
-    (self.toPredictFeatureMatrix, self.sectionIds, self.sectionUserIds) = self.generateFeatureMatrixAndIDsStep(toPredictTripsQuery)
+    (self.toPredictFeatureMatrix, self.sectionIds, self.sectionUserIds) = self.generateFeatureMatrixAndIDsStep(toPredictTrips_it)
     logging.info("generateFeatureMatrixAndIDsStep DONE")
     self.predictedProb = self.predictModesStep()
     #This is a matrix of the entries and their corresponding probabilities for each classification
@@ -147,13 +137,11 @@ class ModeInferencePipeline:
   @staticmethod
   def getSectionQueryWithGroundTruth(groundTruthMode):
     # return {"$and": [{'type': 'move'},
-    return {"$and": [{"$or": [{'type': { '$exists' : False }}, {'type' : 'move'}]},        
-                     ModeInferencePipeline.getModeQuery(groundTruthMode)]}
+    return {ModeInferencePipeline.getModeQuery(groundTruthMode)]}
 
   @staticmethod
   def getSectionQueryWithGroundTruthUser(groundTruthMode, uuid):
-     return {"$and": [{"$or": [{'type': { '$exists' : False }}, {'type' : 'move'}]},        
-                     ModeInferencePipeline.getModeQuery(groundTruthMode)]}
+     return {ModeInferencePipeline.getModeQuery(groundTruthMode)}
 
   # TODO: Refactor into generic steps and results
   def loadTrainingDataStep(self, sectionQuery, sectionDb = None):
@@ -226,25 +214,28 @@ class ModeInferencePipeline:
 
       # So we limit the records to the size of the matrix that we have created
       for (i, section) in enumerate(self.confirmedSections.limit(featureMatrix.shape[0]).batch_size(300)):
-       try:
-        self.updateFeatureMatrixRowWithSection(featureMatrix, i, section)
-        resultVector[i] = self.getGroundTruthMode(section)
+        try:
+          self.updateFeatureMatrixRowWithSection(featureMatrix, i, section)
+          resultVector[i] = self.getGroundTruthMode(section)
     
 
-        if i % 100 == 0:
+          if i % 100 == 0:
             logging.debug("Processing record %s " % i)
-      except Exception, e:
-        logging.debug("skipping section %s due to error %s " % (section, e))
+        except Exception, e:
+          logging.debug("skipping section %s due to error %s " % (section, e))
+     
     
       return (featureMatrix, resultVector)
 
   def getGroundTruthMode(self, section):
       # logging.debug("getting ground truth for section %s" % section)
       if 'corrected_mode' in section:
-          # logging.debug("Returning corrected mode %s" % section['corrected_mode'])
+         
+          logging.debug("Returning corrected mode %s" % section['corrected_mode'])
           return section['corrected_mode']
       else:
-          # logging.debug("Returning confirmed mode %s" % section['confirmed_mode'])
+     
+          logging.debug("Returning confirmed mode %s" % section['confirmed_mode'])
           return section['confirmed_mode']
 
 # Features are:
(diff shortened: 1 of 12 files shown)


emission/analysis/section_features.py
```diff
@@ -15,18 +15,40 @@
...
 
 # The speed is in m/s
-def calSpeed(trackpoint1, trackpoint2):
+def calSpeed(section):
   from dateutil import parser
-  distanceDelta = calDistance(trackpoint1['track_location']['coordinates'],
-                              trackpoint2['track_location']['coordinates'])
-  timeDelta = parser.parse(trackpoint2['time']) - parser.parse(trackpoint1['time'])
+
+  if ((not section['start_loc']) or (not section['end_loc'])):
```
shankari: this is an incorrect port. the previous implementation computed the speed between two points along the trip (e.g. two `emission/core/wrapper/location.py` objects or their subclasses. the new implementation computes the speed between the start and end points, which is not equal to the speed along the trajectory if the trip is not in a straight line, and it rarely is.

shankari: Also this appears to be the speed for the entire section instead of between every pair of points along the trajectory.

emission/analysis/section_features.py
```diff
@@ -15,18 +15,40 @@
...
+#   from dateutil import parser
+#   distanceDelta = calDistance(trackpoint1['track_location']['coordinates'],
+#                               trackpoint2['track_location']['coordinates'])
+#   timeDelta = parser.parse(trackpoint2['time']) - parser.parse(trackpoint1['time'])
+#   # logging.debug("while calculating speed form %s -> %s, distanceDelta = %s, timeDelta = %s" %
+#   #               (trackpoint1, trackpoint2, distanceDelta, timeDelta))
+#   if timeDelta.total_seconds() != 0:
+#     return distanceDelta / timeDelta.total_seconds()
+#   else:
+#     return None
```
shankari: nit: remove commented out code before final merge

emission/analysis/section_features.py
```diff
@@ -154,13 +176,37 @@ def calAvgSpeed(segment):
 # segment0: speed0 / (t1 - t0), segment1: (speed1 - speed0)/(t2-t1),
 # segment2: (speed2 - speed1) / (t3-t2)
 
+# def calAccels(segment):
+#   from dateutil import parser
+
+#   speeds = calSpeeds(segment)
```
shankari: nit: remove commented out code before final merge

emission/analysis/section_features.py
```diff
@@ -154,13 +176,37 @@ def calAvgSpeed(segment):
...
+#       # logging.debug("resulting acceleration is %s" % accel[i])
+#     prevSpeed = currSpeed
+#   return accel
+
 def calAccels(segment):
   from dateutil import parser
 
-  speeds = calSpeeds(segment)
-  trackpoints = segment['track_points']
+  speed = calSpeed(segment)
```
shankari: you can't replace `calSpeeds` by `calSpeed` because `calSpeeds` computes the speeds between each pair of points along the section. Note that this is already present in the new cleaned section object as `speeds`

emission/tests/analysisTests/modeinferTests/TestPipelineNew.py
```diff
@@ -0,0 +1,342 @@
...
+    etc.dropAllCollections(get_db())
+
+    self.ModesColl = get_mode_db()
+    self.assertEquals(self.ModesColl.find().count(), 0)
+
+    self.SectionsColl = get_section_db()
+    self.assertEquals(self.SectionsColl.find().count(), 0)
+
+    etc.loadTable(self.serverName, "Stage_Modes", "emission/tests/data/modes.json")
+    etc.loadTable(self.serverName, "Stage_Sections", "emission/tests/data/newTestModeInferFilePlacesDeleted")
```
shankari: in the new model, data is not stored in `Stage_Sections` but instead, in the timeseries

emission/tests/analysisTests/modeinferTests/TestPipeline.py
```diff
@@ -131,8 +131,14 @@ def testCleanDataStep(self):
     (self.pipeline.modeList, self.pipeline.confirmedSections) = self.pipeline.loadTrainingDataStep(allConfirmedTripsQuery)
     self.testGenerateBusAndTrainStops()
     (self.pipeline.featureMatrix, self.pipeline.resultVector) = self.pipeline.generateFeatureMatrixAndResultVectorStep()
+    print "ConfirmedSections"
+    print self.pipeline.confirmedSections.distinct('_id')
+    print self.pipeline.confirmedSections.count()
 
     (self.pipeline.cleanedFeatureMatrix, self.pipeline.cleanedResultVector) = self.pipeline.cleanDataStep()
+
+    print self.pipeline.cleanedFeatureMatrix.shape
```
shankari: nit: remove print statements or convert to log before final merge

emission/tests/analysisTests/modeinferTests/TestPipelineNew.py
```diff
@@ -0,0 +1,342 @@
...
+from emission.core.wrapper.user import User
+from emission.core.wrapper.client import Client
+import emission.tests.common as etc
+'''
+TODO:
+    Create some outliers and make sure they're stripped out
+
+'''
+class TestPipeline(unittest.TestCase):
+  def setUp(self):
```
shankari: the problem with copying the file over is that I can't see only the changes. can you highlight the changes from old -> new by surrounding with something like `### porting change ###` to make it easier to review?

emission/tests/data/newTestModeInferFilePlacesDeleted
```diff
@@ -0,0 +1,475 @@
...
+  "minute": 5,
+  "second": 49,
+  "weekday": 1,
+  "year": 2014, 
+  "month": 4,
+  "day": 1,
+  "timezone": "America/Los_Angeles"
+},
+"end_ts": 1396371949,    
+"section_id": 0,
```
shankari: no `section_id` in new section wrapper

emission/analysis/classification/inference/mode.py
```diff
@@ -48,15 +50,80 @@ def runPipeline(self):
...
+    logging.info("generateFeatureMatrixAndIDsStep DONE")
+    self.predictedProb = self.predictModesStep()
+    #This is a matrix of the entries and their corresponding probabilities for each classification
+
+    logging.info("predictModesStep DONE")
+    self.savePredictionsStep()
+    logging.info("savePredictionsStep DONE")
+
+
+  def runPipeline(self):
```
jensteff: I created `runPipelineModelStage` because running the `runPipeline` in `model_stage` wouldn't have done what we needed, so I made a new version of the same function where I could play with the function without deleting the version used in `testPipeline`.

emission/analysis/classification/inference/mode.py
```diff
@@ -79,29 +146,39 @@ def getModeQuery(groundTruthMode):
 
   @staticmethod
   def getSectionQueryWithGroundTruth(groundTruthMode):
-    return {"$and": [{'type': 'move'},
+    # return {"$and": [{'type': 'move'},
+    return {"$and": [{"$or": [{'type': { '$exists' : False }}, {'type' : 'move'}]},
```
jensteff: Yes, I had originally kept this in because I was running it both with the old datafile and the new one. Later on I stop doing this, so I take it out of other queries, but I left some of them in with the intention to clean them up later.

emission/analysis/classification/inference/mode.py
```diff
@@ -79,29 +146,39 @@ def getModeQuery(groundTruthMode):
...
   @staticmethod
   def getSectionQueryWithGroundTruth(groundTruthMode):
-    return {"$and": [{'type': 'move'},
+    # return {"$and": [{'type': 'move'},
+    return {"$and": [{"$or": [{'type': { '$exists' : False }}, {'type' : 'move'}]},        
+                     ModeInferencePipeline.getModeQuery(groundTruthMode)]}
+
+  @staticmethod
+  def getSectionQueryWithGroundTruthUser(groundTruthMode, uuid):
+     return {"$and": [{"$or": [{'type': { '$exists' : False }}, {'type' : 'move'}]},
```
jensteff: Yes, I didn't create this function, I just changed the part about the type being move. I don't know how it relates to a specific user either, so I will make it so.

emission/analysis/classification/inference/mode.py
```diff
@@ -79,29 +146,39 @@ def getModeQuery(groundTruthMode):
...
+    logging.debug("Section data set size = %s" % sectionDb.find({'$or':[{'type':{'$exists': False}}, {'type': 'move'}]}).count())
+    #AAAlogging.debug("Section data set size = %s" % sectionDb.find({'type': 'move'}).count())
+
     duration = time.time() - begin
     logging.debug("Getting dataset size took %s" % (duration))
         
     logging.debug("Querying confirmedSections %s" % (datetime.now()))
     begin = time.time()
-    confirmedSections = sectionDb.find(sectionQuery)
+    confirmedSections = sectionDb.find(sectionQuery).sort('_id', 1)
```
jensteff: I don't know either. After you merged it, it got deleted from my code, so I had to put it back in. I'm finding the same thing this morning with some of my code being reverted to how it was when I first pulled it.

emission/analysis/classification/inference/mode.py
```diff
@@ -149,13 +226,16 @@ def generateFeatureMatrixAndResultVectorStep(self):
...
-            logging.debug("skipping section %s due to error %s " % (section, e))
+       try:
+        self.updateFeatureMatrixRowWithSection(featureMatrix, i, section)
+        resultVector[i] = self.getGroundTruthMode(section)
+    
+
+        if i % 100 == 0:
+            logging.debug("Processing record %s " % i)
+      except Exception, e:
+        logging.debug("skipping section %s due to error %s " % (section, e))
```
jensteff: There isn't a difference. I had some print statements around there and commented it some of it out and then put it back in. But it is the same code.

emission/analysis/classification/inference/mode.py
```diff
@@ -191,47 +271,53 @@ def getGroundTruthMode(self, section):
 # 20. both start and end close to train station
 # 21. both start and end close to airport
   def updateFeatureMatrixRowWithSection(self, featureMatrix, i, section):
-    featureMatrix[i, 0] = section['distance']
-    featureMatrix[i, 1] = (section['section_end_datetime'] - section['section_start_datetime']).total_seconds()
+    if 'distance' in section.keys():
+      featureMatrix[i,0] = section['distance']
+    else:
+      featureMatrix[i, 0] = easf.calDistance(section['start_loc']['coordinates'], section['end_loc']['coordinates'])  #####would need to calculate this
```
jensteff: None of the sections have a distance, besides the outlier that we add in (which is why I say in the code comments I need to make a distance outlier using the new model instead of `distance`). In the new model, the only sort of thing related to it that we have is a start location and an end location, so the only way to get any sort of distance is just to get the linear distance between them.

emission/analysis/classification/inference/mode.py
```diff
@@ -191,47 +271,53 @@ def getGroundTruthMode(self, section):
...
+      featureMatrix[i, 2] = section['sensed_mode']
     except ValueError:
       featureMatrix[i, 2] = 0
 
-    featureMatrix[i, 3] = section['section_id']
-    featureMatrix[i, 4] = easf.calAvgSpeed(section)
-    speeds = easf.calSpeeds(section)
-    if speeds != None and len(speeds) > 0:
+    featureMatrix[i, 3] = section['section_id']   ####
+    #featureMatrix[i, 4] = easf.calAvgSpeed(section)
```
jensteff: It isn't. I commented it out to see how it was working and simply forgot to put it back in. Since the entry should be zero anyway, it wasn't obvious I forgot that bit.

emission/analysis/classification/inference/mode.py
```diff
@@ -298,8 +387,17 @@ def buildModelStep(self):
     model = forestClf.fit(self.selFeatureMatrix, self.cleanedResultVector)
     return model
 
+  def buildSpecificModel(self, algorithm_id):
```
jensteff: It isn't called yet. I added it because I knew we were eventually going to use different models to predict and so I wanted to have it there so I could incorporate it into the code now (but only one model would be supported at the moment) so it would be easier to adjust later.

emission/tests/analysisTests/modeinferTests/TestPipelineNew.py
jensteff: yes, I will start doing that.

emission/analysis/section_features.py
```diff
@@ -176,13 +238,43 @@ def calAvgSpeed(segment):
...
+
+  accel = np.zeros(len(speeds) - 1)
+  prevSpeed = 0
+  for (i, speed) in enumerate(speeds[0:-1]):
+    currSpeed = speed # speed0
+    speedDelta = currSpeed - prevSpeed # (speed0 - 0)
+    accel[i] = speedDelta
+    # t1 - t0
+
+    timeDelta = entries[i+1]['data']['ts'] - entries[i]['data']['ts']
```
shankari: seems reasonable, but my brain is not working and doesn't seem to be identical to original code.
Are you sure that the acceleration you are computing is in the correct direction?

emission/tests/analysisTests/modeinferTests/TestPipeline.py
```diff
@@ -29,7 +35,7 @@ def setUp(self):
     self.assertEquals(self.SectionsColl.find().count(), 0)
 
     etc.loadTable(self.serverName, "Stage_Modes", "emission/tests/data/modes.json")
-    etc.loadTable(self.serverName, "Stage_Sections", "emission/tests/data/testModeInferFile")
+    etc.loadTable(self.serverName, "Stage_Sections", "emission/tests/data/newTestModeInferFilePlacesDeleted")
```
shankari: I still think this won't work. You don't want to load the new-style data into `Stage_Sections`

emission/tests/analysisTests/modeinferTests/TestPipeline.py
```diff
@@ -42,13 +48,11 @@ def setUp(self):
     for section in self.SectionsColl.find():
       section['section_start_datetime'] = self.dayago
       section['section_end_datetime'] = self.dayago + timedelta(hours = 1)
-      if (section['confirmed_mode'] == 5):
+      if (('confirmed_mode' in section and section['confirmed_mode'] == 5) or ('sensed_mode' in section and section['sensed_mode'] == 5)):
```
shankari: `sensed_mode == 5` doesn't mean the same thing any more.
from `emission/core/wrapper/motionactivity.py`, `TILTING = 5`

emission/pipeline/model_stage.py
```diff
@@ -11,7 +11,27 @@
...
+			logging.debug("Skipping user %s because of error" % user_id)
+			logging.debug("error %s" % e)
+			mark_mode_failed_for_user(user_id)
+
+def run_mode_pipeline_for_user(MIP, uuid):
+
+	timerange = get_time_range_for_mode_inference(uuid)
+	MIP.runPipelineModelStage(uuid, timerange)
+	mark_mode_done_for_user(uuid, timerange) # I know this timerange should be something else, but I'm not sure what.
+	#											maybe the most recent end ts grabbed? In that case, I may need to return that from runPipelineModelStage
```
shankari: yeah most recent end ts grabbed, and you should probably return it.
I think that's what I do elsewhere.

emission/analysis/classification/inference/mode.py
```diff
@@ -28,14 +35,11 @@ def __init__(self):
                           "start hour", "end hour", "close to bus stop", "close to train stop",
                           "close to airport"]
     self.Sections = edb.get_section_db()
-    self.PredictionDB = edb.get_prediction_db()
 
-  def runPipelineModelStage(self):
+  def runPipelineModelStage(self, uuid, timerange):
```
shankari: so just to clarify, does this build the model and use it to generate predictions?
I thought we were going to split that...

emission/analysis/section_features.py
```diff
@@ -140,14 +162,50 @@ def calVCR(segment):
...
+
 def calSpeeds(segment):
-  trackpoints = segment['track_points']
-  if len(trackpoints) == 0:
-    return None
-  return calSpeedsForList(trackpoints)
+    try:
+        return segment['speeds']
+    except KeyError:
+        return []
```
jensteff: Since in our if statements, we are checking `speeds != None and len(speeds) > 0` it doesn't matter as much here. But in the future, since we're usually expecting a list from speeds, it might be better to return an empty list instead of none, since we might do some list operations, like check the length, later, and if we returned None, it would throw an error. In that way, i think it is safer to return the empty list.

emission/analysis/classification/inference/mode.py
```diff
@@ -270,42 +261,41 @@ def getGroundTruthMode(self, section):
...
+  
     # Deal with unknown modes like "airplane"
     try:
       featureMatrix[i, 2] = section['sensed_mode']
     except ValueError:
       featureMatrix[i, 2] = 0
 
-    featureMatrix[i, 3] = section['section_id']   ####
-    #featureMatrix[i, 4] = easf.calAvgSpeed(section)
+    #featureMatrix[i, 3] = section['_id']   ####
```
jensteff: Yes, I intentionally commented out the `section[_id]` bit because it is a string and not a number, so it makes so sense to put it as an entry in the matrix.

emission/analysis/classification/inference/mode.py
jensteff: Yes, we are going to split it.

Issue #497, opened 2017-07-04
Track the first running of the pipeline reset to handle the badly formatted android motion_activity


shankari:
It's just useful to document the steps for the record

shankari:
First, determine the format of the new motion activity objects, and change the server code to account for them.

shankari:
Here's an example

```
In [515]: edb.get_timeseries_error_db().find_one({"metadata.key": "background/motion_activity"})
Out[515]:
{u'_id': ObjectId('594c9334cb17471ac033ee7e'),
 u'data': {u'zzbjA': 4, u'zzbjB': 75},
 u'metadata': {u'key': u'background/motion_activity',
  u'platform': u'android',
  u'read_ts': 0,
  u'time_zone': u'America/Los_Angeles',
  u'type': u'sensor-data',
  u'write_ts': 1498187291.604},
 u'user_id': UUID('08b31565-f990-4d15-a4a7-89b3ba6b1340')}
```

The `write_ts` is 

```
In [120]: arrow.get(1498187291)
Out[120]: <Arrow [2017-06-23T03:08:11+00:00]>
```

Last release was on Jun 9, 2017. So this is definitely the new format.

Let's do some more sorting...

```
In [517]: list(edb.get_timeseries_error_db().find({"metadata.key": "background/motion_activity"}).sort("metadata.write_ts", 1).limit(3))
Out[517]:
[{u'_id': ObjectId('5923ce3acb17471ac015a27b'),
  u'data': {u'zzbjA': 3, u'zzbjB': 77},
  u'metadata': {u'key': u'background/motion_activity',
   u'platform': u'android',
   u'read_ts': 0,
   u'time_zone': u'America/Los_Angeles',
   u'type': u'sensor-data',
   u'write_ts': 1495343353.589},
  u'user_id': UUID('fd7b4c2e-2c8b-3bfa-94f0-d1e3ecbd5fb7')},
 {u'_id': ObjectId('5923ce3acb17471ac015a27c'),
  u'data': {u'zzbjA': 5, u'zzbjB': 100},
  u'metadata': {u'key': u'background/motion_activity',
   u'platform': u'android',
   u'read_ts': 0,
   u'time_zone': u'America/Los_Angeles',
   u'type': u'sensor-data',
   u'write_ts': 1495343353.614},
  u'user_id': UUID('fd7b4c2e-2c8b-3bfa-94f0-d1e3ecbd5fb7')},
[...]

PR #498, opened 2017-07-04
Update the incoming formatter to match the update to the google play …

shankari:
…library

library was updated in phone release v2.2.0
We really need to resolve
https://github.com/e-mission/e-mission-data-collection/issues/80

This is basically the only thing we have to change in the formatters
and it is very annoying to have to keep doing it.


shankari:
The deployment to deal with fixing this is at
https://github.com/e-mission/e-mission-server/issues/497

shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #503, opened 2017-07-05
Several minor fixes to issues encountered while running the pipeline

shankari:
Issues are in https://github.com/e-mission/e-mission-server/issues/497
Fixes:
- Handle the `user_id = None` case
- Added additional logging for debugging
- Handle the case in which the app was installed after the reset point


shankari:
jenkins, test this please

PR #506, opened 2017-07-05
Habitica Notification  

sunil07t:
- Quest invitation notification

### DIFF
--- a/webapp/www/css/redirect.css
+++ b/webapp/www/css/redirect.css
@@ -1,21 +1,20 @@
 .redirect div {
-  padding: 10px 25% 5px 25%;
+  padding: 10px 5% 5px 5%;
 }
 
-.redirect h1 {
+.redirect h2 {
   /* Welcome to E-mission: */
   text-align: center;
   font-family: AvenirNext-Bold;
   color: #4A4A4A;
-  margin-bottom: 50px;
+  margin-bottom: 20px;
 }
 
-.redirect h3 {
+.redirect h4 {
   /* Install E-mission: */
   font-family: AvenirNext-DemiBold;
   color: #4A4A4A;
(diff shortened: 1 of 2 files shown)


shankari:
ok to test

emission/net/ext_service/habitica/notification.py
```diff
@@ -0,0 +1,40 @@
...
+			party_name = party_res.data.name
+			users_RSVPNeeded[user_id] = party_name
+
+	# Title for pending quest notification
+	# Message an json data is empty to keep notification short
+	message = ""
+	json_data = {}
+
+	for user_id in users_RSVPNeeded:
+		title = users_RSVPNeeded[user_id] " sent a quest invitation"
```
shankari: so `users_RSVPNeeded[user_id]` is the party, right? I know that you should restructure this, but the name of the variable is also confusing.

emission/net/ext_service/habitica/notification.py
```diff
@@ -0,0 +1,40 @@
...
+		# Add users and thier party name to list with pending quest invitation
+		if(user.party.quest.RSVPNeeded):
+			party_res = proxy.habiticaProxy(user_id, 'GET', method_url_get_party, None)
+			party_name = party_res.data.name
+			users_RSVPNeeded[user_id] = party_name
+
+	# Title for pending quest notification
+	# Message an json data is empty to keep notification short
+	message = ""
+	json_data = {}
```
shankari: when the user gets this message, you want them to go to the game screen, right?
you can extend the push notify code to handle redirects, similar to 
https://github.com/e-mission/e-mission-phone/commit/25d078e716a1cada0cc996f69132450d81165267

PR #513, opened 2017-08-16
Minor fixes from the server

shankari:
- Support a more performant way to retrieve information across all users
  for the reset pipeline
- Handle the case when the specified user does not exist


shankari:
jenkins, test this please

PR #514, opened 2017-08-26
Changes to convert old-style stats (in the clientstats_db) to new style stats in the timeseries db

shankari:
This allows us to use standard techniques to analyse and process the data.


shankari:
jenkins, test this please

PR #525, opened 2017-10-16
Support multiple auth modes

shankari:
- Restructure the auth mode into its own module
- Move the google auth into its own auth method since it is now only one among many
- Add a new _token_list_ module that supports a list of hardcoded "tokens" that users can use to login without providing login credentials.


shankari:
So we finally begin populating the `auth` directory that I apparently created when trying to set up the module structure, to be populated later. And that later is now!

shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #527, opened 2017-11-08
requirements for clean python 2.7

jf87:
added missing requirements


shankari:
We've moved to python3 and setup a conda virtual environment (#534)
So this is now obsolete.

PR #529, opened 2017-11-21
Move out moves tests which are not important any more

shankari:
Since we no longer integrate with moves


shankari:
jenkins, test this please

PR #543, opened 2017-12-16
Fix issues encountered while running the pipeline

shankari:
in particular:
https://github.com/e-mission/e-mission-server/issues/530#issuecomment-352197949
https://github.com/e-mission/e-mission-server/issues/530#issuecomment-352202072


shankari:
jenkins, test this please

PR #559, opened 2018-01-13
Handle two kinds of data.ts correctly

shankari:
both strings (which are in ms) and floats (which are in seconds)


shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #566, opened 2018-02-04
environment36.yml for windows

omrachel:


shankari:
As a general rule-of-thumb, you want to avoid removing the version completely. If you have no version, then anybody who installs the server later could end up with an different version with different API calls. This happened frequently before, which is why I added the version in the first place.

I'm calling out the various places where you removed the version completely. Can you restore the major version, leaving the patch level unchanged, and try re-installing? If some packages still don't work, let me know and I'll decide what to do.

setup/environment36.yml
```diff
@@ -48,15 +48,15 @@ dependencies:
...
 - flask=0.12.2=py36_0
 - flask-cors=3.0.2=py36_0
-- freetype=2.5.5=2
+- freetype=2.5.5
 - get_terminal_size=1.0.0=py36_0
 - gevent=1.2.1=py36_0
 - greenlet=0.4.12=py36_0
 - h5py=2.7.0=np112py36_0
-- hdf5=1.8.17=1
+- hdf5
```
shankari: does `hdf5=1.8.17` work? 
Looks like it should
https://anaconda.org/conda-forge/hdf5/files

```
win-64/hdf5-1.8.17-vc9_11.tar.bz2
```

setup/environment36.yml
```diff
@@ -48,15 +48,15 @@ dependencies:
...
 - heapdict=1.0.0=py36_1
 - html5lib=0.999=py36_0
-- icu=54.1=0
+- icu
```
shankari: Does `icu=58.2` work?
https://anaconda.org/anaconda/icu/files

shankari: Actually, try removing this line completely - the packages that need it will automatically install the correct version.

setup/environment36.yml
```diff
@@ -65,22 +65,21 @@ dependencies:
 - ipywidgets=6.0.0=py36_0
 - isort=4.2.5=py36_0
 - itsdangerous=0.24=py36_0
-- jbig=2.1=0
 - jdcal=1.3=py36_0
 - jedi=0.10.2=py36_2
 - jinja2=2.9.6=py36_0
-- jpeg=9b=0
+- jpeg
```
shankari: Try `jpeg=9b`
https://anaconda.org/anaconda/jpeg/files

setup/environment36.yml
```diff
@@ -65,22 +65,21 @@ dependencies:
...
 - lazy-object-proxy=1.2.2=py36_0
-- libiconv=1.14=0
-- libpng=1.6.27=0
-- libtiff=4.0.6=3
-- libxml2=2.9.4=0
-- libxslt=1.1.29=0
+- libiconv=1.14
+- libpng=1.6.27
+- libtiff=4.0.6
+- libxml2
```
shankari: Try `libxml2=2.9.4`
Again, there appears to be a Windows version for this
https://anaconda.org/anaconda/libxml2/files

shankari:
ok to test

PR #509, opened 2017-07-13
Get the mode inference pipeline to work with the new architecture

shankari:
Thanks to @jensteff for most of the actual porting
Design decisions tracked at https://github.com/e-mission/e-mission-server/issues/508

PR #571, opened 2018-02-22
Minor changes to the seed pipeline from running on real data

shankari:
As always, no commit survives the first encounter with real data
- Ensure that the backup database step is read from the database at the correct URL
- deal with some data that is still in lat, lng format during conversion

```
WARNING:root:Found OutOfRangeError while converting=[37.2793693359, -121.9477739912], swapping
WARNING:root:Found OutOfRangeError while converting=[37.8450065077, -122.2505039016], swapping
WARNING:root:Found OutOfRangeError while converting=[37.7934984948, -122.396811555], swapping
WARNING:root:Found OutOfRangeError while converting=[37.8701622504, -122.2680943316], swapping
WARNING:root:Found OutOfRangeError while converting=[37.786435505, -122.4023798827], swapping
WARNING:root:Found OutOfRangeError while converting=[37.8702876992, -122.2685171307], swapping
WARNING:root:Found OutOfRangeError while converting=[37.7890437261, -122.4031406148], swapping
```

### DIFF
--- a/emission/analysis/classification/inference/mode/seed/pipeline.py
+++ b/emission/analysis/classification/inference/mode/seed/pipeline.py
@@ -51,7 +51,7 @@ class ModeInferencePipelineMovesFormat:
     logging.info("initial loadTrainingDataStep DONE")
 
     logging.debug("finished loading current training set, now loading from backup!")
-    backupSections = MongoClient('localhost').Backup_database.Stage_Sections
+    backupSections = MongoClient(edb.url).Backup_database.Stage_Sections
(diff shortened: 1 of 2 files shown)


shankari:
Main changes to the pipeline were in #509 
They were merged to allow me to generate the static model from real data, and use it for prediction going forward.
This change fixes issues encountered while generating the static model on real data.
A subsequent change will fix any issues encountered while the generated model for prediction.

shankari:
Should I check in the generated model? In general, you don't want to check in big files and program outputs. But in this case, checking in the model allows us to avoid checking in the raw data. Will solve that problem in the next checkin.

shankari:
jenkins, test this please.

PR #572, opened 2018-02-23
Actually switch to the new pipeline

shankari:
- Add the classification step
- Return all results based on the inferred section


shankari:
ok, so that is the only failing test. Let us try to run it individually and restore it to see if we can get it to work. If it does not work on one more try, I am going to disable it for now and work out the kinks later.

shankari:
First, confirm that the two tests are identical

```
$ diff emission/tests/netTests/TestMetricsInferredSections.py emission/individual_tests/TestMetricsInferredSections.py
$ 
```

shankari:
Argh! I am not sure why the new version of the test is not getting picked up. Will debug with shane later. For now, disabling the test and merging changes.

shankari:
Ok, so running the test in a separate batch file seems to pass.
Committing the change now...

PR #576, opened 2018-03-14
Minor fixes to inference before redoing it completely

shankari:


shankari:
jenkins, test this please

PR #582, opened 2018-06-11
Pr branch

jesbu1:


shankari:
Why do you need to check in the `matplotlib-ubuntu-*/fontList.py3k.cache`. Those look like matplotlib font caches, which should definitely not be checked in.

shankari:
what are the `emission/tests/data/real_examples/*.timeline` files. They are new here, but I don't see them used anywhere in the tests...

jesbu1:
No idea what the real_examples/ files are

PR #584, opened 2018-07-10
README.md

tarunrkk:
A note regarding the dependency for Javascript has been added.


README.md
```diff
@@ -132,6 +132,8 @@ Tip: Run "bower install" instead if you are prompted password for 'https://githu
     $ cd webapp
     $ bower update
```
shankari: Maybe move the note to the top so that people can see it before they try to type out `bower update`?

PR #594, opened 2018-08-21
Add docker-related files

atton16:
Hi, I have added the docker-related files here. Please also checkout `docker/README.md` for how to set it up and running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+# python 3
+FROM continuumio/anaconda3
+
+MAINTAINER Attawit Kittikrairit
+
+# set working directory
+WORKDIR /usr/src/app
+
+# clone from repo
+RUN git clone https://github.com/e-mission/e-mission-server.git .
+
+# setup python environment
+RUN conda env update --name emission --file setup/environment36.yml
+RUN /bin/bash -c "source activate emission; pip install six --upgrade"
+
+# install nodejs, npm and bower
+RUN apt-get update
+RUN apt-get install -y build-essential
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+RUN apt-get -y install nodejs
+RUN npm install -g bower
+WORKDIR /usr/src/app/webapp
+RUN bower update --allow-root
+WORKDIR /usr/src/app
+
+# install nano for editing
+RUN apt-get -y install nano
+
+# start the server
+ADD docker/start_script.sh /usr/src/app/start_script.sh
+RUN chmod u+x /usr/src/app/start_script.sh
+
+EXPOSE 8080
+
+CMD ["/bin/bash", "/usr/src/app/start_script.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # python 3
-FROM continuumio/anaconda3
+FROM continuumio/miniconda3
 
 MAINTAINER Attawit Kittikrairit
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN bower update --allow-root
 WORKDIR /usr/src/app
 
 # install nano for editing
-RUN apt-get -y install nano
+RUN apt-get -y install nano vim
 
 # start the server
 ADD docker/start_script.sh /usr/src/app/start_script.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN /bin/bash -c "source activate emission; pip install six --upgrade"
 
 # install nodejs, npm and bower
 RUN apt-get update
+RUN apt-get install build-essential
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
 RUN apt-get -y install nodejs
 RUN npm install -g bower
@@ -24,6 +25,10 @@ WORKDIR /usr/src/app
 
 # install nano for editing
 RUN apt-get -y install nano vim
+
+# cleanup
+RUN apt-get -y remove --purge build-essential
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # start the server
 ADD docker/start_script.sh /usr/src/app/start_script.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,6 @@ RUN /bin/bash -c "source activate emission; pip install six --upgrade"
 
 # install nodejs, npm and bower
 RUN apt-get update
-RUN apt-get install -y build-essential
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
 RUN apt-get -y install nodejs
 RUN npm install -g bower
--- a/docker/README.md
+++ b/docker/README.md
@@ -59,6 +59,5 @@
      --restart=always \
      --mount type=bind,source="$(pwd)"/conf/storage/db.conf.docker.sample,target=/usr/src/app/conf/storage/db.conf,readonly \
      --mount type=bind,source="$(pwd)"/conf/net/api/webserver.conf.docker.sample,target=/usr/src/app/conf/net/api/webserver.conf,readonly \
-     --mount type=bind,source="$(pwd)"/docker/config/google_auth.json,target=/usr/src/app/conf/net/auth/google_auth.json,readonly \
      e-mission-server:latest
    ```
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,35 @@
+# python 3
+FROM continuumio/anaconda3
+
+MAINTAINER Attawit Kittikrairit
+
+# set working directory
+WORKDIR /usr/src/app
+
```
shankari: Don't you need to install mongodb as well?

docker/start_script.sh
```diff
@@ -0,0 +1,7 @@
+#!bin/bash
+
+# change python environment
+source activate emission
+
+# launch the webapp
+./e-mission-py.bash emission/net/api/cfc_webapp.py
```
shankari: And launch it?

docker/README.md
```diff
@@ -0,0 +1,28 @@
...
+3. Run the server
+
+   Assuming `config` files are located at `docker/config` folder
+
+   ```
+   docker run -d \
+     -p 8080:8080 \
+     --name=e-mission-server-1 \
+     --net="e-mission" \
+     --mount type=bind,source="$(pwd)"/docker/config/db.conf,target=/usr/src/app/conf/storage/db.conf,readonly \
```
shankari: This is pretty cool. I assume I leave out this and the next two lines if I just want to start the container with defaults?
Because for development, people may want to just start the app and not want to figure out how to set config files.

shankari: You probably want to have development versus production/deployment scripts anyway because the port mapping will be different. Development maps `8080:8080` but I would assume production should do `443:443`. And production will want to configure SSL certificates as well.

Dockerfile
atton16: I will be submitting Dockerfile and instruction for mongodb separately on `e-mission-docs` repo.

docker/README.md
atton16: For production, we are using another script that runs as a "web gateway". This helps us separate the problem and handle the SSL modularly. Again, I will be submitting the files in `e-mission-docs` repo along with the mongodb.

Dockerfile
atton16: Or should we include mongodb script in this repo??

shankari: I think that the `mongodb` script should go into this repo. Think about it:
- is mongodb needed if you are running the server? Yes
- is mongodb not needed if you are not running the server? Yes

So basically, mongodb is a server dependency and only a server dependency. It should go here.

docker/README.md
shankari: @atton16 I think that the production script should go here as well. Basically, I think that there should be two separate scripts, one for each conceptual module. After the UI component is separated from the server component (#596), we could have three.

We want to have as few script locations as possible - fewer commands to type out - while still supporting flexible deployments. So basically one script per _module_ or _tier_

shankari:
So I thought about it some more, and I think that it would be good to have the mongodb as a separate script, but checked in to this repository since it is a dependency. But it is true that in the multi-tier case, the database tier will be configured separately from the appserver tier. So make it easy to copy that script to the host running the database and people can use it for multi-tier (rare) and everything-on-one-host (common).

shankari:
@sam11b @valin @ecvick85 you can try this out while setting up the new server instead of following the manual steps

shankari:
please make sure to report back here and tell us how it went!

Dockerfile
atton16: Check out my last commit

docker/start_script.sh
atton16: This only occurs when we do `docker run`

shankari:
Is this primarily focused around starting the devapp? What about the analysis pipeline? The pipeline is run periodically, and I currently launch it from a cronjob.
https://github.com/e-mission/e-mission-docs/blob/master/docs/e-mission-server/deploying_your_own_server_to_production.md#the-analysis-pipeline

However, this is clearly a good start, so I approve of merging this once we get confirmation from somebody that it works.

@brianmickel, @sam11b @valin @ecvick85 please let me whether this works for your setup. I'm also donwloading Docker now so that I can test myself.

atton16:
Oh the daily pipeline task is new for me. Let me try it with our server first then I will update the file once I understand the process more clearly. Also, since the pipeline process can be trigger separately from the server app. I am thinking about building a separate docker image as `e-mission-pipeline-runner` just to run the pipeline with configurable execution frequency.

atton16:
Also, you might wanted to consider upload Docker image on [Docker Hub](https://hub.docker.com/) to even make it much more easier to others who do not wanted to tangle with the building process (which might yield a different result on different OSs). So others can simple execute a `docker run` and go!

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# clone from repo
+RUN git clone https://github.com/e-mission/e-mission-server.git .
+
+# setup python environment
+RUN conda env update --name emission --file setup/environment36.yml
+RUN /bin/bash -c "source activate emission; pip install six --upgrade"
+
+# install nodejs, npm and bower
+RUN apt-get update
+RUN apt-get install -y build-essential
```
shankari: why do you need `build-essential`?
That installs the entire build toolchain including gcc.
We are not really building any c programs - `apt-get install nodejs` should install the compiled version of node...

shankari:
The server is definitely running but I cannot connect to it.

```
# apt-get install procps
# ps -aef
....
root        11    10  0 21:34 ?        00:00:00 python emission/net/api/cfc_webapp.py
```

docker/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+   Assuming `config` files are located at `docker/config` folder
+
+   ```
+   docker run -d \
+     -p 8080:8080 \
+     --name=e-mission-server-1 \
+     --net="e-mission" \
+     --restart=always \
+     --mount type=bind,source="$(pwd)"/docker/config/db.conf,target=/usr/src/app/conf/storage/db.conf,readonly \
```
shankari: you can check in the default docker conf as `conf/storage/db.conf.docker.sample`.
Then you can remove step 4 and change this to `source="$(pwd)"/conf/storage/db.conf.docker.sample`

docker/README.md
```diff
@@ -0,0 +1,64 @@
...
+     --mount type=bind,source="$(pwd)"/docker/config/webserver.conf,target=/usr/src/app/conf/net/api/webserver.conf,readonly \
```
shankari: Ditto, where `webserver.conf.docker.sample` changes `"host" : "localhost"` -> `"host" : "0.0.0.0"`

docker/README.md
```diff
@@ -0,0 +1,64 @@
...
+     --mount type=bind,source="$(pwd)"/docker/config/google_auth.json,target=/usr/src/app/conf/net/auth/google_auth.json,readonly \
```
shankari: This is not going to be required for all users. If the users choose to use google auth, they will have to do this; if they use a different auth method, they will need to copy the appropriate json file; if they want to continue with development and the `skip` mode, they will not need to copy any file (copying this file was the root cause of #605).

We should remove this from the README and add a note saying that configuration options can be specified using this mechanism

docker/README.md
```diff
@@ -0,0 +1,64 @@
...
+   We will be using Official MongoDB Docker image, so no need to build one.
+
+   ```
+   docker run -d \
+     --name=e-mission-mongo-1 \
+     --net="e-mission" \
+     --restart=always \
+     --mount source=e-mission-mongo-1_data,target=/data/db \
+     --mount source=e-mission-mongo-1_config,target=/data/configdb \
+     mongo:latest \
```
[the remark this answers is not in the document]
atton16: That's a good recommendation. We don't normally use Docker compose just because we wanted to stay bare-metal. But this time we will definitely do.

Dockerfile
atton16: There are a lot of missing dependencies. many of npm packages requires compile from source when you do `npm install`.

shankari:
@atton16 you applied the suggestion, but didn't make the change in the comments.

> you can check in the default docker conf as conf/storage/db.conf.docker.sample.
> Then you can remove step 4 and change this to source="$(pwd)"/conf/storage/db.conf.docker.sample

shankari:
@atton16 also, I suggested removing the build-essential package *after* installing npm.

> hm, I removed build-essential after container was created and there didn't appear to be any dependencies. So maybe we could remove it at the end after the rest of the installation is complete. I am not sure exactly how much disk space we will save, though...

PR #608, opened 2018-10-26
Greentrip Version 01

valin1:
Applied diff patch from emission/greentrip and forked branch and the suggestion system works for addresses that have a business name associated through Google Map's API call, working on switching over to Nominatim.


shankari:
@valin1 this is just the patch file. You need to actually apply it to your branch.

shankari:
You might want to edit it to remove changes to `Timeseries_Sample.ipynb` etc for example

shankari:
or `git checkout` in the original branch before generating the patch

PR #609, opened 2018-10-26
Greentrip Version 01: Applied the patch 

valin1:
Applied diff patch from emission/greentrip and forked branch (now it's actually applied to the branch) and the suggestion system works for addresses that have a business name associated through Google Map's API call, working on switching over to Nominatim. 

Removed changes in TimeSeriesSample.ipynb


shankari:
aha, but you don't want to commit the patch now that you have already applied it

PR #613, opened 2018-11-14
Converted functions to use Nominatim and addressed previous pull request comments

valin1:
Changed the API connection to reflect the nominatim function
Added comments to clarify the use of each function, nominatim functions work
- Will change the Nominatim to use our server Nominatim, but right now it is using the main URL for all the API requests
Switched queries to use conf files
Changed parameters in nominatim.py file for try catch on conf files as well, also added assert statements (to check if file path to conf file exists) in the suggestions_sys file
- To address question as to how someone would use the suggestion system and obtain the conf files with the API keys, they will have to contact me for the conf files and I will send the files to them


emission/core/wrapper/suggestion_sys.py
```diff
@@ -193,65 +235,90 @@ def return_address_from_location_yelp(location='0,0'):
...
 '''
 def review_start_loc(location = '0,0'):
     try:
         #Off at times if the latlons are of a location that takes up a small spot, especially boba shops
 
         #IF RETURN_ADDRESS_FROM_LOCATION HAS A BUSINESS LOCATION ATTACHED TO THE ADDRESS
-        if (len(return_address_from_location_yelp(location)) == 3):
-            business_name, city, address = return_address_from_location_yelp(location)
+        if (len(return_address_from_location_google(location)) == 3):
+            business_name, city, address = return_address_from_location_google(location)
```
shankari: why is this still calling `return_address_from_location_google`? I thought you were not using the google API?

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -15,18 +15,18 @@
...
     googlemaps_key_file = open("conf/net/ext_service/googlemaps.json")
-    GOOGLE_MAPS_KEY = json.load(googlemaps_key_file)["api_key"]
+    GOOGLE_MAPS_KEY = json.load(googlemaps_key_file)["access_token"]
 except:
     print("google maps key not configured, falling back to nominatim")
 
 try:
     nominatim_file = open("conf/net/ext_service/nominatim.json")
-    NOMINATIM_QUERY_URL = json.load(nominatim_file)["query_url"]
+    NOMINATIM_QUERY_URL = json.load(nominatim_file)["base_url"]
```
shankari: why not just change your conf file to use `query_url` instead of `base_url`?
In general, you want to adapt new code to what already exists instead of changing existing code and risking a regression. If you can't do that for some reason, you need to explain why

PR #617, opened 2018-11-30
New docker images for the emission-server + updated docker instructions

alvinalexander:


shankari:
Can you pull out only the docker changes into a separate pull request? There is no reason why the docker changes and the trip generator need to be combined, and it makes the PR harder to review.
Also, why do a two step process with `e-mission-server-base` and `e-mission-server`? Can't you include the environment variables directly in `e-mission-server`? Can you document your rationale?

Also, what is `setup/alvin_environment.yml`? Doesn't seem to be used anywhere...

alvinalexander:
Reason for two images: The base image is less like to change and takes a long time to build(~6GB). The server-image is more likely to change and can be built in a couple seconds.  For instance, you might want to add or remove environment variables, change the startup shell script, mount different config files etc. However, it seems unnecessary to rebuild the whole image(pull from github, install the conda env etc.) everytime you want to make changes to the configuration of the server.

alvinalexander:
setup/alvin_environemnt.yml includes the packages installed to create the data generator

shankari:
> The base image is less like to change and takes a long time to build(~6GB).

The base image is no longer 6GB after the shift from anaconda -> miniconda. It is more like 2GB, 
https://hub.docker.com/r/emission/e-mission-server/tags/
and can potentially get smaller as we prune the set of packages based on what @skxu3 finds for the minimal install.

So I don't think that the split is necessary and would prefer to have one file until it stabilizes with the enclave code.

shankari:
superceded by #618

PR #618, opened 2018-12-03
New docker setup for emission-server

alvinalexander:


docker/Dockerfile-base
```diff
@@ -0,0 +1,31 @@
+# python 3
+FROM continuumio/anaconda3
```
shankari: This file doesn't include the changes to the Dockerfile in 
https://github.com/e-mission/e-mission-server/pull/594

This is probably because you split the Dockerfile into two so the changes to `Dockerfile` were not merged into `Dockerfile-base`. Please merge the changes manually.

Also, as I pointed out in https://github.com/e-mission/e-mission-server/pull/617#issuecomment-443917798, after the change to miniconda, the server image is only 2GB. So I would recommend keeping one image for consistency but I will not oppose two images.

docker/Dockerfile
```diff
@@ -1,15 +1,13 @@
...
 FROM continuumio/miniconda3
 
-MAINTAINER Attawit Kittikrairit
-
 # set working directory
 WORKDIR /usr/src/app
 
 # clone from repo
-RUN git clone https://github.com/e-mission/e-mission-server.git .
+RUN git clone https://github.com/alvinalexander/e-mission-server.git .
```
shankari: leave this as e-mission in the checked in code. You will soon be merging the generation code into the server repository as well!

docker/Dockerfile
```diff
@@ -24,14 +22,19 @@ RUN bower update --allow-root
 WORKDIR /usr/src/app
 
 # install nano for editing
-RUN apt-get -y install nano vim
+RUN apt-get -y install nano
```
shankari: why remove vim? :(
I think that we should either not have any editors to keep the image size small. People can just install their favorite editor if they open a shell. Or we should also install vi/vim, which is the standard editor in *nix-like systems

docker/Dockerfile
```diff
@@ -1,15 +1,13 @@
 # python 3
 FROM continuumio/miniconda3
 
-MAINTAINER Attawit Kittikrairit
```
shankari: put me as the maintainer instead

`MAINTAINER K. Shankari (shankari@eecs.berkeley.edu)`

docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    networks:
+      - emission
+  db:
+    image: mongo:latest
+    deploy:
+      replicas: 1
+      restart_policy:
+        condition: on-failure
+    ports:
+      - "27017:27017"
```
shankari: this is only if you want to access the DB from the host. Add a note to that effect?

docker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+#If you want to persist the data in the DB you need to mount a volume.
+    #Instructions can be found here: https://docs.docker.com/compose/compose-file/#volume-configuration-reference
```
shankari: I think you should mount a volume by default because the default use case will be to persist data. You can mount a directory on the local host if you would like, similar to the prior instructions.

docker/start_script.sh
```diff
@@ -1,5 +1,27 @@
-#!bin/bash
+#!/usr/bin/env bash
+#Configure web server
 
+#set database URL using environment variable
+echo ${DB_HOST}
```
shankari: nice!

docker/Dockerfile
shankari: This is unresolved.

docker/Dockerfile
shankari: This is unresolved

PR #619, opened 2018-12-05
Suggestion System Debugged 

valin1:
All the nominatim functions are switched to using the reverse geocode function on nominatim.py. 
The function calculate_yelp_server_suggestion_nominatim does produce suggestions now, from what I tested using the trips that I loaded in my database. 
Here is an example suggestion that was produced: 
{'message': "Why didn't you bike from Maxwell Family Field, Piedmont Avenue, Southside, Berkeley, Alameda County, 94704, United States of America to Guacamole 61 (tap me to view) Guacamole 61 has better reviews, closer to your original starting point, and has a rating of 4.0", 'method': 'bike'} 

I haven't finished implementing the single trip function, in which the user can choose which trip they would like to see a suggestion for, but there is at least a suggestion showing up for at least one of the trips.


emission/core/wrapper/suggestion_sys.py
```diff
@@ -38,18 +39,24 @@
 """
 Checks if conf files exists or not. The conf files will be given to the user through request.
 """
-assert os.path.exists(google_maps_path), "File %s not found" % google_maps_path
-assert os.path.exists(yelp_json_path), "File %s not found" % yelp_json_path
-assert os.path.exists(nominatim_path), "File %s not found" % nominatim_path
+try:
+    google_maps_json = open('conf/net/ext_service/googlemaps.json', 'r')
```
shankari: why do you need these? the `conf/net/ext_service` files are already loaded in `nominatim.py`, aren't they? Please remove

emission/core/wrapper/suggestion_sys.py
```diff
@@ -38,18 +39,24 @@
...
+    google_maps_auth = json.load(google_maps_json)
+except:
+    print("google maps key not configured, falling back to nominatim")
 
-google_maps_json = open('conf/net/ext_service/googlemaps.json', 'r')
-yelp_json = open('conf/net/ext_service/yelpfusion.json', 'r')
-nominatim = open('conf/net/ext_service/nominatim.json', 'r')
+try:
+    nominatim = open('conf/net/ext_service/nominatim.json', 'r')
```
shankari: ditto

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -84,13 +84,13 @@ def get_json_reverse(cls, lat, lng):
...
-        # except:
-        #     print "defaulting"
+        try:
+            jsn = cls.get_json_reverse(lat, lng)
+            business_name = jsn["display_name"]
+            address = jsn["address"]
+            return business_name, address
+        except:
+            print("defaulting")
         return _do_google_reverse(lat, lng) # Just in case
```
shankari: shouldn't this be within the `except`:

emission/core/wrapper/suggestion_sys.py
```diff
@@ -38,18 +39,24 @@
...
+    nominatim_auth = json.load(nominatim)
+except:
+    print("nominatim not configured either, place decoding must happen on the client")
 
+try:
```
shankari: you do need this but not the others

emission/core/wrapper/suggestion_sys.py
```diff
@@ -91,10 +98,9 @@ def request(host, path, api_key, url_params=None):
         'Authorization': 'Bearer %s' % api_key,
     }
 
-    print(u'Querying {0} ...'.format(url))
+    # print(u'Querying {0} ...'.format(url))
 
     response = requests.request('GET', url, headers=headers, params=url_params)
```
shankari: Can remove the TODO about `urllib` given that you have already moved to `requests`?

emission/core/wrapper/suggestion_sys.py
```diff
@@ -360,7 +358,39 @@ def geojson_to_lat_lon_separated(geojson):
...
+        mode = cleaned_sections.iloc[section_counter]["sensed_mode"]
+        return mode, section_counter + 1
+    while endsec_loc_lat != end_loc_lat and endsec_loc_lon != end_loc_lon and section_counter < len(cleaned_sections) :
+        mode = cleaned_sections.iloc[section_counter]["sensed_mode"]
+        modes_from_section.append(mode)
+        endsec_location = cleaned_sections.iloc[section_counter]["end_loc"]
+        endsec_loc_lon, endsec_loc_lon = geojson_to_lat_lon_separated(endsec_location)
+        if (mode == 0):
+            return mode, section_counter+1
+    return mode, section_counter + 1
```
shankari: which mode are you returning for the trip? if there is only one mode, you return it, but I am not sure what you are doing with multiple modes?

valin1: For the modes, I will be redoing the function to determine mode of comparison for the suggestion function and currently am not using the function for the suggestion function.

emission/core/wrapper/suggestion_sys.py
valin1: Removed the nominatim conf files

valin1:
Fix the functions that use google maps json to switch to nominatim.py and get rid of the google maps json try accept

PR #620, opened 2018-12-27
Function using only Trip ID completed and should be working

valin1:
The trip ID function works, using the trip ID and user ID as inputs, still debugging on different trips, because some just returns None from the function which may be the reason why nothing shows up on the phones. 

Also, I don't know why it's unable to automatically merge? update: never mind saw there were some merge conflicts in the original single trip function


emission/core/wrapper/suggestion_sys.py
```diff
@@ -437,23 +438,32 @@ def review_start_loc_nominatim(lat, lon):
...
 '''
 
-def calculate_yelp_server_suggestion_singletrip_nominatim(tripid):
-    return_obj = { 'message': "Good job walking and biking! No suggestion to show.", 'method' : 'bike'}
-    start_lat, start_lon = geojson_to_lat_lon_separated(tripid["start_loc"])
-    end_lat, end_lon = geojson_to_lat_lon_separated(tripid["end_loc"])
+
+def calculate_yelp_server_suggestion_singletrip_nominatim(tripid, uuid):
+    all_users = pd.DataFrame(list(edb.get_uuid_db().find({}, {"uuid": 1, "_id": 0})))
+    user_id = all_users.iloc[all_users[all_users.uuid == uuid].index.tolist()[0]].uuid
```
shankari: again, you have the `uuid`, but you find all the entries from the database and then find the uuid of the matching one. I think I commented on this in an earlier review but didn't insist on it being removed. You should really remove this pattern from all places in your code.
- is there any case in which `user_id != uuid`?
- you really want to avoid the use of `edb` in the code since it impedes the ability to switch to other timeseries (potentially for scalability) in the future. Please remove it from the imports so you are not tempted to use it.

emission/core/wrapper/suggestion_sys.py
```diff
@@ -437,23 +438,32 @@ def review_start_loc_nominatim(lat, lon):
...
+    timeseries = esta.TimeSeries.get_time_series(user_id)
+    cleaned_trips = timeseries.get_data_df("analysis/cleaned_trip", time_query = None)
+    spec_trip = cleaned_trips.iloc[cleaned_trips[cleaned_trips._id == tripid].index.tolist()[0]]
```
shankari: Again, this is the kind of code you would write in an exploratory setting (e.g. in `ipython`), not in a script. You are reading all trips into memory and then retrieving the trips that matches. This is fine if the user has 100 trips, but what happens if the user has 10,000 trips? You should really query directly using the `trip_id` as well. Look at the functions in `abstract_timeseries` to see what else you can use.

emission/core/wrapper/suggestion_sys.py
```diff
@@ -437,23 +438,32 @@ def review_start_loc_nominatim(lat, lon):
...
+    start_location = spec_trip.start_loc
+    end_location = spec_trip.end_loc
+    distance_in_miles = spec_trip.distance * 0.000621371
```
shankari: nit: why do you need to convert the distance into miles? Just curious.

emission/core/wrapper/suggestion_sys.py
```diff
@@ -437,23 +438,32 @@ def review_start_loc_nominatim(lat, lon):
 11: "AIR_ON_HSR"
 '''
 
-def calculate_yelp_server_suggestion_singletrip_nominatim(tripid):
-    return_obj = { 'message': "Good job walking and biking! No suggestion to show.", 'method' : 'bike'}
-    start_lat, start_lon = geojson_to_lat_lon_separated(tripid["start_loc"])
-    end_lat, end_lon = geojson_to_lat_lon_separated(tripid["end_loc"])
+
+def calculate_yelp_server_suggestion_singletrip_nominatim(tripid, uuid):
```
shankari: nit: IIRC, wrt coding convention, I typically put the uuid as the first argument, since it is implied that it is for that user.

emission/core/wrapper/suggestion_sys.py
shankari: or I think that there might be some decorators in `emission.storage.decorations` that might be useful.

emission/core/wrapper/suggestion_sys.py
valin1: The distance in miles is to compare with the calculated distances between the better suggested end point and original starting point.

emission/core/wrapper/suggestion_sys.py
```diff
@@ -12,6 +12,7 @@
 import pprint
 import requests
 import os
+import emission.net.ext_service.geocoder.nominatim as geo
 from emission.net.ext_service.geocoder.nominatim import Geocoder
```
shankari: you can then remove this import

emission/core/wrapper/suggestion_sys.py
```diff
@@ -187,48 +137,10 @@ def return_address_from_location_nominatim(lat, lon):
...
-    base_url = 'https://maps.googleapis.com/maps/api/geocode/json?'
-    latlng = 'latlng=' + location
-    try:
-        #This try block is for our first 150,000 requests. If we exceed this, use Jack's Token.
-        key_string = '&key=' + ACCESS_TOKEN
-        url = base_url + latlng + key_string #Builds the url
-        result = requests.get(url).json() #Gets google maps json file
-        cleaned = result['results'][0]['address_components']
 
+def return_address_from_google_trial(location):
```
shankari: nit: is this name right? the implementation seems to call `nominatim`

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -94,6 +99,82 @@ def reverse_geocode(cls, lat, lng):
             return _do_google_reverse(lat, lng) # Just in case
 
 ## Failsafe section
+def check_against_business_location(location='0, 0', address = ''):
```
shankari: wait, why are you passing in the location as a string instead of two ints? converting to/from string seems to be a lot of extra work.

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -94,6 +99,82 @@ def reverse_geocode(cls, lat, lng):
...
+            cleaned = result['results']
+            for i in cleaned:
+                if address == i['vicinity']:
+                    return (True, i['name'])
+            else:
+                return (False, '')
+        except:
+            raise ValueError("Something went wrong")
+
+def return_address_from_location_google(location='0,0'):
```
shankari: ditto

shankari: isn't there an existing reverse geocode implementation for google in here? why not just re-use that?

emission/core/wrapper/suggestion_sys.py
shankari: ok, but is there a reason you can't perform that comparison directly in `km`? Most scientific work is done in SI units.

valin1:
Suggestions are being generated at least from all the trips I've tested but most are either categorized as "university housing" so there isn't really any suggestions for that and all the bugs that I know of should be fixed since most were not working because some of the functions were entering the except clause which messed things up. Also, in regards to the google reverse lookup function, I was trying to replace the google reverse lookup function with the nominatim.py version of the function, but it encountered an error in the pygeocoder file that is in anaconda, so I wasn't really sure how to debug through that file, so I just moved the google reverse functions from suggestion_sys.py to nominatim.py, since those were the functions that were calling the google json file again in suggestion_sys.py. I remember you saying that we should reduce the number of times of opening the same json file and nominatim.py was already calling the google maps and nominatim json files, so that's why these lookup functions are now in nominatim.py.

valin1:
Also, we are still using google reverse lookup in the functions in the except clause, because the nominatim reverse lookup sometimes is very vague and is unable to pinpoint the exact address and gives a broad "address", of the street and the city, without a full address with a specific location. Such as Piedmont Ave, Berkeley, CA

shankari:
Can you document all the stuff in these comments in the code? Now you know why people ask for well-commented code!

You still need to address a couple of the small cleanup/coding convention issues that I requested earlier.

emission/net/ext_service/geocoder/nominatim.py
shankari: This is still unresolved....

emission/core/wrapper/suggestion_sys.py
shankari: See https://github.com/e-mission/e-mission-server/pull/620#discussion_r244630629

emission/net/ext_service/geocoder/nominatim.py
shankari: and it also just looks ugly. using a comma separated set of ints is a presentation issue, conforming to the format for query parameters (I think). You should format values for presentation as late as possible to give greater flexibility in case of changes.

emission/core/wrapper/suggestion_sys.py
```diff
@@ -302,10 +310,21 @@ def calculate_yelp_server_suggestion_singletrip_nominatim(uuid, tripid):
     user_id = uuid
     timeseries = esta.TimeSeries.get_time_series(user_id)
     cleaned_trips = timeseries.get_data_df("analysis/cleaned_trip", time_query = None)
```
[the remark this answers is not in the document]
valin1: Used the timeseries.get_entry_from_id

valin1:
Resolved all the comments I believe?

emission/core/wrapper/suggestion_sys.py
```diff
@@ -13,6 +13,7 @@
 import os
 import emission.net.ext_service.geocoder.nominatim as geo
 from emission.net.ext_service.geocoder.nominatim import Geocoder
+import emission.storage.decorations.trip_queries as tripqu
```
shankari: nit: wait, why do you even need to import this? It is not used anywhere; you use the timeseries directly,

emission/core/wrapper/suggestion_sys.py
```diff
@@ -13,6 +13,7 @@
 import os
 import emission.net.ext_service.geocoder.nominatim as geo
 from emission.net.ext_service.geocoder.nominatim import Geocoder
```
shankari: nit: why do you need two imports from the same module?

emission/core/wrapper/suggestion_sys.py
```diff
@@ -309,22 +309,24 @@ def review_start_loc_nominatim(lat, lon):
...
     
     # spec_trip = cleaned_trips.iloc[cleaned_trips[cleaned_trips._id == tripid].index.tolist()[0]]
-    spec_trip = cleaned_trips[cleaned_trips._id == tripid]
-    start_location = spec_trip.start_loc.tolist()[0]
-    end_location = spec_trip.end_loc.tolist()[0]
+    # cleaned_trips[cleaned_trips._id == tripid]
+    # spec_trip.end_loc.tolist()[0]
+    spec_trip = cleaned_trips.get("data")
+    start_location = spec_trip["start_loc"]
+    end_location = spec_trip["end_loc"]
```
shankari: this is an entry object, so you can directly access `cleaned_trips.data.start_loc`

valin1:
ETA latest would be tonight, but I think I can get it done this morning.

emission/core/wrapper/suggestion_sys.py
valin1: I wasn't sure how to use the Geocoder from the import as geo so I used "from emission.net.ext_service.geocoder.nominatim import Geocoder"

shankari: If you already have `import emission.net.ext_service.geocoder.nominatim as geo` then basically `geo` is a shortcut for `emission.net.ext_service.geocoder.nominatim`. so you would do `geo.Geocoder`

I prefer `import ... as` instead of `from .... ` because then if you have multiple imported geocoders (e.g. nominatim and google), you can have two geocoder objects distinguished by their prefix

valin1:
I think all the nits are fixed in the above commit.

shankari:
Did you re-test after fixing the nits?
I am going to file an issue to add some basic unit tests for this...

valin1:
Yes I re-tested on the ipython notebook after and it still works

shankari:
merging with squash so don't be surprised when you try to delete the branch.

PR #624, opened 2019-01-18
Fixed bugs in suggestions

valin1:
There was a bug with some people's trips and suggestions in that the end trip location was either a town, in which I couldn't access using the "city" keyword, or there was only a postcode(zipcode) information. Thus, I had to build a function that converted zipcodes to cities. 

Now (most) people can obtain suggestions, although Trevor's application has a bug in it. 
However, the suggestions are not exactly the most accurate as the suggestion still scopes the area for any serviceable places despite the ending location being a residential area, so I will create another case if the end location is in a residential area or cut down the scope of searching.


emission/net/api/cfc_webapp.py
```diff
@@ -263,6 +263,16 @@ def getSuggestion():
   logging.debug("Output of ret_dir = %s" % ret_dir)
   return ret_dir
 
+@post('/suggestion_sys')
+def getSingleTripSuggestion(tripid):
```
shankari: Does this work? Don't you need a `<tripid>` parameter or read the tripid from the request json?
Can you show me the logs that indicate that this function is being called instead of `def getSuggestion():`?
Isn't the path for both of them the same (`/suggestion_sys`)?

emission/core/wrapper/suggestion_sys.py
```diff
@@ -54,6 +54,14 @@
 SEARCH_LIMIT = yelp_auth['search_limit']
 
 
+ZIPCODE_API_KEY = yelp_auth['zip_code_key']
```
shankari: You need to update the `yelpfusion.json.sample` file to indicate where to get these fields from.
Are you actually using yelp for zipcode -> city mapping?

PR #628, opened 2019-01-24
Minor fixes to the docker setup

shankari:
- Add documentation to the dockerfile
- Fix the imagename in docker-compose
- Clarify the README with the location of the docker files and how to use the
  compose file without swarm


shankari:
@alvinalexander did you test the docker-compose after you changed to the standard image? The standard image name was wrong :(
But everything else was fine. Thanks for submitting #618, it does make it a *lot* easier.

shankari:
This clarifies the questions in #627

PR #630, opened 2019-01-31
Reverted past changes

chendiw:
Reverted the changes as instructed by comments.


bin/debug/intake_single_user.py
```diff
@@ -17,7 +17,7 @@
 
 if __name__ == '__main__':
     np.random.seed(61297777)
-
+
```
shankari: whitespace change, should be reverted

conf/net/api/webserver.conf.sample
```diff
@@ -13,4 +13,5 @@
     "timeout" : "3600",
     "auth": "skip"
   }
+
```
shankari: whitespace change, should be reverted

docker/Dockerfile
```diff
@@ -6,7 +6,7 @@ MAINTAINER K. Shankari (shankari@eecs.berkeley.edu)
 WORKDIR /usr/src/app
 
 # clone from repo
-RUN git clone https://github.com/e-mission/e-mission-server.git .
+RUN git clone https://github.com/alvinalexander/e-mission-server.git .
```
shankari: we want to use the official docker image, this should be reverted

docker/start_script.sh
```diff
@@ -27,3 +26,4 @@ source activate emission
 
 # launch the webapp
 ./e-mission-py.bash emission/net/api/cfc_webapp.py
+
```
shankari: whitespace changes, should be reverted

emission/core/get_database.py
```diff
@@ -21,6 +21,7 @@
 
 print("Connecting to database URL "+url)
 _current_db = MongoClient(url).Stage_database
+#config_file.close()
```
shankari: commented out line is the only change, should be reverted

emission/net/usercache/formatters/android/motion_activity.py
```diff
@@ -22,6 +22,8 @@ def format(entry):
     fc.expand_metadata_times(metadata)
     formatted_entry.metadata = metadata
 
+    logging.info('*** Motion Data write_ts: %d' % metadata.write_ts)
```
shankari: let's not add more log spew. comment this out or remove it.

emission/storage/pipeline_queries.py
```diff
@@ -252,7 +252,7 @@ def get_time_range_for_stage(user_id, stage):
 def get_current_state(user_id, stage):
     curr_state_doc = edb.get_pipeline_state_db().find_one({"user_id": user_id,
                                                             "pipeline_stage": stage.value})
-    # logging.debug("returning curr_state_doc  %s for stage %s " % (curr_state_doc, stage))
+    logging.debug("returning curr_state_doc  %s for stage %s " % (curr_state_doc, stage))
```
shankari: comment out again

bin/debug/intake_single_user.py
chendiw: I'm not sure if "git revert" would work in this case. Since the original commit removed then added a line, reverting that commit would just remove the new added line and add back the old one (which is also a whitespace change).

shankari: But it is not a whitespace change *over the original*. Adding back to old one will keep the original whitespace, so it won't show up as a *change* and won't pollute the PR with a bunch of changes that have no real meaning.

chendiw: Oh okay! I think I get what you mean. Thanks!

shankari:
Merged #644 to handle this

PR #622, opened 2019-01-10
Data Generator and Load Tester

alvinalexander:
I am still working on a read me for the load tester

PR #644, opened 2019-03-27
Data Generator and Load Tester Updated PR

njriasan:
I made most of the changes requested in #622 which usually just meant reverting files.

The one area I am unsure about is  emission/net/ext_service/geocoder/nominatim.py where I was not sure why Alvin made changes. However all his tests pass after I revert it so I reverted it (which maybe I will need to revoke).

I also made 1 additional change due to a bug in two of his tests in emission/tests/simulationTests/test_user.py. I am confident in the server upload test but I am unsure what the states should be for test_create_otp_trip. Since it appears to just be caching I believe home and work should be acceptable.


shankari:
> The one area I am unsure about is emission/net/ext_service/geocoder/nominatim.py where I was not sure why Alvin made changes. However all his tests pass after I revert it so I reverted it (which maybe I will need to revoke).

It looks like he changed it to run scripts from directories other than the main `e-mission-server`. He also enabled the geocode function which was commented out. I am not quite sure why it was commented out but `get_json_geo` seems like a perfectly adequate substitute. I should really just remove the `geocode` and `reverse_geocode` functions.

shankari:
> I am confident in the server upload test but I am unsure what the states should be for test_create_otp_trip. Since it appears to just be caching I believe home and work should be acceptable.

I am not quite sure what the difference is between `curr_coordinate`, `next_coordinate`, `cur_loc` and `next_loc`. It looks like `curr_coordinate`, and `next_coordinate` are supposed to represent (lat, lng) and `curr_loc`, `next_loc` represent configurations with modes between them.

so I think this should really be `test_create_otp_trip(home, work, 'home', 'work')`? but he falls back to the default mode so the test passes and I think I will just merge with this.

emission/storage/timeseries/cache_series.py
```diff
@@ -58,7 +58,7 @@ def insert_entries(uuid, entry_it):
             ts.insert(entry)
             tsdb_count = tsdb_count + 1
         else:
-            ucdb.save(entry)
+            ucdb.insert_one(entry)
```
shankari: can you revert this change, or replace it with `edb.save(ucdb, entry)`? `save` != `insert`, and I want to be careful about making a change to the core functionality in `emission/storage/`.

shankari:
ok to test

emission/storage/timeseries/cache_series.py
njriasan: Ok I reverted it.

shankari:
ok to test

shankari:
failure was just habitica timeout.
Merging...

PR #645, opened 2019-03-29
adding greentrip site

sam11b:
Inside webapp/www/greentrip_site/ you will find the newest version of greentrip's website. The only leftover pages to be added are the contact us page (contingent on us having an email setup) and the join the study page (once we have all the QR codes and apple store setup figured out). All images used were public domain, along with all icons used.


webapp/www/greentrip_site/related_work.html
```diff
@@ -0,0 +1,22 @@
...
+			<div class = "item"> <a href = "related_work.html"> Related Works </a> </div>
+			<div class = "button"> JOIN STUDY </div>
+		</div>
+	<link rel="stylesheet" type="text/css" href="assets/css/index.css">
+	<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=EB+Garamond" />
+	<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.8.1/css/all.css" integrity="sha384-50oBUHEmvpQ+1lW4y57PTFmhCaXp0ML5d60M1M7uH2+nqUivzIebhndOJK28anvf" crossorigin="anonymous">
+	</head>
+	<div class = "big_box" id = "related_work">
+		<div class = "large_text" id = "large_text_1"> Related Work </div> 
+		<div class = "smaller_text"> The previous group working with the emission study used polar bears as a representation of the emotional connectivity to reducing our environmental impact and used a simple suggestion system as a representation of informative connectivity. The suggestion system used for Greentrip was largely based on the previous group's suggestion system model.
```
shankari: @sam11b aren't  you going to add a link to their tech report?
I thought I had a link on the google doc, but here it is again.
https://www2.eecs.berkeley.edu/Pubs/TechRpts/2019/EECS-2019-2.html

webapp/www/greentrip_site/learn_more.html
```diff
@@ -0,0 +1,25 @@
...
+			<div class = "item"> <a href = "related_work.html"> Related Works </a> </div>
+			<div class = "button"> JOIN STUDY </div>
+		</div>
+	<link rel="stylesheet" type="text/css" href="assets/css/index.css">
+	<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=EB+Garamond">
+	<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.8.1/css/all.css" integrity="sha384-50oBUHEmvpQ+1lW4y57PTFmhCaXp0ML5d60M1M7uH2+nqUivzIebhndOJK28anvf" crossorigin="anonymous">
+	</head>
+	<div class = "big_box" id = "second_page"> 
+		<div class = "large_text" id = "large_text_1"> Learn More </div>
+		<div class = "smaller_text"> <i class="fas fa-leaf"></i> Greentrip is a study examining how sustainable suggestions affect travel behavior.  </div>
```
shankari: I wonder if we should say this up front. Because if you are part of the control group, none of this is going to apply to you, and then you will know that you are in the control group. How do you avoid that?

shankari:
@sam11b all your files for this change are in `webapp/www/greentrip_site/` but `cfc_webapp` serves up files from `webapp/www`. Is there a reason you didn't just overwrite the existing website?

shankari:
@sam11b did you test this through the regular e-mission server? because I don't see how it could have worked. Not only is the directory that it expects for static files wrong, but also, e-mission is very picky about which files it returns. A common attack vector for hackers is to read configuration files that are stored in the web directory and get access to passwords, api keys, etc.

so the e-mission webapp only returns certain kinds of files, and only returns them from certain specified directories.
https://github.com/e-mission/e-mission-server/blob/greentrip/emission/net/api/cfc_webapp.py#L112

and the directory structure that it expects is not the one that you have.

You have to either change your directory structure or change `cfc_webapp.py

PR #646, opened 2019-03-31
added link to previous work

sam11b:
Added previous group's link to work to the related works page. Very few changes in this pull request.


shankari:
yup, that is a pretty trivial change!

PR #647, opened 2019-04-01
Fixed Zipcode API erroring

valin1:
There was an error because the code wasn't going into the backup API key for zipcode requests.

### DIFF
--- a/emission/core/wrapper/suggestion_sys.py
+++ b/emission/core/wrapper/suggestion_sys.py
@@ -161,8 +161,6 @@ NOMINATIM API: Checks if location given by nominatim call is a residential locat
 def is_service_nominatim(business):
     if "Hall" in business:
         return False
-    if "Road" in business:
-        return False
     return True
 
 '''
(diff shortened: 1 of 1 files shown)


emission/core/wrapper/suggestion_sys.py
```diff
@@ -330,17 +328,21 @@ def review_start_loc_nominatim(lat, lon):
...
-        response = requests.request('GET', url=url)
-        return response.json()
+    
+    # Use this API key first. 
+    url = ZIP_HOST_URL + ZIPCODE_API_KEY + ZIP_FORMAT + zipcode + ZIP_DEGREE
+    response = requests.request('GET', url=url)
+    results = response.json()
 
-    except:
+    if "error_code" in results:
```
shankari: so how did this ever work? or did you never get a chance to test the error condition?

valin1: I checked it on the ipython notebook for Sam and Bill's trips and it works, since before it was going into the automatic error condition.

shankari: I am not asking how you tested it, although that is good to know and I wish you had included it in the original PR description.

What I want to know is why you used a try/catch block in the first place.

valin1: I didn't realize that the error would be written in the json results and thought that it would error while retrieving the results. The code still ran because the json went through since the json itself contained something like {"error_code": 429, "error_message": blah blah}.

shankari: ok so you did not test the error condition prior to this change - you wrote the code based on assumptions of how it would work.

shankari: This is why unit testing is so important!

PR #648, opened 2019-04-04
updated greentrip directory structure and added client setup

sam11b:
Altered directory structure to run on server. Added client setup html with appropriate qr code. Still to do: javascript with 50 control/50 experimental group qr code control. Adaptations for phone (screen size adjustments). Also still need contact us page pending email construction. Fixed merge conflict.


webapp/www/templates/client_setup.html
```diff
@@ -26,20 +26,22 @@ <h2>Thank you for participating - only two steps left!</h2>
                   </g>
               </g>
           </svg>
-        <h4>Install the {{base_app}} app</h4>
+        <h4>Install the e-mission app</h4>
```
shankari: aren't you going to have people install the emTripLog app, not the e-mission app?

webapp/www/templates/client_setup.html
```diff
@@ -26,20 +26,22 @@ <h2>Thank you for participating - only two steps left!</h2>
...
         </div>
         <div class="row responsive-sm">
-          <a class="col" href='{{android_base_app}}'><img style="height:75px;" alt='Get it on Google Play' src='https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png'/></a>
-          <qrcode class="col" aria-label="qrcode for google play installation" data="{{android_base_app}}" size="200"></qrcode>
+          <a class="col" href='https://play.google.com/store/apps/details?id=edu.berkeley.eecs.embase'><img style="height:75px;" alt='Get it on Google Play' src='https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png'/></a>
+          <qrcode class="col" data="https://play.google.com/store/apps/details?id=edu.berkeley.eecs.emission" size="200"></qrcode>
```
shankari: there's an inconsistency there - the link is to emTripLog and the qrcode is to emission

webapp/www/templates/client_setup.html
```diff
@@ -57,14 +59,16 @@ <h4>Install the {{base_app}} app</h4>
           <h4>Join this study</h4>
         </div>
         <div id="downloadRow">
-          <a id="groupLink" href="emission://change_client?new_client={{client_label}}&clear_local_storage={{clear_local_storage_flag}}&clear_usercache={{clear_usercache_flag}}">
+          <a id="groupLink" href="emission://change_client?new_client={{greentrip}}&clear_local_storage={{clear_local_storage_flag}}&clear_usercache={{clear_usercache_flag}}">
```
shankari: I am completely confused now. Aren't you going to randomly assign people to either greentrip or control group?
If you hardcode greentrip here, how is your recruitment supposed to happen?

sam11b:
Added testing for qr code

webapp/www/templates/client_setup.html
sam11b: fixed this

webapp/www/templates/client_setup.html
sam11b: It's hardcoded to greentrip because I don't have the other qr code. The generated qr code through angular was not working so now I'm going to have it randomly choose from the two qr code images to display.

shankari: as I said in slack, you can get the other QR code by using `new_client=urap2017control` instead of `new_client=greentrip`.

shankari: Is this supposed to be resolved? Because I still don't see any kind of randomness here.

sam11b: The code originally without the images of the qr codes presented nothing with no error so there was no way I could debug. It seemed as though the qr code generator was just not working period, which is why I switched to images.

shankari: I have no objection with switching to images, but you need to randomly switch between two images, right?

sam11b:
I can't implement randomness until I get a qr code image for the control group. I cannot generate one myself because the qr code generation built into client_setup is not working.

webapp/www/templates/client_setup.html
shankari: where is the `Math.random` that chooses between the images? which line number?

shankari:
You can generate one from the production e-mission server.
FTFY
 https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=urap2017control&clear_usercache=true&clear_local_storage=true

webapp/www/templates/client_setup.html
sam11b: The qr code generation is not working. Changing this did not help.

shankari:
@sam11b do you want me to merge and deal with the randomization later?

sam11b:
ok. Another pull request will be out before Wednesday with the js.

PR #650, opened 2019-04-09
website debugging, alternating javascript, learn more page changes

sam11b:
Attempting to debug a couple changes with the website but still largely unsure why the Greentrip icon is resulting in an error. Also trying to fix error with learn more button, made a minor change that may fix this. The client_setup page now has a 50% chance to display the experiment group and a 50% chance that it will display the control group. The bottom QR code changes accordingly.

Finally, the learn more page has more limited information so that the user does not know which group they are a part of (ie control vs experimental).


shankari:
> Also trying to fix error with learn more button, made a minor change that may fix this.

Why *may* fix this? can you indicate testing done?

sam11b:
Essentially one of the links to the learn_more page was incorrect, so I changed that. I changed shankari.JPG to shankari.jpg (this was probably not the error but had to try something). I sincerely don't know why the logo isn't working. All links are organized in the same way so it really doesn't make sense why the site results in a 404 error sometimes unless I've made a spelling mistake or entered the wrong pathway (which was the case for learn_more.html on one of the pages). As far as testing for the js, I've been refreshing client_setup and watching the links/qr codes change (approx 50% on both). If you'd like I can set up a video of this as well.

shankari:
> As far as testing for the js, I've been refreshing client_setup and watching the links/qr codes change (approx 50% on both).

That is testing the client setup. how are you testing the rest of the website? Why don't you know whether the jpg -> JPG change fixed the broken image?

sam11b:
Locally, the image shows up on both Safari and Chrome. However, it was also displaying locally before this change. I'm just trying out different changes to see if that is the reason the server is buggy/resulting in 404 errors. I'm unsure how to test this without pushing to the server other than just looking at it locally?

shankari:
define "looking at it locally". I think that is the problem.

sam11b:
A 404 error really doesn't tell us anything about the errors. Additionally, when the server was updated before there needed to be some changes to conf files which could also be resulting in the errors? I did not make these changes last time.

sam11b:
By locally, I mean testing it straight from the directory (ie file:///Users/Samantha/Desktop/e-mission-server/webapp/www/index.html). I'm unsure how else to look at the website.

sam11b:
I'm aware this isn't the best way to test but I don't understand how else I can test it?

shankari:
> By locally, I mean testing it straight from the directory (ie file:///Users/Samantha/Desktop/e-mission-server/webapp/www/index.html)

You start a local copy of the server and look at localhost:8080
https://github.com/e-mission/e-mission-docs/blob/master/docs/e-mission-server/manual_install.md#run

sam11b:
O lovely. Totally forgot we could do this. Ok so now with my local server I can see your image and the learn_more page seems to be working. Greentrip logo still results in an error and I'm unsure as to why.

shankari:
you can add logging to `cfc_webapp.py` to see which file you are getting a 404 for and how it is being mapped in the server

sam11b:
I'm fairly certain this has something to do with the conf changes I didn't make before. I tried redirecting  the greentrip logo to a copy of the index I pasted into the templates folder called "index_copy.html" and the 404 error still references "index.html" meaning it's not taking my changes.

shankari:
Sometime, you need to force refresh

sam11b:
This worked! Just needed to refresh. I redirected the greentrip logo to a copy of index.html and this seems to be working.

shankari:
Maybe remove the old index.html to ensure that you don't get confused (which you will). Given your path in the fix, it looks like just using `index.html` instead of `index_copy.html` should work.

sam11b:
I was keeping it there because cfc_webapp.py uses index.html to open up the website. I don't want to change cfc_webapp.py because I didn't write it. I think it should be ok since index_copy.html is within the templates folder? I could also rename index_copy to something else if you think that would help/reduce confusion.

shankari:
as I said, 

> it looks like just using index.html instead of index_copy.html should also work. 

since the essence of your change was removing the relative path (`../`) before `index.html`

PR #651, opened 2019-04-16
Client Setup matches look of rest of site, contact us page complete, greentrip logo refresh fixed, switched to greentripcontrol

sam11b:
Client setup page now matches the rest of the website and the control link has been changed to greentripcontrol (as opposed to urap2017control). Contact us page has been added with our new email. Greentrip logo refresh has been tested and updated. 

All excess commits can be squashed. After this pull request is confirmed I will refork and restart so that future updates will not be this messy. 

Most changes were made and tested before the server went down.


shankari:
Given the number of commits in this PR and the number of files changed, it is very hard to review.
It is not sufficient to "squash" the commits since the files are actually changing.

It looks like most of the commits are updates to master that were not generated by you. I would recommend first merging from master -> greentrip and then generating your PR so that it only includes your changes.

PR #655, opened 2019-04-18
Added parameters to make suggestions less wordy

valin1:
On the UI side, suggestions should look less wordy if outputted in this sort of format:
```
How about this location? 
(Business Name) 
We saw that you went to (previous end location) through vehicle.
Instead there is (Business Names) that has better reviews and closer to you. 
Rating of (rating) 
Mode of transportation that you could take: (mode of transportation)
```

Logs: 
{'businessid': 'mr-mamas-las-vegas-7',
 'message': 'We saw that you took a vehicle from 89135, United States of America to 89130, United States of America',
 'method': 'public',
 'question': 'How about this location?',
 'rating': '4.5',
 'suggested_loc': 'Instead, there is Mr Mamas, which has better reviews and closer to your original starting point',
 'tripid': ObjectId('5c3ead4669afd753a2b9a5ad')}


shankari:
@valin1 why did you submit the survey specs as well? Was that intentional? And why are they called `....sample` when they are not samples, they are real surveys?

valin1:
Sorry, that was not intentional. I can remove it from this pull request.

shankari:
LGTM! Merging with squash because of the corrections.

PR #656, opened 2019-04-18
Update to Client Setup Page on Website for GreenTrip Study

b-i-l-l-c-a-o:
This is a pull request to change the website for the study, specifically client setup. It simply adds one image to /img and also updates the html for client_setup. The behavior is as expected and deviates between control/treatment. Tested on iPhone on safari + chome and PC on chrome + edge.

Mobile test
mobile test

PC test
pc test


shankari:
LGTM! At least a lot better than what we have now.
@b-i-l-l-c-a-o can you get more feedback from your friends as well before tonight?

PR #658, opened 2019-04-19
Fix numpy encodings while returning metrics results

shankari:
+ bonus fix to log nominatim queries for better debugging


shankari:
jenkins, test this please

shankari:
jenkins, test this please

PR #659, opened 2019-04-19
Making sure that the sample files accurately reflect all the keys that are needed

shankari:
@valin1 can you please review; I think that this is a pretty simple change.


shankari:
I will merge in ~20 mins because this is really a trivial change.

valin1:
Yes, looks good to me.

PR #672, opened 2019-04-25
Make it easier to test individual candidate functions

shankari:
Enhance the test harness to compare the performance of various implementations
of the same algorithm. In particular:
- add a list of candidate implementations for each step in the pipeline
- allow users to choose which candidates they want to test
- default to testing all candidates if the user does not specify anything

We also keep track of the results for each step and print out a comparison at the end

Also, refactor the existing suggestion system to allow us to compare the
various choices (e.g. nominatim v/s google v/s yelp?) for reverse geocoding and
search by name v/s search by address for category lookup v/s something even
more awesome?

Testing done:

With no categories specified (which also ensures that the refactoring of the
suggestion system did not did not have a regression by testing
`category_of_business_nominatim`, which returned the same result as last night)

```
./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.py category_of_business_nominatim
INFO:root:candidate: category_of_business_nominatim, accuracy = 4.0
INFO:root:candidate: category_from_name_wrapper, accuracy = 6.0
INFO:root:candidate: category_from_address_wrapper, accuracy = 4.0
INFO:root:candidate: category_of_business_awesome, accuracy = 6.0
```

With categories specified

```
./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.py category_of_business_nominatim --candidates category_of_business_awesome
INFO:root:candidate: category_of_business_awesome, accuracy = 6.0
```

### DIFF
--- a/emission/core/wrapper/suggestion_sys.py
+++ b/emission/core/wrapper/suggestion_sys.py
@@ -175,18 +175,33 @@ Then the function will enter the Google reverse lookup and choose a business tha
 latitude and longitude given
 '''
 
+### BEGIN: Pulled out candidate functions so that we can evaluate individual accuracies
+def find_destination_business_google(lat, lon):
+    return return_address_from_google_nomfile(lat, lon)
+
+def find_destination_business_yelp(lat, lon):
+    return (None, None, None, None)
+
+def find_destination_business_nominatim(lat, lon):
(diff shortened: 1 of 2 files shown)


valin1:
Looks good to me

PR #676, opened 2019-04-26
Debugged the nominatim function, so it runs

valin1:
This did not improve accuracy, but I found some tests that technically should've passed in terms of finding the correct business name, but didn't match the street address exactly. Because nominatim's street address includes the business name as well in the street address. --> Will need to address the expected results in the tests. 

Verdict: I think we should update our server, because when I tried changing the link to the public server like https://nominatim.openstreetmap.org/reverse? and the MapQuest version with our key http://open.mapquestapi.com/nominatim/v1/reverse.php?key=[our key], I got this error. 

```
Traceback (most recent call last):
  File "/Users/vanessalin/Desktop/e-mission-server/emission/net/ext_service/geocoder/nominatim.py", line 95, in reverse_geocode
    jsn = cls.get_json_reverse(lat, lng)
  File "/Users/vanessalin/Desktop/e-mission-server/emission/net/ext_service/geocoder/nominatim.py", line 87, in get_json_reverse
    response = urllib.request.urlopen(request)
  File "/Users/vanessalin/anaconda2/envs/emission/lib/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/Users/vanessalin/anaconda2/envs/emission/lib/python3.6/urllib/request.py", line 532, in open
    response = meth(req, response)
  File "/Users/vanessalin/anaconda2/envs/emission/lib/python3.6/urllib/request.py", line 642, in http_response
    'http', request, response, code, msg, hdrs)
  File "/Users/vanessalin/anaconda2/envs/emission/lib/python3.6/urllib/request.py", line 570, in error
    return self._call_chain(*args)
  File "/Users/vanessalin/anaconda2/envs/emission/lib/python3.6/urllib/request.py", line 504, in _call_chain
    result = func(*args)
  File "/Users/vanessalin/anaconda2/envs/emission/lib/python3.6/urllib/request.py", line 650, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 403: Forbidden
```

Example of false tests: 
```
DEBUG:root:Test Mission San Jose High School Fremont failed, output = ['Mission San Jose High School', 'Mission San Jose High School, Mission Creek Trail, Mission San Jose District, Fremont, Alameda County, 94539, United States of America', 'Fremont', False], expected ['Mission San Jose High School', 'Palm Avenue', 'Fremont', True]
```
```
[...]

PR #677, opened 2019-04-26
Check for the name in `find_destination_business` + add noise + run on test at a time

shankari:
- Change the `find_destination_business` harness to check for the name only
instead of all matches. See https://github.com/e-mission/e-mission-server/pull/676 for the request.
See below for more details on the change. (https://github.com/e-mission/e-mission-server/pull/677/commits/1b2ebc0bda5beaab5df229f589bdb4acbce1e4e5)
- Support running one test at a time with the `-t` option to make it easier to debug errors in the lookup 
(https://github.com/e-mission/e-mission-server/pull/677/commits/c9a1ed5ea4aa461e142b66139db2a1d8386c27d5)
- Support adding noise, and specifying a seed to reproduce given the new source of randomness (https://github.com/e-mission/e-mission-server/pull/677/commits/1ed6ba0e2df3306b289e8c76e1df4ecd9b45734d)

### DIFF
--- a/emission/integrationTests/suggestionsys/algorithm_test_harness.py
+++ b/emission/integrationTests/suggestionsys/algorithm_test_harness.py
@@ -4,11 +4,44 @@ import logging
 logging.basicConfig(level=logging.DEBUG)
 import argparse
 import json
+import random
+import geojson as gj
+import math
 
 import emission.core.wrapper.suggestion_sys as sugg
 
-def test_find_destination_business(cfn, params, exp_output):
-    lat, lon = sugg.geojson_to_lat_lon_separated(params["loc"])
+### Add noise from a uniform distribution with range `noise_in_meters`
+def add_noise(loc_geojson, noise_in_meters):
+    # Formula from
+    # http://www.movable-type.co.uk/scripts/latlong.html
+    # section "Destination point given distance and bearing from start point"
+    R = 6371e3 # radius of the earth in meters from the same location
+
+    # if we don't have this check, then the subsequent randrange call fails
+    # with the error
+    # ValueError: empty range for randrange()
+    if noise_in_meters == 0:
+        logging.info("noise_in_meters = 0, skipping add_noise")
+        return loc_geojson
+    delta_dist = float(random.randrange(noise_in_meters))
+    brng = math.radians(random.randrange(360))
+
+    old_lat = math.radians(loc_geojson["coordinates"][1])
+    new_lat = math.asin(math.sin(old_lat)*math.cos(delta_dist/R) +
+                    math.cos(old_lat)*math.sin(delta_dist/R)*math.cos(brng) );
+    old_lng = math.radians(loc_geojson["coordinates"][0])
+    new_lng = old_lng + math.atan2(
+        math.sin(brng)*math.sin(delta_dist/R)*math.cos(old_lat),
+        math.cos(delta_dist/R)-math.sin(old_lat)*math.sin(old_lat));
+    new_loc = gj.Point((math.degrees(new_lng), math.degrees(new_lat)))
+    logging.info("After adding noise, converted %s -> %s" %
+        (loc_geojson["coordinates"], new_loc["coordinates"]))
+    return new_loc
+
+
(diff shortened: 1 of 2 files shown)


shankari:
>  the accuracy is 22%

Note that this accuracy is really too high because in some cases the street name being returned as the business name actually matches the street name in the previous result format - e.g.

```
DEBUG:root:Nominatim found destination business ('California Circle', 'California Circle, Milpitas, Santa Clara County, 95035, United States of America', 'Milpitas', False)
DEBUG:root:found match! name = California Circle, comparing with ['Chevron Milpitas', 'California Circle', 'Milpitas', True]
```

We really need to fix the datasets first to remove the addresses and then re-run the test.

shankari:
The expected result is still a list because different sources (e.g.
nominatim/yelp/google) may have different ways to format the same name. The
test will now pass if any of the result matches any of the entries in the
expected output.

Since this is a more complex check, had to move the checking logic into the
wrapper functions for consistency.

Also changed the expected result of the sample entry "Mountain View Kohl's" to
illustrate the change. I have NOT changed any of the other entries - if people
find alternate spellings for them while evaluating google/yelp, they should
change the datasets as part of their commit.

With this fix, but without
https://github.com/e-mission/e-mission-server/pull/676, and using our nominatim server,
the accuracy is 22%

```
./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.py find_destination_business --candidate find_destination_business
...
DEBUG:root:For candidate find_destination_business, after instance Department of Motor Vehicles, successfulTests = 7, failedTests = 24
INFO:root:Testing candidate find_destination_business complete, overall accuracy = 22.580645161290324
INFO:root:Test complete, comparison results =
INFO:root:candidate: find_destination_business, accuracy = 22.580645161290324
```

shankari:
Support running a single test at a time

As people debug the results from the various business lookup options, it may be easier to debug the tests one by one instead of running all of them at once.

Support running one test at a time with the `-t` option

shankari:
Support the ability to add noise to the locations

This supports more sophisticated testing since the raw data is not guaranteed to be at the exact lat/lng of the business. On android, we detect a trip end if the user has not moved more than 100 mts in 10 minutes, on iOS there is a proprietary algorithm that we have no control over. So the destination lat/lng can be noisy.

Once we have figured out the non-noisy mappings, we should test with noisy inputs as well

Since we now have randomness, also introduce the ability to specify a random seed for reproducibility

Testing done:

With noise, we get different (false) results every time

    ```
    ./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.p
y find_destination_business --candidate find_destination_business --test "Mountain View
Kohls" --noise 50
    DEBUG:root:no match! name = California Street, comparing with ["Kohl's", "Kohl's Mou
ntain View"]

    ./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.p
y find_destination_business --candidate find_destination_business --test "Mountain View
Kohls" --noise 50
    DEBUG:root:no match! name = Pacchetti Way, comparing with ["Kohl's", "Kohl's Mountai
n View"]

    ./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.p
y find_destination_business --candidate find_destination_business --test "Mountain View
Kohls" --noise 50
    DEBUG:root:Nominatim found destination business ('T-Mobile', 'T-Mobile, Pacchetti Wa
y, Castro City, Mountain View, Santa Clara County, 94023, United States of America', 'Mo
untain View', False)
    DEBUG:root:no match! name = T-Mobile, comparing with ["Kohl's", "Kohl's Mountain View"]
    ```

With noise and seed, we get the same (false) result every time

    ```
    $ ./e-mission-py.bash emission/integrationTests/suggestionsys/algorithm_test_harness.py find_destination_business --candidate find_destination_business --test "Mountain View Kohls" --noise 50 --seed pleasework
    INFO:root:Configuring random with seed pleasework
    DEBUG:root:no match! name = California Street, comparing with ["Kohl's", "Kohl's Mountain View"]
    ....
    ```

valin1:
Looks good.

PR #681, opened 2019-05-01
Gt server5

trevor-wu:
INFO:root:candidate: find_destination_business_yelp, accuracy = 71.42857142857143
INFO:root:candidate: find_destination_business_google, accuracy = 0.0
INFO:root:candidate: find_destination_business, accuracy = 14.285714285714286

Yelp is very accurate with restaurants, much more so than the other APIs, but in regular testing , find_destination_business works better.

Yelp does not work well with other establishments because the search function in the api has a hard time querying non restaurant businesses. Since large businesses such as Walmart of Costco have very large footprints, reducing the search radius would return an empty query.


shankari:
@trevor-wu please revert the whitespace-only changes because they make the change hard to review.
Also, which dataset are these are results for?
Remember, this is a report for me to review, not an autograder submission!

shankari:
also, which other parameters from the yelp search API did you experiment with? I don't see any results for that, in spite of our discussion yesterday.

trevor-wu:
Screen Shot 2019-05-05 at 2 54 16 PM

These are the coordinates inside costco. I put them into PostMan and ran it through the YelpAPI. It does not show any costco-related establishments like costco or costco gas.
These also show the parameters I experimented with: lat/lon, search radius, and sort by distance, to make sure that I can get the closest business with the lat/lon coordinates.
However, if I increase the radius of search, businesses do pop up, just not anything costco related.

Screen Shot 2019-05-05 at 2 54 24 PM

However, Yelp is very good at indexing restaurants, with a 70% accuracy rate among the dataset entries which are eating establishments. 

Other options given by the Yelp API are match - which doesn't work because it basically takes a name and address and matches it to a business id, it would be unhelpful since here we are just trying to identify the business.
There's also phone search, which matches a business with its phone number.

trevor-wu:
My solution was to use the original algorithm when Yelp fails to index any businesses. This improves accuracy because since the search radius is small, when Yelp querys a business, it is most likely the business that the user visited. However, since Yelp's search function has trouble querying certain types of businesses, it will return nothing when someone visits those types of business. Since we know that Yelp will query nothing at businesses like Costco, it would be wise to let the original API try to identify it since it at least has a chance.

shankari:
@trevor-wu I understand that you experimented with 4 parameters.

> also, which **other** parameters from the yelp search API did you experiment with?

Also

> please revert the whitespace-only changes because they make the change hard to review.

Please actually address my comments - you already had results with 4 parameters 5 days ago, and I was looking for something more.

trevor-wu:
Possible Parameters for Search
term - gives a query based on search term - not useful because we don't have any other data besides location
location string - only useful if we have a street address
latitude - used
longitude - used
radius - used
categories - not useful because we don't know the categories the business belongs in if we don't know the business
locale - not useful because https://www.yelp.com/developers/documentation/v3/supported_locales locale is too broad of a term
limit - used, limit = 3
offset - offsets list (ex. if offset=1, then list will contain 2nd result onward), tested in postman, will only make results worse as current results are sorted by distance -> will return a business further away.
sort_by - used, distance would be most relevant, sort_by best doesn't always return best result in this case
price - we don't know the price of the business we just visited
open_now - would limit options, and not useful if we don't know the business. Can be useful for suggestions since we wouldn't want to route someone to somewhere closed
open_at - same as above
attributes - not useful if we don't know the business, can be useful when recommending if we want to find a similar business.

MATCH
name - required, not useful since we don't know the name

trevor-wu:
I have experimented most with radius, since some areas are sparser. 

Github said it hid the whitespace changes

shankari:
It's true that you don't know the category, but you can try multiple
categories. did you see what Sam did with google? Also, read the
description of the term to understand the current behavior

trevor-wu:
I just tested hard-coding categories, and it does allow costco to appear if I hard code in the category "shopping".
Screen Shot 2019-05-05 at 5 54 18 PM
However, after hardcoding categories into suggestion_sys.py, I hardcoded : 
```
url_params = {
        'latitude': lat,
        'longitude': lon,
        'radius' : radius,
        'limit': SEARCH_LIMIT,
        'sort_by': 'distance',
        'categories': 'shopping, restaurants, auto, beautysvc, food, professional, education, arts, active'
    }
```
The accuracy of the algorithm decreased to 19.5122, whereas it is 26.829268292682926
without the categories hardcoded.

shankari:
> The accuracy of the algorithm decreased to 19.5122, whereas it is 26.829268292682926
without the categories hardcoded.

That seems like a counter intuitive result. Did you try investigating why entry by entry as @sam11b did and I've asked everybody to do

shankari:
superceded by #685

PR #683, opened 2019-05-06
added new qr control code

sam11b:
Updated qr control code to our studies qr control group and not last semester's. Apologies this was incorrect before.


shankari:
@sam11b can you add the testing done?

sam11b:
Would you like this in the form of a video link like last time?

shankari:
yeah video link would be good

sam11b:
https://youtu.be/PoO8kOQJWM4 

I've cropped out my email password but as you can see I was given the control QR code and it was generated correctly.

PR #684, opened 2019-05-06
PR For Updated Dataset

b-i-l-l-c-a-o:
Successful tests is 2, failed is 22 from me running it


emission/core/wrapper/suggestion_sys.py
```diff
@@ -6,7 +6,9 @@
 import json
 import logging
 import re
-import emission.storage.timeseries.abstract_timeseries as esta
+import sys
+
+#import emission.storage.timeseries.abstract_timeseries as esta
```
shankari: > NOTE #2: If you comment out the line
"import emission.storage.timeseries.abstract_timeseries as esta"
in suggestion_sys, you don't need to have the database running. Please make
sure not to commit this change

shankari:
Squashing before merge because the commit history is a mess.

PR #690, opened 2019-05-10
Major refactoring of the giant suggestion system function to make it easier to support bid

shankari:
Most of this is just moving things around as opposed to changing the functionality. Actual functionality changes to support the new bid-specific method will be in the new PR to reduce reviewer burden for this admittedly already complex PR.

Functionality changes are called out in the commits, repeating them here for reference
- change the distance function to take lat/lng instead of address, since we don't look up by address any more (https://github.com/e-mission/e-mission-server/commit/abc777b8220e048e16047b1f652c40a76b29ef43)
- switch find_destination_business to the yelp version so that the suggestion generation is actually exercised (https://github.com/e-mission/e-mission-server/commit/685e3fd7fa9c244189b6f3b0983497c5d7c62e6b)

Testing done:
- basic syntax check, running through without matching destination business (https://github.com/e-mission/e-mission-server/commit/abc777b8220e048e16047b1f652c40a76b29ef43)
- executing the suggestion code and finding an alternate suggestion (https://github.com/e-mission/e-mission-server/commit/685e3fd7fa9c244189b6f3b0983497c5d7c62e6b)

Suggestions for easier review:
- the PR is intentionally split into multiple commits to reduce the burden; you may find it easier to look at individual commits instead of the entire PR
- the most complex of the commits is (obviously) https://github.com/e-mission/e-mission-server/commit/abc777b8220e048e16047b1f652c40a76b29ef43 It may also help to review the high level functional split in the commit and then just look at the final result (https://github.com/shankari/e-mission-server/blob/have_sugg_system_use_businessid/emission/core/wrapper/suggestion_sys.py)

Sorry this is so complex, but I tried to minimize the functionality to make it easier to review. And once this is done, other changes will be much easier.


shankari:
@valin1 second PR to review. This one is likely to be more complex; I have some reviewing suggestions in the description. I think probably one more PR to get the basic functionality in place.

valin1:
Is category of business not needed anymore?

shankari:
I don't think so; I put my reasons in the commit (https://github.com/e-mission/e-mission-server/pull/690/commits/d0110c1a7e3810c468ec27de8fb937041d833a56) dunno if you agree.
Again, I strongly recommend looking at the individual commits and their commit messages. My commit messages are fairly descriptive.

valin1:
Everything looks good to me and I see the reasoning why categories of business is not needed.

shankari:
we can always restore it if it turns out that we need it; it is in the git history.
Merging now...

PR #696, opened 2019-05-11
Very minor change 

shankari:
- Return an Entry from the newly added method (added in
  5ef5868) to be consistent with the existing
  timeseries based method

Testing done:
- Updated the two new tests in TestTripQueries. Both pass.


shankari:
this is just habitica again. And it passed the tests the first time around. Merging...

PR #697, opened 2019-05-13
Switch the suggestion system API methods to use the stored business id if it exists

shankari:
Both the single trip and the general API now look up the `destination_confirm` object, and only if it exists, return a suggestion.

This required a couple of other changes:
- minor change: renamed `cleaned_trip` to `cleaned_trips` in `calculate_yelp_server_suggestion_singletrip_nominatim` since we have retrieved one trip, not a list of trips. Meaningful variable names are important!
- more substantial change: changed the data access method for the list of cleaned trips from `get_data_df` to `find_entries`. Rationale:
- we weren't using the dataframe functionality, and were just iterating through the list of trips anyway
- the function to lookup the `destination_confirm` object requires a trip entry. I could look up the trip object using the id from the dataframe, but that is an additional lookup, and one that is unnecessary if we change the retrieval method

because of this change, we read the fields of the trip object differently, but the changes should be pretty straightforward; essentially find/replace

Testing done:
Added new integration tests
All 5 tests pass

@valin @sam11b, please check the expected behavior of the suggestion function if it doesn't find any trips with valid suggestions. What does the UX expect?

### DIFF
--- a/emission/core/wrapper/suggestion_sys.py
+++ b/emission/core/wrapper/suggestion_sys.py
@@ -8,11 +8,13 @@ import logging
 import re
 
 import emission.storage.timeseries.abstract_timeseries as esta
+import emission.storage.decorations.trip_queries as esdt
 import argparse
 import pprint
 import requests
 import os
 import emission.net.ext_service.geocoder.nominatim as geo
+import emission.core.wrapper.entry as ecwe
 import bson
 
 # Yelp Fusion no longer uses OAuth as of December 7, 2017.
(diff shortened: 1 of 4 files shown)


shankari:
Testing results for the record

```
./e-mission-py.bash emission/integrationTests/suggestionsys/TestSuggestionSys.py
Ran 5 tests in 18.098s

OK
```

valin1:
Looks good to me.

shankari:
> @valin @sam11b, please check the expected behavior of the suggestion function if it doesn't find any trips with valid suggestions. What does the UX expect?

We currently return None (see last test case). Not sure if that is what the UX expects? I am happy to change this, but I need to know the expected behavior of the interface.

shankari:
Merging this for now so we have at least the basic implementation available.

PR #705, opened 2019-06-16
small change to allow things to work

trevor-wu:
See https://github.com/e-mission/e-mission-phone/pull/582 for more details. This change just lets the suggestion know the picked destination.


shankari:
@trevor-wu this is the wrong way to integrate the server and phone changes.
I have been pretty clear in slack that you need to change the front end to use `destination_confirm` to match the backend and not vice versa.

The destination is not the mode. Variable names need to match the data that they represent to be meaningful.

PR #706, opened 2019-06-17
Added business id suggestion function and bump in accuracy for tests

valin1:
Changed tests to reflect accuracy of suggestion system
- A closer airport shouldn't be suggested b/c people have already bought tickets depending on which airport
- Locations like banks or schools should not be suggested either 

The suggestion system thus is changed to not return any business id if the end location does not make sense to be suggested.


shankari:
@valin1 I am not quite sure what this PR is doing - the summary here and the code changes don't seem to match.

Even at a high level:
- why do you delete the automated test that I wrote (`emission/integrationTests/suggestionsys/TestSuggestionSys.py`)?
- why are there a bunch of changes related to google (e.g. `return_address_from_google_candidates_nomfile_sam`)?
- what is the difference between `calculate_yelp_server_suggestion_for_locations_business_id` and the `calculate_yelp_server_suggestion_for_bid` that I already wrote and you reviewed?
- why do you revert my changes to get the confirmed destination and use it to generate the suggestion? Do you claim that the destination matching code is significantly more accurate now?

I spent significant time refactoring your previous code for readability and re-use and I am really reluctant to merge yet another giant copy-pasted function that appears to revert all my changes without some more justification of why it is necessary.

PR #707, opened 2019-06-20
Made changes in tests for better accuracy using suggestion function 

valin1:
Changed tests to reflect accuracy of suggestion system
- A closer airport shouldn't be suggested b/c people have already bought tickets depending on which airport
- Locations like banks or schools should not be suggested either

Sorry, when I was making my changes, I did not check to fetch or pull changes from the greentrip branch so I had forgotten about the refactored changes. There weren't any differences between the refactored code and my code, so the only changes in this pull request are the added updates to the tests.


emission/integrationTests/suggestionsys/find_candidate_business.dataset.json
```diff
@@ -0,0 +1,435 @@
...
+    {
+        "test_name": "Mountain View Kohls",
+        "test_type": "Suburban big box store in local anchor shopping center ",
+        "input": {
+            "loc": {
+                "coordinates": [-122.1078300, 37.4035500],
+                "type": "Point"
+            }
+        },
+        "output": ["Kohl's Mountain View", "350 Showers Drive", "Mountain View", true]
```
shankari: what is this dataset and how is it different from `find_destination_business.dataset.json`

emission/integrationTests/suggestionsys/algorithm_test_harness.py
```diff
@@ -94,7 +109,7 @@ def test_single_instance(test_fn, cfn, instance, noise_in_meters):
         sugg.find_destination_business
     ],
     "calculate_yelp_server_suggestion_for_locations": [
-        sugg.calculate_yelp_server_suggestion_for_locations
+        sugg.calculate_yelp_server_suggestion_for_bid
```
shankari: this seems wrong. don't you want to compare `calculate_yelp_server_suggestion_for_locations` against `calculate_yelp_server_suggestion_for_bid`

shankari: I don't even see how this works. `calculate_yelp_server_suggestion_for_bid` has different arguments than `calculate_yelp_server_suggestion_for_locations`

valin1:
```
successfulTests = 18, failedTests = 18
INFO:root:Testing candidate calculate_yelp_server_suggestion_for_bid complete, overall accuracy = 50 
INFO:root:Test complete, comparison results = 
INFO:root:candidate: calculate_yelp_server_suggestion_for_bid, accuracy = 50
```

The accuracy ended up becoming 50%, and the "None"s were not detected b/c it was comparing None to string "None"s in the json, so including those results, accuracy was 50%.

emission/integrationTests/suggestionsys/algorithm_test_harness.py
valin1: I used a business id for the inputs for calculate_yelp_server_suggestion_for_bid.

emission/integrationTests/suggestionsys/find_candidate_business.dataset.json
valin1: Sorry, this was part of the version that was not up to date with the recent fetch to the branch. It is deleted now.

emission/integrationTests/suggestionsys/algorithm_test_harness.py
shankari: Ah I see that this change is at https://github.com/e-mission/e-mission-server/pull/707/files#diff-9cc1403419659aaacfffa4bd067a5a34R59

A link would have helped clarify the change. Also part of the confusion is that instead of creating a new `calculate_yelp_server_suggestion_for_bid`, you modified the `test_calculate_yelp_server_suggestion_for_locations` so that the function name is not valid any more.

I referred to this in https://github.com/e-mission/e-mission-server/pull/707#discussion_r295889544
You want to be able to compare different choices, including the one that uses locations only and the one that uses business id only and see which one is better.

shankari:
@valin1 just to clarify, please call `calculate_yelp_server_suggestion_for_bid` from a new set of tests for `test_calculate_yelp_server_suggestion_for_bid` - you are not generating the suggestion from a location and you should reflect that.

Once you do that, I can merge. But it would also be interesting to see why the accuracy given the bid is only 50%. Why aren't we finding all the suggestions? You could try for multiple implementations of the bid -> suggestion algorithm and compare them. Also, you may want to distinguish between precision and recall since users are likely to be more forgiving to not finding a possible suggestion than to finding an incorrect suggestion.

valin1:
@shankari By making a new set of tests, I am not exactly sure by what do you mean from that, do you mean to reflect the None values or to not include the start locations? 

I changed the algorithms_test_harness.py to change the results to string, so the None values can be checked to be matched. 
As for the start locations, I don't think we can get rid of that since we need to use it to calculate the distance between the start location and the end location.

valin1:
```
INFO:root:Testing candidate calculate_yelp_server_suggestion_for_bid complete, overall accuracy = 52.63157894736842 
INFO:root:Test complete, comparison results = 
INFO:root:candidate: calculate_yelp_server_suggestion_for_bid, accuracy = 52.63157894736842
```

shankari:
> @shankari By making a new set of tests, I am not exactly sure by what do you mean from that, do you mean to reflect the None values or to not include the start locations?

@valin1 neither. If you look at the harness structure, there are harnesses or wrappers for different kinds of functions, and each of them tests the various implementations of that kind of function - e.g. `test_find_destination_business` tests functions that take a location and return a bid and the candidate algorithms are using google, yelp, etc https://github.com/e-mission/e-mission-server/blob/greentrip/emission/integrationTests/suggestionsys/algorithm_test_harness.py#L90

Similarly, `test_calculate_yelp_server_suggestion_for_locations` currently expects to test functions that take start and end locations and generate suggestions.

If you want to test functions that take start location and end bid and generate suggestions, you need to make a new wrapper/harness function. You can't just edit `test_calculate_yelp_server_suggestion_for_locations`.

shankari:
You will then have potentially many candidate algorithms for the new wrapper `test_calculate_yelp_server_suggestion_for_bid`, hopefully with steadily increasing accuracy.

emission/core/wrapper/suggestion_sys.py
```diff
@@ -129,7 +129,7 @@ def business_details(api_key, business_id):
 """
 def return_address_from_location_nominatim(lat, lon):
     geocode_obj = geo.Geocoder()
-    return geocode_obj.reverse_geocode(lat, lon)
+    return geocode_obj.reverse_geocode(float(lat), float(lon))
```
shankari: what did you need this for? Was there a change to the `reverse_geocode` API?

emission/core/wrapper/suggestion_sys.py
```diff
@@ -519,14 +519,19 @@ def format_suggestion(start_lat, start_lon, orig_bus_details,
...
         # TODO: Can't we just use the business name here directly instead of an
         # address. Seems like that will be a lot more meaningful to people
         end_string_address = " ".join(orig_bus_details["location"]["display_address"])
+        print(end_string_address)
         alt_bus_details = alt_sugg["bdetails"]
         return {
-            'message': 'We saw that you took a vehicle from '+begin_string_address
-                + ' to '+ end_string_address,
+            'message': 'We saw that you took a vehicle from '+ str(begin_string_address)
+                + ' to '+ str(end_string_address),
```
shankari: ditto

emission/core/wrapper/suggestion_sys.py
valin1: It's needed because it throws a syntax error saying that the lat and lon passed is a string.

emission/core/wrapper/suggestion_sys.py
valin1: Same for here, this returns an address object and throws an error saying that it's unable to concatenate the address object and a string.

emission/core/wrapper/suggestion_sys.py
shankari: ok so I don't know how that ever worked. but I don't have time to dig deeper right now.

shankari:
Squashing before merge since there were so many incremental changes. Remember to use a new branch off master for any further changes!

PR #711, opened 2019-08-05
Add config options to control more asserts

shankari:
While developing the pipeline, I added several asserts to flag "issues that
need to be investigated" while retaining fallbacks so that we could comment out
the assertions if there was no time to investigate.

As other groups start using this, they don't have the energy and skill to
invesigate issues, and I certainly don't have the energy to investigate every
single one of them.

Let's add the ability to ignore these asserts in production, replacing them by
log errors instead.

I have not added config options for the zig-zag detection asserts since we have
not encountered them so far, but everything else should be covered.

Testing done:

without this change, ran into the error reported in
https://github.com/e-mission/e-mission-server/pull/709#issuecomment-518142916

After setting `output.conversion.validityAssertions` to false, I get

```
(emission) C02KT61MFFT0:e-mission-server shankari$ ./e-mission-py.bash bin/debug/intake_single_user.py -e test-subway-tram
Connecting to database URL localhost
google maps key not configured, falling back to nominatim
transit stops query not configured, falling back to default
2019-08-05T14:44:58.147111-07:00**********UUID 0f1e49b9-c928-43d5-a079-70be130d4a94: moving to long term**********
2019-08-05T14:44:58.178816-07:00**********UUID 0f1e49b9-c928-43d5-a079-70be130d4a94: filter accuracy if needed**********
2019-08-05T14:44:58.190401-07:00**********UUID 0f1e49b9-c928-43d5-a079-70be130d4a94: segmenting into trips**********
2019-08-05T14:45:09.418732-07:00**********UUID 0f1e49b9-c928-43d5-a079-70be130d4a94: segmenting into sections**********
/Users/shankari/e-mission/e-mission-server/emission/analysis/intake/segmentation/section_segmentation_methods/flip_flop_detection.py:58: RuntimeWarning: invalid value encountered in double_scalars
  sm.update({"trip_pct": (curr_section_time * 100)/total_trip_time})
2019-08-05T14:45:11.659074-07:00**********UUID 0f1e49b9-c928-43d5-a079-70be130d4a94: smoothing sections**********
/Users/shankari/OSS/anaconda/envs/emission/lib/python3.6/site-packages/pandas/core/frame.py:891: UserWarning: DataFrame columns are not unique, some columns will be omitted.
  "columns will be omitted.", UserWarning)
2019-08-05T14:45:12.473511-07:00**********UUID 0f1e49b9-c928-43d5-a079-70be130d4a94: cleaning and resampling timeline**********
[...]

PR #713, opened 2019-08-05
Add config options to control more asserts

shankari:
This is an analog to https://github.com/e-mission/e-mission-server/pull/711
(https://github.com/e-mission/e-mission-server/pull/711/commits/a2acdfbf9c44953ab5390667572abc52350af68f) but for the files that are specific to the gis pipeline code

While developing the pipeline, I added several asserts to flag "issues that
need to be investigated" while retaining fallbacks so that we could comment out
the assertions if there was no time to investigate.

As other groups start using this, they don't have the energy and skill to
invesigate issues, and I certainly don't have the energy to investigate every
single one of them.

Let's add the ability to ignore these asserts in production, replacing them by
log errors instead.

I have not added config options for the zig-zag detection asserts since we have
not encountered them so far, but everything else should be covered.

### DIFF
--- a/emission/analysis/classification/inference/mode/rule_engine.py
+++ b/emission/analysis/classification/inference/mode/rule_engine.py
@@ -45,8 +45,11 @@ class RuleEngineModeInferencePipeline:
             time_query=timerange)
         if (len(self.toPredictSections) == 0):
             logging.debug("len(toPredictSections) == 0, early return")
-            assert self.last_section_done is None, ("self.last_section_done == %s, expecting None" % \
-                self.last_section_done)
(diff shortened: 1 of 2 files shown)


shankari:
jenkins, test this please

shankari:
@DO3B would you like to review this?

shankari:
jenkins, test this please

shankari:
jenkins, test this please

shankari:
This failure is due to the known issue with the ground truth having changed. Since I am merging to `gis-based-mode-detection` and that is broken there as well, I am merging this anyway.

PR #715, opened 2019-08-14
Ensure that we don't extract the motion activity twice

shankari:
This fixes
https://github.com/e-mission/e-mission-docs/issues/356
More investigation + details in the issue

Bonus fix: allows a timezone to be specified so that we extract data between those time ranges **in that timezone** instead of in UTC.


shankari:
jenkins, test this please

PR #717, opened 2019-08-29
Multiple small fixes reported by fabmob from their deployment

shankari:
Primarily related to i18n and robustness to unusual inputs


shankari:
@PatGendre was this the `UTF-8` fix?

shankari:
jenkins, test this please

PatGendre:
Hi @shankari I checked today with Yann:
`"encoding": "UTF-8"` should apply only when there is a `filename` attribute, otherwise the exception is raised.
So in your commit patch you should add a filename value each time you'va added an `"encoding": "UTF-8"` .
Also, you must add a MISSING COMMA in the JSON file.

PR #726, opened 2020-01-13
Add a script to upload a timeline to a remote server

shankari:
This is the equivalent of `load_timeline_for_day_and_user.py` but for a remote server.

- The timeline can be downloaded from the phone (Profile -> Download data dump).
- The server *must* have "skip" authentication since the script does not support
token retrieval.

The use case is during alpha testing, where a selected set of users are invited
to experiment with a new UI but already have longitudinal data collection on a
long-term server that they don't want to miss.

Instead of installing the app in order to divert data from the longitudinal
data collection to populate the new server, they can just download their data
and upload it to the alpha testing server.

Unsure if this is a super common use case, but I wanted it, so I wrote it.


shankari:
Testing done:
- loaded data to a demo server with skip notification turned on and then logged into the server to verify that the data was loaded correctly

shankari:
@kafitz please vote for this if you want to use it :) Add reactions at the top

kafitz:
Looks like this will be pretty helpful for me after all!

PR #727, opened 2020-03-10
Get the tests to pass again

shankari:
Since we shut down a bunch of the e-mission servers.
At this point, all tests pass

```
----------------------------------------------------------------------
Ran 211 tests in 923.050s

OK
```

PR #728, opened 2020-03-13
Add an intial travis.yml

shankari:
As we can see from the history for https://github.com/e-mission/e-mission-server/pull/727, several of the jenkins builds are failing. And this seems to be due to jenkins issues - setting up conda fails sometimes, and connecting to mongodb fails sometimes. RISE supports suggests that I use a dockerized build instead.

But I also cannot rely on having access to the RISE build resources forever, specially as I transition out of UC Berkeley. Therefore, this might be good motivation to move to an external automated build/test system.

Some options that have been verified by GitHub, along with their evaluation matrix, are shown below:

Option | Cost | Limitations | non-docker support | docker/compose support |
------- | ----- | ----------- | --------------------|--------------------------|
CirrusCI | Free | number of concurrent runs per user |  Unix or Windows VMs, but is not part of the regular examples | all examples | docker has good support, unsure about docker-compose | 
CircleCI | certain number of credits for linux and OSX. Unsure about windows - inconsistent info | docs say that it supports VMs, not a lot of docs though | docker has good support, and so does docker-compose | 
TravisCI | Free | Can't see any | Yes (new VM for every run) | docker and docker-compose support |

Based on this, I think I will go with TravisCI for now. It has support for open source projects without any limitations, and it supports non-dockerized builds, which will help us transition the testing better. It also seems to have a better GitHook integration.

Tried to use TravisCI, but there seems to be an inconsistency in the documentation.

I found this guide on open source projects with Travis, 
https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/#stq=&stp=0which said that:

> We are implementing these changes with GitHub Apps, which provides a tighter integration with GitHub and finer control over repository and subscription ownership and permissions. 

However, when I try to actually sign up using travis-ci.com I get the following auth page, which gives travis the ability to write code to any of my repositories!! In travis-ci.org, I only had to give status write access, not code.

| Travis.com (new) | Travis.org (obsolete) |
|---------------|----------------|
| Screen Shot 2020-03-12 at 9 30 51 PM | Screen Shot 2020-03-12 at 9 37 44 PM |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 # Template from 
 # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/use-conda-with-travis-ci.html
-language: python
+language: generic
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
   - "3.6.1"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+# Template from 
+# https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/use-conda-with-travis-ci.html
+language: python
+python:
+  # We don't actually use the Travis Python, but this keeps it organized.
+  - "2.7"
+  - "3.5"
+  - "3.6"
+install:
+  - sudo apt-get update
+  # We do this conditionally because it saves us some downloading if the
+  # version is the same.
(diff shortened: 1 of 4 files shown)


shankari:
Screen Shot 2020-03-17 at 9 25 26 PM

PR #731, opened 2020-03-18
Enable docker-enabled testing

shankari:
And test it using Github Actions.
This should also make it easier to switch to other CI systems in the future


shankari:
Aha! Adding the test actions through the github UI to master and then pulling into this seems to have fixed it. I don't know why it wasn't triggering before. Permissions, maybe?

At any rate, it is running now, and assuming it passes, I can:
- remove the default action
- merge this, and move ahead with moving out the simulator into its own container

PR #744, opened 2020-05-06
Remove unused depends

shankari:
- remove unnecessary dependencies
- remove localdate from storage code and fix all references


shankari:
subsumed by https://github.com/e-mission/e-mission-server/pull/745 and https://github.com/e-mission/e-mission-server/pull/746

PR #749, opened 2020-06-04
fix empty function

nlehuby:
Running `./e-mission-py.bash emission/pipeline/model_stage.py` was outputing:

```
  File "emission/pipeline/model_stage.py", line 43
    def run_model_pipeline(process_number, uuid_list):
      ^
IndentationError: expected an indented block
```

### DIFF
--- a/emission/pipeline/model_stage.py
+++ b/emission/pipeline/model_stage.py
@@ -38,6 +38,7 @@ How to know if the section has never been predicted?
 
 '''
 def setup(self):
+	pass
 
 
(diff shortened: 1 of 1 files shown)


shankari:
If I don't get back to using the common trip model, I should just remove this. But I might get back to it soon as part of the Agile Urban Planning stuff. So I will leave this in for now.

@nlehuby thanks for the fix!

shankari:
Just for the record, `model_stage.py` is not currently actively maintained. `intake.py` is the main pipeline that generates the travel diary.

PR #750, opened 2020-06-11
convert to geojson with tracejson

nlehuby:
part of https://github.com/e-mission/e-mission-docs/issues/542

There are some fixtures that hold exported data (`*.ground_truth` files in `emissions/tests/data`) that could be updated accordingly. But as there are no test about these properties, I'm not sure how valuable it is.


shankari:
hm.. it is true that I don't actually check for the timestamps. Let me see whether the UI uses them. If so, I should add that to the tests and we should ensure that the `.ground_truth` files should also be updated accordingly.

shankari:
Unfortunately, we do include `times` in the phone code both in the detail screen and during the incident reporting. I will wait until I fix the phone code to merge this.

shankari:
@nlehuby thanks for fixing the `ground_truth` files as well. I will merge this after I fix the UI as part of https://github.com/e-mission/e-mission-docs/issues/542

PR #757, opened 2020-08-03
Handle the motion activity entries properly in the formatter

shankari:
This corresponds to
https://github.com/e-mission/e-mission-data-collection/pull/186/commits/944c091ecab7b91740d80c3693d48731b6581bac

- Change the fallback `if` statements to only be executed if the field actually exists
    Since the new format directly has `type` and `confidence`, they should
        represent the fallback for the `if` statement. And in that case, we
        shouldn't do anything.

- Similarly, only overwrite the `ts` if it doesn't already exist

### DIFF
--- a/emission/net/usercache/formatters/android/motion_activity.py
+++ b/emission/net/usercache/formatters/android/motion_activity.py
@@ -25,6 +25,7 @@ def format(entry):
     #logging.info('*** Motion Data write_ts: %d' % metadata.write_ts)
     
     data = ad.AttrDict()
+
     if 'agb' in entry.data:
         data.type = ecwa.MotionTypes(entry.data.agb).value
     elif 'zzaEg' in entry.data:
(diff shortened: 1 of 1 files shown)


shankari:
Testing done:
- Create motion activity entries in the emulator
- Send them to a server with the fix

No errors while processing the data

in the logs:
```
2020-08-03 11:04:25,637:DEBUG:4816780736:write_ts = 1596476564.192
2020-08-03 11:04:25,639:DEBUG:4816780736:module_name = emission.net.usercache.formatters.android.motion_activity
2020-08-03 11:04:25,639:DEBUG:4816780736:write_ts = 1596476567.432
2020-08-03 11:04:25,640:DEBUG:4816780736:module_name = emission.net.usercache.formatters.android.motion_activity
```

in the database:

```
>>> edb.get_timeseries_error_db().find().count()
0
```

PR #758, opened 2020-08-03
Pull from master before CEO project

shankari:


shankari:
Closing this so that I can first pull to the GIS branch

PR #761, opened 2020-08-12
Upgrade all dependencies to the latest versions

shankari:
- Upgraded to the most recent version of everything, with the following caveats:
    - miniconda is 4.8.3 instead of 4.8.4 since there doesn't appear to be a
      4.8.4 labelled binary yet and I don't want to use `latest`
    - remove versioned python since it is installed automatically
    - specifying python also causes major dependency incompatibilities, since
      all packages have not been upgraded to the latest version
- Remove the manual installs since we are upgrading to the most recent version

Testing done:
- Setup works locally
- Running tests on CI; might run locally with persistent failures

### DIFF
--- a/setup/environment36.notebook.additions.yml
+++ b/setup/environment36.notebook.additions.yml
@@ -3,7 +3,7 @@ channels:
 - conda-forge
 - defaults
 dependencies:
-- branca=0.2.0=py_1
-- folium=0.5.0=py_0
-- ipython=5.3.0=py36_0
-- jupyter=1.0.0=py36_3
+- branca=0.4.1
+- folium=0.11.0
+- ipython=7.17.0
+- jupyter=1.0.0
--- a/setup/environment36.yml
+++ b/setup/environment36.yml
@@ -3,24 +3,23 @@ channels:
 - conda-forge
 - defaults
 dependencies:
-- arrow=0.12.1
-- attrdict=2.0.0
-- cheroot=6.2.*=py36_0
-- future=0.16.0
-- google-auth=1.2.1
-- numpy=1.12.1=py36_0
-- pandas=0.20.1=np112py36_0
-- pip=9.0.1=py36_1
-- python=3.6.1=2
-- python-dateutil=2.6.0=py36_0
-- pytz=2017.2=py36_0
-- requests=2.14.2=py36_0
-- scikit-learn=0.18.1=np112py36_1
-- scipy=0.19.0=np112py36_0
-- utm=0.4.2
+- arrow=0.13.1
+- attrdict=2.0.1
+- cheroot=8.4.2
+- future=0.18.0
+- google-auth=1.20.1
+- numpy=1.19.1
+- pandas=1.1.0
+- pip=20.2.2
+- python-dateutil=2.8.1
+- pytz=2020.1
+- requests=2.24.0
+- scikit-learn=0.23.2
+- scipy=1.5.2
+- utm=0.5.0
 - pip:
-  - jwcrypto==0.4.2
-  - pyfcm==1.4.3
+  - jwcrypto==0.7
+  - pyfcm==1.4.7
   - pygeocoder==1.2.5
-  - pymongo==3.5.1
+  - pymongo==3.11.0
 prefix: /Users/shankari/OSS/anaconda/envs/emission
--- a/setup/export_versions.sh
+++ b/setup/export_versions.sh
@@ -1,1 +1,1 @@
-export EXP_CONDA_VER=4.5.12
+export EXP_CONDA_VER=4.8.3
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -7,20 +7,5 @@
 
 source setup/checks/check_for_conda.sh
 
-echo "Setting up blank environment"
-conda create --name emission python=3.6
-conda activate emission
-
-echo "Downloading packages"
-curl -o /tmp/cachetools-2.1.0-py_0.tar.bz2 -L https://anaconda.org/conda-forge/cachetools/2.1.0/download/noarch/cachetools-2.1.0-py_0.tar.bz2
-curl -o /tmp/geojson-2.3.0-py_0.tar.bz2 -L https://anaconda.org/conda-forge/geojson/2.3.0/download/noarch/geojson-2.3.0-py_0.tar.bz2
-curl -o /tmp/jsonpickle-0.9.6-py_1.tar.bz2 -L https://anaconda.org/conda-forge/jsonpickle/0.9.6/download/noarch/jsonpickle-0.9.6-py_1.tar.bz2
-curl -o /tmp/more-itertools-8.2.0-py_0.tar.bz2 -L https://anaconda.org/conda-forge/more-itertools/8.2.0/download/noarch/more-itertools-8.2.0-py_0.tar.bz2
-curl -o /tmp/pyasn1-0.4.8-py_0.tar.bz2 -L https://anaconda.org/conda-forge/pyasn1/0.4.8/download/noarch/pyasn1-0.4.8-py_0.tar.bz2
-curl -o /tmp/pyasn1-modules-0.2.7-py_0.tar.bz2 -L https://anaconda.org/conda-forge/pyasn1-modules/0.2.7/download/noarch/pyasn1-modules-0.2.7-py_0.tar.bz2
-
-echo "Installing manually downloaded packages"
-conda install /tmp/*.bz2
-
-echo "Updating using conda now"
+echo "Installing using conda now"
 conda env update --name emission --file setup/environment36.yml
--- a/setup/setup_conda.sh
+++ b/setup/setup_conda.sh
@@ -12,7 +12,7 @@ else
     INSTALL_PREFIX=$HOME/miniconda-$EXP_CONDA_VER
     SOURCE_SCRIPT="$HOME/miniconda-$EXP_CONDA_VER/etc/profile.d/conda.sh"
     
-    curl -o miniconda.sh -L https://repo.continuum.io/miniconda/Miniconda3-$EXP_CONDA_VER-$PLATFORM.sh;
+    curl -o miniconda.sh -L https://repo.continuum.io/miniconda/Miniconda3-py38_$EXP_CONDA_VER-$PLATFORM.sh;
     bash miniconda.sh -b -p $INSTALL_PREFIX
     source $SOURCE_SCRIPT
     hash -r
--- a/setup/setup_notebook.sh
+++ b/setup/setup_notebook.sh
@@ -7,21 +7,6 @@
 
 source setup/checks/check_for_conda.sh
 
-echo "Setting up blank environment"
-conda create --name emission python=3.6
-conda activate emission
-
-echo "Downloading packages"
-curl -o /tmp/cachetools-2.1.0-py_0.tar.bz2 -L https://anaconda.org/conda-forge/cachetools/2.1.0/download/noarch/cachetools-2.1.0-py_0.tar.bz2
-curl -o /tmp/geojson-2.3.0-py_0.tar.bz2 -L https://anaconda.org/conda-forge/geojson/2.3.0/download/noarch/geojson-2.3.0-py_0.tar.bz2
-curl -o /tmp/jsonpickle-0.9.6-py_1.tar.bz2 -L https://anaconda.org/conda-forge/jsonpickle/0.9.6/download/noarch/jsonpickle-0.9.6-py_1.tar.bz2
-curl -o /tmp/more-itertools-8.2.0-py_0.tar.bz2 -L https://anaconda.org/conda-forge/more-itertools/8.2.0/download/noarch/more-itertools-8.2.0-py_0.tar.bz2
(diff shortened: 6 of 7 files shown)


shankari:
While testing this locally, ran into a regression caused by an assertion error in place squishing

The mismatch is in one of the initial speed computations

```
  File "/Users/kshankar/e-mission/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", l
ine 969, in link_trip_start
    ts, cleaned_trip_data, cleaned_start_place_data)
  File "/Users/kshankar/e-mission/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", l
ine 1152, in _fix_squished_place_mismatch
    assert False
```

The change is small, so depending on the results of the investigation, we could just fix the "ground truth" and move on. But let's try to understand how it happened first.

```
[0.0 1.3918799677570426

[0.0 1.391207493734843
```

shankari:
So the original speeds were almost identical, but after inserting, there was a mismatch.

```
1  5f34bd7e6fdd4851b0124abd  1.452702e+09  2016-01-13T08:22:27.391337-08:00
1  37.875767 -122.258413   41.756399  1.391880           1.391880
```

```
1  5f34bcc6be14470893941331  1.452702e+09  ...  1.391880           1.391207
```

The change happens while inserting

```
fix_squished_place: before inserting, speeds = [0.0, 0.31120032155672667, 5.279835474516015, 5.590681152758044, 5.590709355970388, 5.590737558966657, 5.422935527879857, 5.413431856340051, 6.160750357248137, 3.013363168584424]
fix_squished_place: after inserting, speeds = [0.0, 1.3918799677570426, 0.31120032155672667, 5.279835474516015, 5.590681152758044, 5.590709355970388, 5.590737558966657, 5.422935527879857, 5.413431856340051, 6.160750357248137]
```

```
fix_squished_place: before inserting, speeds = [0.0, 0.3111992781051712, 5.279608473667694, 5.590681151577752, 5.590709354830196, 5.590737557826475, 5.422942143907315, 5.413431855536246, 6.160681173619549, 3.013497148854115]
fix_squished_place: after inserting, speeds = [0.0, 1.391207493734843, 0.3111992781051712, 5.279608473667694, 5.590681151577752, 5.590709354830196, 5.590737557826475, 5.422942143907315, 5.413431855536246, 6.160681173619549]
```

Looking at the code, we insert and then fix a bunch of values. The insert check depends on the `distance_delta`, which is also slightly different

```
  distance_delta = 41.75639903271128 < 100, continuing with squish
  distance_delta = 41.73622481204529 < 100, continuing with squish
```

And the reset values are identical, except for some rounding.

```
squishing mismatch: resetting trip start_loc [-122.2584129, 37.8757669] to cleaned_start_place.location [-122.2587421, 37.8754958]
squishing mismatch: resetting trip start_loc [-122.258413, 37.875767] to cleaned_start_place.location [-122.258742, 37.875496]
```

shankari:
Let's see where we got these values from and how we rounded them. We got those
values from the `raw_start_place`, which is again consistent bar rounding.

```
Adding distance 8.619875014824052 to original 3111.61930762 to extend section start from [-122.2584253, 37.8758438] to [-122.2584129, 37.8757669]
After subtracting time 1.566663247456032 from original 539.978000164 to cover additional distance 8.619875014824052 at speed 5.502059889910047, new_start_ts = 1452702147.39

Raw place is 5f34bcc1be144708939411ad and corresponding location is 56967ed55771abda98
Adding distance 8.626548757749633 to original 3111.6193076220175 to extend section start from [-122.258425, 37.875844] to [-122.258413, 37.875767]
After subtracting time 1.5678762009787335 from original 539.978000164032 to cover additional distance 8.626548757749633 at speed 5.502059889910047, new_start_ts = 1452702147.3901238
```

But back in section segmentation, the related place seems to be accurate

with iloc
 'fmt_time': '2016-01-13T08:24:59.147000-08:00'
 'coordinates': [-122.2657953 37.8742511]}
 'ts': 1452702299.1470001

 'fmt_time': '2016-01-13T08:29:28.817000-08:00'
 'coordinates': [-122.2819714 37.8702897]}
 'ts': 1452702568.8169999

with iloc
 'coordinates': [-122.2657953 37.8742511]}
 'fmt_time': '2016-01-13T08:24:59.147000-08:00'
 'ts': 1452702299.147

 'coordinates': [-122.2819714 37.8702897]}
 'fmt_time': '2016-01-13T08:29:28.817000-08:00'
 'ts': 1452702568.817
```

**Comment by shankari** [comment]
Ok so looking further, the `[-122.258413, 37.875767]` is a recreated location,
and the timestamps are slightly different.

```
 'metadata': {'key': 'analysis/recreated_location'
 'data': Recreatedlocation({'fmt_time': '2016-01-13T08:22:27.391337-08:00'
 'loc': {'type': 'Point' 'coordinates': [-122.2584129 37.8757669]}
 'ts': 1452702147.3913367
 'distance': 41.75639903271128
 'speed': 1.3918799677570426

 'metadata': {'key': 'analysis/recreated_location'
 'data': Recreatedlocation({'fmt_time': '2016-01-13T08:22:27.390124-08:00'
 'loc': {'type': 'Point' 'coordinates': [-122.258413 37.875767]}
 'ts': 1452702147.3901238
 'distance': 41.73622481204529
 'speed': 1.391207493734843
```

Which is because the raw trip point is different

```
 'metadata': {'key': 'segmentation/raw_trip'
 'start_ts': 1452701818.444
 'start_fmt_time': '2016-01-13T08:16:58.444000-08:00'
 'start_loc': {'type': 'Point' 'coordinates': [-122.2583791 37.8757955]}
 'end_ts': 1452701938.572
 'end_fmt_time': '2016-01-13T08:18:58.572000-08:00'
 'end_loc': {'type': 'Point' 'coordinates': [-122.2584129 37.8757669]}
 'duration': 120.12800002098083
 'distance': 4.349090147122709}})

 'metadata': {'key': 'segmentation/raw_trip'
 'start_ts': 1452701818.444
 'start_fmt_time': '2016-01-13T08:16:58.444000-08:00'
 'start_loc': {'type': 'Point' 'coordinates': [-122.258379 37.875796]}
 'end_ts': 1452701938.572
 'end_fmt_time': '2016-01-13T08:18:58.572000-08:00'
 'end_loc': {'type': 'Point' 'coordinates': [-122.258413 37.875767]}
 'duration': 120.12800002098083
 'distance': 4.393622836921463}})
```

**Comment by shankari** [comment]
Not sure why the change would trigger an assert, though. While this may be different from the ground truth, it should be internally consistent.

**Comment by shankari** [comment]
So the mismatch while inserting the entry *is* caused by the distance mismatch

```
distance_delta = 41.75639903271128 < 100, continuing with squish
distance_delta = 41.73622481204529 < 100, continuing with squish
```

```
>>> 41.75639903271128 / 30
1.3918799677570426
>>> 41.73622481204529 / 30
1.391207493734843
```

```
fix_squished_place: after inserting, speeds = [0.0, 1.3918799677570426]
fix_squished_place: after inserting, speeds = [0.0, 1.391207493734843]
```

But why is it different while recalculating?

The new reconstructed location is the same

```
fix_squished_place: added new reconstructed location Recreatedlocation({'loc': {'type': 'Point', 'coordinates': [-122.2587421, 37.8754958]}, 'fmt_time': '2016-01-13T08:21:57.391337-08:00' to match new start point

fix_squished_place: added new reconstructed location Recreatedlocation({'loc': {'type': 'Point', 'coordinates': [-122.2587421, 37.8754958]}, 'fmt_time': '2016-01-13T08:21:57.390124-08:00' to match new start point
```

So the newly added point is the same, and the second point is the same, so the recalculated speeds are the same. But the distance was different so the non-validated insertion was different. What is different between the two distance calculations?

**Comment by shankari** [comment]
Ah, so I think I have the root cause! It looks like the coordinates in the `loc` field are rounded from the original `lat`, `lon` values.

```
fix_squished_place: after updating, old first location data = {'key': 'analysis/recreated_location', 'data': Recreatedlocation({'latitude': 37.8757669, 'longitude': -122.2584129, 'loc': {'type': 'Point', 'coordinates': [-122.2584129, 37.8757669]}})

fix_squished_place: after updating, old first location data = {'key': 'analysis/recreated_location', 'data': Recreatedlocation({'latitude': 37.8757669, 'longitude': -122.2584129, 'loc': {'type': 'Point', 'coordinates': [-122.258413, 37.875767]}})
```

When we compute the delta distance, we use the coordinates, which are rounded in this version

```
    distance_delta = ecc.calDistance(cleaned_trip_data.start_loc.coordinates,
                                     cleaned_start_place_data.location.coordinates)
```

```
squishing mismatch: resetting trip start_loc [-122.2584129, 37.8757669] to cleaned_start_place.location [-122.2587421, 37.8754958]
squishing mismatch: resetting trip start_loc [-122.258413, 37.875767] to cleaned_start_place.location [-122.258742, 37.875496]
```

But when we recompute the speeds and distances, we use the latitude and longitude fields.

```
    point_list = [ad.AttrDict(row) for row in points_df.to_dict('records')]
    zipped_points_list = list(zip(point_list, point_list[1:]))

    distances = [pf.calDistance(p1, p2) for (p1, p2) in zipped_points_list]
    distances.insert(0, 0)
```


```
def calDistance(point1, point2):
    return ec.calDistance([point1.longitude, point1.latitude], [point2.longitude, point2.latitude])
```

A simple fix would be to use a consistent value (either coordinates or lat/lon fields) throughout.
A different fix would be to see why the coordinates are rounded and fix that.

In this case, since we are upgrading, probably fixing the rounding is the safer fix.

**Comment by shankari** [comment]
Aha! from https://pypi.org/project/geojson/

> GeoJSON Object-based classes in this package have an additional precision attribute which rounds off coordinates to 6 decimal places (roughly 0.1 meters) by default and can be customized per object instance.

So our options are:
1. fix this particular assert by using the `pf.calDistance` method for both calculations. Note that this will retain precision differences between the two data sources and is sub-optimal.
1. fix this completely by specifying a custom precision at all places where we create geojson objects
1. fix this completely by rounding all lat/lon points to 6 decimal places. This is probably the right long-term solution, since the extra precision gives us a false measure of accuracy (see [also](https://stackoverflow.com/questions/63179028/serialize-geojson-in-python-with-rounded-coordinates)) but is likely to be a fairly heavy lift code-wise
1. punt this by reverting to an older version of geojson

Given that this change is already long and complex, I vote for (4).

**Comment by shankari** [comment]
From https://github.com/jazzband/geojson/blob/master/CHANGELOG.rst, this was introduced in version 2.5.0.
Reverting to 2.4.2 should fix it.

**Comment by shankari** [comment]
That fixed the assertion error but resulted in a ground truth regression. Let's try reverting back to 2.3.0, and if that doesn't work, we have to debug further.

```
FAIL: testZeroDurationPlaceInterpolationSingleSync (__main__.TestPipelineRealData)
AssertionError: 80961.11998618588 != 80856.14036388611 within 2 places (104.97962229976838 difference)
```

**Comment by shankari** [comment]
Reverting back to 2.3.0 didn't fix it. However, looking at the full list of regressions, we have an issue with some of the modules (e.g. location smoothing). If we are not smoothing out the correct values, that could be a reason for the difference in the distances. Let's fix that first.

**Comment by shankari** [comment]
The filtering error is

```
FAIL: testFilterSection (__main__.TestLocationSmoothing)
  File "emission/tests/analysisTests/intakeTests/TestLocationSmoothing.py", line 240, in testFilterSection
    self.assertEqual(len(filtered_points_entry.data.deleted_points), 12)
AssertionError: 8 != 12
```

And indeed we deleted 8 points instead of the original 12

```
bad_segments = [Segment(25, 26, 0.0), Segment(64, 65, 0.0), Segment(114, 121, 0.0), Segment(123, 124, 0.0), Segment(126, 127, 0.0), Segment(130, 131, 0.0)]
bad_segments = [Segment(25, 26, 0.0), Segment(64, 65, 0.0), Segment(113, 114, 0.0), Segment(121, 123, 7.372715972019115), Segment(124, 125, 0.0), Segment(126, 127, 0.0), Segment(130, 131, 0.0)]
```

```
after setting values, outlier_mask = [ 25  64 114 115 116 117 118 119 120 123 126 130]
after setting values, outlier_mask = [ 25  64 113 121 122 124 126 130]

```
after filtering ignored points, 12 -> 12
deleted 12 points

after filtering ignored points, 8 -> 8
deleted 8 points
```

shankari:
The main difference seems to start here:

```
shortest_non_cluster_segment = 9
shortest_non_cluster_segment = 5
```

Checking the segment list, we have (in both cases):

```
0. For cluster 0 - 25, distance = 8627.966220167396, is_cluster = False
1. For cluster 25 - 26, distance = 0.0, is_cluster = True
2. For cluster 26 - 64, distance = 14053.184009991968, is_cluster = False
3. For cluster 64 - 65, distance = 0.0, is_cluster = True
4. For cluster 65 - 114, distance = 15577.108477310048, is_cluster = False
5. For cluster 114 - 121, distance = 0.0, is_cluster = True
6. For cluster 121 - 124, distance = 7304.23306547181, is_cluster = False
7. For cluster 124 - 126, distance = 771.1159696538559, is_cluster = False
8. For cluster 126 - 127, distance = 0.0, is_cluster = True
9. For cluster 127 - 130, distance = 121.51096760524027, is_cluster = False
10. For cluster 130 - 138, distance = 1696.5583526264854, is_cluster = False
```

Segment 9 is clearly the shortest non-cluster segment.
Segment 5 is shorter, but is a cluster segment.

Why is the code picking it?

PR #787, opened 2021-01-24
Need help

corinne-hcr:
For this part, I was thinking that model_stage.py save the data to database. But I am not sure if it is necessary to use this file to process the data, I copy the "main" part so that this file can run.
image

For this part, I try to call get_all_common_places_for_user(user_id) in common_place_queries.py. I think uuid is user_id, and the previous code creates a list of uuid.

image

This part I was trying to see if it is enough to call "get_all_common_places_for_user()", but it doesn't work.
image

From this result, I think it is not the right function to plot. "get_place_ui(place)" It seems the "place" should be further processed. 
image


.gitignore
```diff
@@ -1,6 +1,7 @@
 *.pyc
 *.pdf
 *.swp
+.idea/*
```
shankari, line 4: revert? I think you are the only user who uses pycharm. Open to being convinced that it is really needed.

emission/analysis/plotting/leaflet_osm/our_plotter.py
```diff
@@ -103,7 +103,7 @@ def get_map_for_geojson_trip(geojson_trip):
                 print(section["properties"]["feature_type"], section["id"])
                 if (section["properties"]["feature_type"] == "section"):
                     section_line = get_section_ui(section)
-                    location_points.extend(section_line.location)
+                    location_points.extend(section_line.locations)
```
shankari, line 106: Revert, has already been merged.

emission/net/ext_service/habitica/proxy.py
```diff
@@ -19,7 +19,7 @@
 
 
 try:
-    key_file = open('conf/net/ext_service/habitica.json')
+    key_file = open('conf/net/ext_service/habitica.json.sample')
```
shankari, line 22: revert. if this is not present, we will just disable habitica. It is worse to try and connect to a non-existent URL

shankari (changes requested):
- still 5 files in the PR several of which seem unnecessary
- you might want to check what is going to be commited with git status before git commit
- if you are familiar with vi/vim, you can also use git commit (without -m ...) which will open up an editor that shows the files you are committing 

Other comments inline

shankari:
I'm confused about your question.
It looks like you are able to successfully retrieve all common places for a user. Congratulations!

image

shankari:
Couple of other high level thoughts:
- Why are you running the pipeline in the notebook? You already ran the pipeline from the command line, right? That stores the results in a database. Now you just have to read from the database to visualize.
- The PR is not complete.
   - I don't see the change to the environment file for pykov
   - This is not addressing other errors where the matplotlib import fails (e.g. in `emission/analysis/modelling/tour_model/featurization.py`)

corinne-hcr:
> Not sure what you mean by this. `get_all_common_places_for_user()` does in fact return a list of common places and you are able to print them as json as seen above.

I can't see my original comments. I used two ways to call `get_all_common_places_for_user()`. This one failed.
image





For this one, I was trying to see if the the place can be plot. I was calling `get_place_ui(place)`, But it seems the `place`should be processed.
image





> Why are you running the pipeline in the notebook? You already ran the pipeline from the command line, right? That stores the results in a database. Now you just have to read from the database to visualize.

What do you mean I run the pipeline in the notebook? Do you mean that I run` model_stage`? I wasn't very clear that if I should run this file in order to call `get_all_common_places_for_user()`since this file would finally store data through `save_common_place(common_place)`(in `common_place_queries.py`) in database. Since I run `get_all_common_places_for_user()`, it returns data from database,  I am not sure if the data should be processed (e.g. by `model_stage`)first.

So, I think I don't understand how the pipeline work. I just know it contains a series of processes, but I am not sure in which part (or which file I should call) to get the data I need.



After seeing your question. I trace `intake_single_user.py`, it looks that the data will be stored in Timeseries. Then, I find another way to call `get_all_common_places_for_user()`. I assume that it should be the way you want me to do.
image



> I don't see the change to the environment file for pykov

But in my environment, I do have pykov

image

> This is not addressing other errors where the matplotlib import fails (e.g. in emission/analysis/modelling/tour_model/featurization.py)

I don't understand what error it is. I do have matplotlib package. Where can I see `the matplotlib import fails`?

shankari:
> But in my environment, I do have pykov
> I don't understand what error it is. I do have matplotlib package. Where can I see the matplotlib import fails?

You probably installed `pykov` and `matplotlib` manually into your environment. So it works for you, but will not work in general because the next person to setup using `setup.sh` will not have those modules.

You need to add `pykov` to the list of modules loaded by `setup.sh`, which are in `environment36.yml`
You could try to add `matplotlib` there as well, but that just leads to mixing computation and visualization, which is generally a Bad Thing because it leads to module bloat.

You are correct that you don't appear to understand how pipelines work, and how we want to separate computation and visualization. You may want to read up on AI/ML pipelines before tomorrow's meeting.

shankari:
"machine learning pipeline"

shankari:
This is what a correct PR should look like:
https://github.com/e-mission/e-mission-server/pull/788

shankari:
Subsumed by #788

emission/pipeline/model_stage.py
```diff
@@ -11,7 +11,7 @@
 import emission.net.usercache.abstract_usercache as enua
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.storage.decorations.tour_model_queries as esdtmq
-from emission.core.get_database import get_db, get_mode_db, get_section_db
+# from emission.core.get_database import get_db, get_mode_db, get_section_db
```
corinne-hcr, line 14: Is this one not useful in the model_stage?
Trying to see if I can add any comment to a closed PR.

shankari: Looks like you can!
If you search for `get_db`, `get_mode_db` etc in this file, you don't find any matches. So it is not used any more. This has also bitrotted - as you can see from the blame (https://github.com/e-mission/e-mission-server/blame/20aac1f52204bd8af402f0cb63c917648836d60f/emission/core/get_database.py) it was there many years ago but has been removed now

PR #789, opened 2021-01-25
Fix notebook dependencies by adding matplotlib

shankari:
We didn't have matplotlib in the notebook dependencies!
Which is pretty surprising wrt plotting

Also include a fixed version of jedi to allow autocomplete to work
https://github.com/ipython/ipython/issues/12740
Should remove once fix is in place

Also remove unused notebook and notebook.exported files
notebook.exported since it is too OSX specific
notebook, because it has bitrotted compared to the main environment already

The additions is the way to keep things up to date since it avoids duplication (DRY principle)


shankari:
@corinne-hcr the visualization libraries go into the notebook additional yml
In your case, there is not much difference between having them in the basic yml and the notebook yml
but it ensures that we don't have to have the viz dependencies in the analytics pipeline.

PR #791, opened 2021-01-27
Refactor to pull out the visualization code

shankari:
Pull out the visualization code so it can be run in a notebook
New code is here:
https://github.com/e-mission/e-mission-eval-private-data/pull/14

Required a minor refactor since the prior call to graph was in the middle of
`delete_bins` - after the cutoff and before the remove.

We clearly don't want to generate graphs for this as part of a production
system. So it is correct to remove the `graph()` call.

To access the state of the object between the cutoff and the remove, we pull
out the cutoff into a separate function

### DIFF
--- a/emission/analysis/modelling/tour_model/similarity.py
+++ b/emission/analysis/modelling/tour_model/similarity.py
@@ -77,8 +77,7 @@ class similarity(object):
                 self.bins.append([a])
         self.bins.sort(key=lambda bin: len(bin), reverse=True)
 
-    #delete lower portion of bins
-    def delete_bins(self):
+    def calc_cutoff_bins(self):
         if len(self.bins) <= 1:
             self.newdata = self.data
             return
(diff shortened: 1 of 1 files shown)


emission/analysis/modelling/tour_model/similarity.py
```diff
@@ -77,8 +77,7 @@ def bin_data(self):
                 self.bins.append([a])
         self.bins.sort(key=lambda bin: len(bin), reverse=True)
 
-    #delete lower portion of bins
-    def delete_bins(self):
+    def calc_cutoff_bins(self):
         if len(self.bins) <= 1:
             self.newdata = self.data
             return
```
corinne-hcr, line 83: What will return here?

shankari: an empty return returns None
Note that I did not change that line

shankari:
It looks like @corinne-hcr can add comments to merged PRs as well, so merging this.

PR #795, opened 2021-02-17
Server side changes to support infinite scroll

shankari:
+ support querying by other time keys instead of only `metadata.write_ts`. This
ensures that we can accurately query for analysed results as well, since they
can be generated at any time after the raw data is collected.
+ support returning both the max and min value of any data field
    - Change the existing call to get_max_value_for_field to use it
    - Add a new method to the API that gets the pipeline range, not just the last entry
+ use this to figure out when to properly stop the infinite scroll

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -195,6 +195,15 @@ def getPipelineState():
     user_uuid = getUUID(request)
     return {"complete_ts": pipeline.get_complete_ts(user_uuid)}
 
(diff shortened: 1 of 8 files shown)


shankari:
This partially fixes https://github.com/e-mission/e-mission-docs/issues/226
There is a phone PR that complements this that I will check in soon as well.

shankari:
Related phone changes are at:
https://github.com/e-mission/e-mission-phone/pull/736

PR #798, opened 2021-02-20
add key to get confirmed trip in read_data

corinne-hcr:


emission/analysis/modelling/tour_model/cluster_pipeline.py
```diff
@@ -40,9 +40,14 @@
 """
 
 #read the data from the database. 
-def read_data(uuid=None):
-    trips = esda.get_entries(esda.CLEANED_TRIP_KEY, uuid,
-                             time_query=None, geo_query=None)
+def read_data(uuid=None,key='cleaned_trip'):
+    trips = None
+    if key == 'confirmed_trip':
```
shankari, line 45: This is functionally correct, but poor design. It would require modification every time we needed to switch to a new type of trip.

Please pass the key type in directly. The calling site will then need to pass in `esda.CONFIRMED_TRIP_KEY` or `esda.CLEANED_TRIP_KEY`.

shankari (changes requested):

emission/analysis/modelling/tour_model/cluster_pipeline.py
corinne-hcr: But all existing files that run this pipeline doesn't pass in key type. I am afraid that would need to change many files...

shankari: you would have to change all the files even with your change, though. Your check is `if 'confirmed_trip':... elif 'cleaned_trip'`

If it is neither, what will your code do?

corinne-hcr: Then we will see what kind of trip other people need to get. It indeed needs to add other if statements and other entry keys but we don't need to modify other files.
Even if I change this file now, that one line of code`trips = esda.get_entries(esda.CONFIRMED_TRIP_KEY, uuid, time_query=None, geo_query=None)` just read one kind of trips. Then I need to add all keys and maybe some of them wouldn't be used. I think we just make changes as needed.

corinne-hcr: I don't need to change all files. Before I made this change, Naomi just read in cleaned trips, so I set `cleaned_trip` to default, other existing files can run correctly without passing in key type. the evaluation I am doing (needs confirmed trip, different from the existing files) needs to pass in`confirmed_trip` and the pipeline can choose confirmed trip key.

shankari: you're right! I missed that detail.

But you can do that with `esda.CLEANED_TRIP` as well. Just set that to the default.
Fundamentally, I don't see what creating a new string and adding if checks gives you instead of just using `esda.CLEANED_TRIP` directly.

emission/analysis/modelling/tour_model/cluster_pipeline.py
```diff
@@ -40,9 +40,14 @@
...
+        trips = esda.get_entries(esda.CONFIRMED_TRIP_KEY, uuid,
+                                 time_query=None, geo_query=None)
+    elif key == 'cleaned_trip':
+        trips = esda.get_entries(esda.CLEANED_TRIP_KEY, uuid,
+                                 time_query=None, geo_query=None)
```
shankari, line 50:
```suggestion
def read_data(uuid=None,key=esda.CLEANED_TRIP):
    trips = esda.get_entries(key, uuid,
        time_query=None, geo_query=None)
```

shankari:
Can you please indicate the testing done before I merge?

corinne-hcr:
OK. All checks passed.
I can get confirm trips after making this change by calling         
`trips = pipeline.read_data(uuid=user, key=esda.CONFIRMED_TRIP_KEY)`

shankari:
I don't think the checks actually cover this code path (which is a different fix). But good to know that it works in the notebook as well.

PR #806, opened 2021-05-05
Purge user

PatGendre:
hi @shankari 
this is my proposal for adding the deletion of the usercache, timeseries_error and profiles collection to the purge_user.py script


PatGendre:
the only diff is in the bin/debug/common.py file

shankari (approved):
LGTM!

PR #808, opened 2021-05-06
Add a simple script to display the checked in/out report on the comma…

shankari:
…nd line

This is a workaround until we get started on the deployer dashboard.
It will be run automatically on login for now.


shankari:
This finally fixes all parts of https://github.com/e-mission/e-mission-docs/issues/633

PR #814, opened 2021-05-26
updated similarity code

corinne-hcr:


emission/analysis/modelling/tour_model/similarity.py
```diff
@@ -52,14 +61,16 @@ def __init__(self, data, radius):
                 end_lat = end_place.data.location["coordinates"][1]
                 logging.debug("endpoints are = (%s, %s) and (%s, %s)" %
                               (start_lon, start_lat, end_lon, end_lat))
-                if self.distance(start_lat, start_lon, end_lat, end_lon):
-                    self.data.remove(t)
+                if similarity(data, radius).distance(start_lat, start_lon, end_lat, end_lon):
+                    data.remove(t)
```
shankari, line 65: wait, you are creating a similarity object to compute the distance between two points?!!

emission/analysis/modelling/tour_model/similarity.py
```diff
@@ -35,11 +35,20 @@ class similarity(object):
     
     def __init__(self, data, radius):
         self.data = data
-        if not data:
-            self.data = []
```
shankari, line 39: So the constructor used to filter the data, but it doesn't any more. So you are changing the functionality of the constructor as part of a refactoring, which is generally a bad idea. A refactoring should change the structure, *not* the functionality.

shankari (changes requested):
This has taken multiple days, and is still very far from complete. I am going to make the change and submit a new PR instead of spending more time on explaining this or putting in suggestions, specially to code that is not even changed.

shankari:
Please work with https://github.com/e-mission/e-mission-server/pull/815 instead.

PR #820, opened 2021-06-08
Phone video instructions

start0318:
webpage with a single drop down button for phone videos.


webapp/www/faq/app_installation_instructions.html
```diff
@@ -0,0 +1,155 @@
...
+
+ </head>
+ <body>
+
+ <h3>Instructions</h3>
+ <p>Click on the button to open the dropdown menu. Select the video based on your phone type and follow the video guidelines.</p>
+ <p>1. Read and accept the privacy policy by clicking on the "I accept" button.</p>
+ <p>2. Please use the token given by your program administrator when "Login via token" window is prompted.</p>
+ <p>3. Select a username.</p>
+ <p>4. Complete the onboard survey.</p>
```
shankari, line 77:
```suggestion
 <p>4. Complete the onboarding survey.</p>
```

webapp/www/faq/app_installation_instructions.html
```diff
@@ -0,0 +1,155 @@
...
+
+     <div class="dropdown">
+       <button onclick="myFunction()" class="dropbtn">Dropdown</button>
+       <div id="myDropdown" class="dropdown-content">
+         <a href="#iOS 11-12">iOS 11-12</a>
+               <div id="videoDiv">
+                    <video id="video"
+                            width="480" height="360"
+                            controls>
+                        <source src="https://github.com/e-mission/e-mission-docs/raw/master/docs/assets/editediOS11_12.mov">
```
shankari, line 88: None of these videos actually show up. I think that only a small subset of videos is supported by the `<video>` tag
https://www.w3schools.com/tags/tag_video.asp

Note that the existing videos are all mp4.

webapp/www/faq/app_installation_instructions.html
```diff
@@ -0,0 +1,155 @@
...
+ <p>2. Please use the token given by your program administrator when "Login via token" window is prompted.</p>
+ <p>3. Select a username.</p>
+ <p>4. Complete the onboard survey.</p>
+   <nav>
+
+     <div class="dropdown">
+       <button onclick="myFunction()" class="dropbtn">Dropdown</button>
+       <div id="myDropdown" class="dropdown-content">
+         <a href="#iOS 11-12">iOS 11-12</a>
+               <div id="videoDiv">
```
shankari, line 84: Not sure we need the dropdown. If the videos are not part of the main text, then the yellow `<body>` part ends too early, and then the header is invisible as the user scrolls. See video. I think it is fine to simplify this and replace it with a static table.

shankari (changes requested):
I can ask my kids to fix the page if you are running out of time.

shankari:
Missing videos and invisible header screen.

Screen Recording 2021-06-08 at 7 45 24 AM mov

webapp/www/faq/app_installation_instructions.html
start0318: Converted videos to mp4 extension and uploaded to the e-mission docs repo.

shankari:
@start0318 I have merged the videos, so you can verify that they work, and potentially change the format to remove the dropdown before I merge this. LMK if you would like @mRaffill and @jRaffill to work on that part instead.

shankari:
Merging this for now; @mRaffill will make additional changes in a separate PR.

PR #823, opened 2021-06-21
Itinerum boolean commit #1

jruzekowicz:


emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -83,7 +83,8 @@ def __init__(self, user_id):
                 "inference/prediction": self.analysis_timeseries_db,
                 "analysis/inferred_section": self.analysis_timeseries_db,
                 "analysis/confirmed_trip": self.analysis_timeseries_db,
-                "analysis/confirmed_section": self.analysis_timeseries_db
+                "analysis/confirmed_section": self.analysis_timeseries_db,
+                "itinerum": self.analysis_timeseries_db
```
shankari, line 87: we don't need this.

```suggestion
```

setup/activate.sh
```diff
@@ -1,2 +1,2 @@
-source setup/activate_conda.sh
+# source setup/activate_conda.sh
```
shankari, line 1: This is specific to your windows machine; please revert.

emission/itinerum/itinerum_boolean.py
```diff
@@ -0,0 +1,49 @@
...
+            itinerum_boolean = False
+            ib = ecwib.Itinerum_Boolean()
+            ib.trip_id = trip.get_id()
+            ib.itinerum_received = itinerum_boolean
+            ib.start_ts = trip.data.start_ts
+            ib.end_ts = trip.data.end_ts
+            #Insert into the database as the boolean "false"
+            self.ts.insert_data(self.user_id, esda.ITINERUM_BOOLEAN, ib)
+            if self._last_trip_done is None or self._last_trip_done.data.end_ts < trip.data.end_ts:
+                self._last_trip_done = trip
```
shankari, line 49: @jruzekowicz @kafitz this is the key of what needs to be changed. The implementation here should be pretty straightforward, read all the data (trips, sections, points, etc) and store into a file labeled with the time range. No need to work with `esda.ITINERUM_BOOLEAN` which doesn't even exist

shankari (changes requested):
I think we are actually relatively close. Please ensure that changes specific to your windows environment (e.g. `miniconda.sh` and `activate.sh`) are not in the PR.

emission/itinerum/itinerum_boolean.py
shankari: To store the geojson output, look at https://github.com/e-mission/e-mission-server/blob/2177582fdea9a52faa8c1a238e62b111dbbcedba/emission/analysis/plotting/geojson/geojson_feature_converter.py

you can generate geojson for each trip instead of reading all the data. I would experiment with saving different kinds of outputs

shankari:
@jruzekowicz at a high level, please also submit your next PR from a branch and not from master.

emission/itinerum/itinerum_boolean.py
```diff
@@ -0,0 +1,49 @@
...
+    def last_trip_done(self):
+        return self._last_trip_done
+
+    def run_itinerum_boolean_pipeline(self, user_id, time_range):
+        #This is where we need to set the boolean?? 
+        #How is this done? 
+        #1.) seems to need to set the time series 
+        self.ts = esta.TimeSeries.get_time_series(user_id)
+        self.toSetFalseTrips = esda.get_entries(
+                esda.ITINERUM_BOOLEAN_KEY, user_id, time_query=time_range)
```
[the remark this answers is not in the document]
jruzekowicz: Can you expand on what you mean for this? Are you referencing the four lines or the entire file itself? In this file I was attempting to set a boolean to false. For the time series stuff, I was simply replicating what I found in other implementations as I am not concrete on the time series usage quite yet.

emission/core/wrapper/confirmedtrip.py
```diff
@@ -19,6 +19,7 @@ class Confirmedtrip(ecwt.Trip):
 # the user input will have all `manual/*` entries
 # let's make that be somewhat flexible instead of hardcoding into the data model
                   "user_input": ecwb.WrapperBase.Access.WORM
+                  "itinerum_boolean": ecwb.WrapperBase.Access.WORM
```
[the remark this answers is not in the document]
jruzekowicz: To clarify, are we not touching the confirmed trip object at all? Can you elaborate on how the confirmed trip fits into the entire intake pipeline and what is not included within the confirmed trip?

emission/core/wrapper/itinerum_boolean.py
```diff
@@ -0,0 +1,14 @@
+import emission.core.wrapper.wrapperbase as ecwb
+
+class ItinerumBoolean(ecwb.WrapperBase):
```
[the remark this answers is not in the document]
jruzekowicz: To clarify, is this because we are not adding to the confirmed trip object? Gabriel and I spoke about his pipeline stage which seems to be incorporated into the confirmed trip object. How does this differ from what we are doing here?

jruzekowicz:
> @jruzekowicz at a high level, please also submit your next PR from a branch and not from master.

will do, should I just branch from my master and commit from there?

jruzekowicz:
Do we have a structured way to test this? What would be the best method and approach here?

emission/core/wrapper/pipelinestate.py
```diff
@@ -24,6 +24,7 @@ class PipelineStages(enum.Enum):
     USER_MODEL = 7
     RECOMMENDATION = 8
     OUTPUT_GEN = 9
+    ITINERUM_BOOLEAN = 15
```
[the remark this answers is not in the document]
jruzekowicz: renamed to EXPORT_DATA

emission/core/wrapper/confirmedtrip.py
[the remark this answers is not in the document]
shankari: correct. The pipeline is structured so that each stage takes in an input and generates a distinct output. We don't modify inputs, consistent with standard pipeline guidelines. I document most of my design in issues - e.g. 
https://github.com/e-mission/e-mission-docs/issues/476

emission/core/wrapper/itinerum_boolean.py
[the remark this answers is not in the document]
shankari: correct; you are not changing the trip objects since "exported or not" is not an inherent property of a trip. In Gabriel's case, the trip labels are inherently part of the trip. The exported or not is currently represented by the pipeline state object for this state.

emission/itinerum/itinerum_boolean.py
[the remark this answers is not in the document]
shankari: the first comment was about `ITINERUM_BOOLEAN_KEY`.
You can't just use `esda.ITINERUM_BOOLEAN_KEY` without actually changing `esda`.
The other timeseries code works because there are constants defined there.

The second comment is that setting a boolean to false does not achieve your goal. Your goal is to actually export the trips in a way that @kafitz can read. which means that you need to actually save the trips (and the associated locations....)

See additional comments at https://github.com/e-mission/e-mission-server/pull/823#discussion_r655677462

shankari:
> will do, should I just branch from my master and commit from there?

For this PR, you can continue with this. For future PRs, please follow the fork and PR instructions.

> Do we have a structured way to test this? What would be the best method and approach here?

The server code includes unit tests in `emission/tests`

emission/itinerum/itinerum_boolean.py
jruzekowicz: Currently looking at the extract_timeline_for_day_range_and_user file. Starting with the main function, it begins the parser and a mutually exclusive group. Is this something I should also incorporate? My understanding of this part of the code is that it sets up an args object to get alist of uuid's. However, we appear to be working with a single uuid within the run_export_data_pipeline. Thus, I am assuming this will be slightly different.

jruzekowicz: It then calls export_timeline_for_users, passing in the uuid list and args. 
```
export_timeline(user_id=curr_uuid, start_day_str=args.start_day,
                end_day_str= args.end_day, timezone=args.timezone,
                file_name=args.file_prefix)
```
The main  function is this export_timeline section that pulls from args to get eh required variables. This is my thought process for needing the args object. 
However, in my run_export_data_pipeline, it passes in self, user_id, and time_range. I am trying to figure out the best way to build from here.

shankari: @jruzekowicz the code pointers that I provide to you are intended to give you pointers that you can adapt to your needs. They are not intended for you to be able to copy-paste directly. You need to understand the code, figure out which parts of it are relevant, and use it appropriately.

e.g. look up what is a parser (https://duckduckgo.com/?t=ffsb&q=python+argparse&ia=web). Is it relevant to what you are doing?

kafitz: Remember from a high level, there will be some sort of scheduler process that invokes your script periodically. This probably won't be supplying any arguments itself. If argparse is responsible for taking in UUIDs from the CLI, there will be some other way our script will have to get these variables. 

Perhaps it's even easiest to hardcode a list now until it can be generated, depends on where you want to start.

shankari: @kafitz this is currently structured as an additional stage of the intake pipeline, 
https://github.com/e-mission/e-mission-server/pull/823/files#diff-694dff3ba8c686137cb4726369a03b612728562e37fde55292b887269153fdaeR176

so it would be invoked as part of the regular intake script

To recap:
- there is a **new stage** which will be invoked as part of the intake pipeline
   - it will export values to a file
   - it will get its inputs from the intake pipeline and the pipeline state
- there is an **existing** script designed for one-time usage which exports **all** data
    - it gets its inputs from argparse
    - its export code can be reused/adapted for the new stage
    - however it cannot be directly copy-pasted

My question/hint was to indicate that Jenna should not use argparse for the new stage since the new stage will get its inputs from the pipeline

PR #825, opened 2021-06-28
Implement label inference pipeline

GabrielKS:
+ Adds a stage to the intake pipeline to perform label inference
+ Includes several placeholder label inference algorithms to test the system
+ Unit tests coming soon!


emission/analysis/userinput/matcher.py
```diff
@@ -97,3 +98,11 @@ def get_user_input_dict(ts, tct, input_key_list):
     logging.debug("for trip %s, returning user input dict %s" % (tct.get_id(), tct_userinput))
     return tct_userinput
 
+# For a given trip, find the corresponding list of label inferences if it has been generated
+def get_inferred_labels_list(trip):
+    candidates = esdt.get_sections_for_trip(esda.INFERRED_LABELS_KEY, trip.user_id, trip.get_id())
+    if len(candidates) == 0: return {}  # Perhaps we have not run the inference step for this trip
+    assert len(candidates) == 1, \
+        "Multiple label inference list objects for trip "+str(trip.get_id())
+    return candidates[0]["data"]["prediction"]
```
shankari, line 107: Let's think through this some more. If we haven't yet run the inference step for this trip, and we run it later, how will those labels assigned to the `confirmed_trip` object? The `last_processed_ts` for the `eaum.create_confirmed_objects` step would have gone past this trip, so we won't consider it again.

For the actual user inputs, we deal with this by having an additional `eaum.match_incoming_user_inputs` whose `last_processed_ts` is keyed to the `write_ts` of the user input objects. (see background at https://github.com/e-mission/e-mission-docs/issues/476#issuecomment-561727669) I don't see similar code here.

shankari (changes requested):
Looks good to be overall except the questions about the prediction structure and the matching of the labels and the confirmed trips.

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -0,0 +1,121 @@
...
+            esda.CLEANED_TRIP_KEY, user_id, time_query=time_range)
+        for trip in self.toPredictTrips:
+            prediction = predict_trip(trip)
+            lp = ecwl.Labelprediction()
+            lp.trip_id = trip.get_id()
+            lp.prediction = prediction
+            lp.start_ts = trip.data.start_ts
+            lp.end_ts = trip.data.end_ts
+            # Insert the Labelprediction into the database as its own independent document
+            self.ts.insert_data(self.user_id, esda.INFERRED_LABELS_KEY, lp)
```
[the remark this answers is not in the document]
GabrielKS: Am I understanding correctly that for mode inference, `inference/prediction` is to store each algorithm's result and `analysis/inferred_section` is to store the final (perhaps ensemble) value? If that's the case, then yes, I suppose it might be useful to do the same for label inference.

emission/core/wrapper/labelprediction.py
```diff
@@ -0,0 +1,16 @@
+# Based on modeprediction.py
+import emission.core.wrapper.wrapperbase as ecwb
+
+class Labelprediction(ecwb.WrapperBase):
+    props = {"trip_id":   ecwb.WrapperBase.Access.WORM,  # the trip that this is part of
+            "prediction": ecwb.WrapperBase.Access.WORM,  # What we predict
```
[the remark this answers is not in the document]
GabrielKS: Will do in the next commit.

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -0,0 +1,121 @@
...
+        self._last_trip_done = None
+    
+    @property
+    def last_trip_done(self):
+        return self._last_trip_done
+
+    def run_prediction_pipeline(self, user_id, time_range):
+        self.ts = esta.TimeSeries.get_time_series(user_id)
+        self.toPredictTrips = esda.get_entries(
+            esda.CLEANED_TRIP_KEY, user_id, time_query=time_range)
```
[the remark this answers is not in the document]
GabrielKS: Sounds good.

emission/analysis/classification/inference/labels/pipeline.py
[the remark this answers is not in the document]
shankari: Correct. At least, that is the long-term goal. Right now, there is only one implementation 😄

emission/core/wrapper/labelprediction.py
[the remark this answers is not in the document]
GabrielKS: Added in https://github.com/e-mission/e-mission-server/pull/825/commits/0f1ed52e7525ea5f41d63642bca1eaeb9d5af8ba.

emission/analysis/classification/inference/labels/pipeline.py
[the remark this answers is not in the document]
GabrielKS: As of https://github.com/e-mission/e-mission-server/pull/825/commits/dee24619a30582baa8c8699fe7ec7ac7edb9bfde, we are now reading confirmed trips anyway.

emission/analysis/classification/inference/labels/pipeline.py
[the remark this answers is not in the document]
GabrielKS: Implemented in https://github.com/e-mission/e-mission-server/pull/825/commits/d0136fc8c041739bf2809f74b3e8b44846957785.

emission/analysis/userinput/matcher.py
GabrielKS: Cleaned up in https://github.com/e-mission/e-mission-server/pull/825/commits/dee24619a30582baa8c8699fe7ec7ac7edb9bfde.

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        self.ts = esta.TimeSeries.get_time_series(user_id)
+        self.toPredictTrips = esda.get_entries(
+            esda.CONFIRMED_TRIP_KEY, user_id, time_query=time_range)
+        for trip in self.toPredictTrips:
+            results = self.compute_and_save_algorithms(trip)
+            ensemble = self.compute_and_save_ensemble(trip, results)
+
+            # Add final prediction to the confirmed trip entry in the database
+            trip["data"]["inferred_labels"] = ensemble["prediction"]
+            self.ts.update(trip)
```
shankari, line 119: I saw your question about this. In general, we have had the input to each pipeline stage be read-only, so each stage generates new output. That is why we avoid `ts.update`. This made sense when the outputs could be significantly different from the inputs (e.g. the cleaned trip is the segmented trip + some extrapolation at the start). 

But it gets a bit more complicated when we are just changing a few fields in the object; creating a new object (e.g. `labeled_trip`) is going to require rewriting the client as well to read the new kind of object.

If we did want to go down that route, we would create a new `labeled_trip` or `inferred_labeled_trip` datastructure, which would include inferred labels but not user inputs and be the output of this stage.

we would then create `confirmed_trip` objects from the `inferred_labeled_trip` instead of `cleaned_trip`. Now that I've written it out, maybe that is cleaner after all than modifying the `confirmed_trip` directly.

It retains the high level assumptions of the pipeline and does not require client changes.

Thoughts?

shankari:
The other changes look good.

emission/analysis/classification/inference/labels/pipeline.py
GabrielKS: I agree that we could do it without `ts.update` reasonably cleanly and without requiring client changes by adding a new trip data structure. It does seem a bit overkill to copy the entire trip just to add a single field, but I'm not too well-versed in "classic ML algorithms"; if this is an established standard, maybe we should follow it.

shankari: Maybe more "classic ML systems" - e.g. Spark et al. Making the data structures read-only allows you to maintain genealogy and roll forward computations efficiently.  How big is each trip structure? Unlike sections, they don't store the actual trajectory-related information, so I would assume that they are fairly lightweight.

GabrielKS: You're right, the `repr()` of each confirmed trip is on the order of 1500 characters.

GabrielKS:
Et voilà, no `ts.update`. Everything is very pretty now. :)

shankari:
This looks fine to me. Are you going to include unit tests now, or in a later PR?

GabrielKS:
I'll write unit tests now.

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -53,11 +53,11 @@ def placeholder_predictor_1(trip):
 
 # This third scenario provides labels designed to test the soundness and resilience of
 # the client-side inference processing algorithms.
-trip_n = 0  # ugly use of globals for testing only
 def placeholder_predictor_2(trip):
-    global trip_n  # ugly use of globals for testing only
-    trip_n %= 6
-    trip_n += 1
+    # Hardcoded to match "test_july_22" -- clearly, this is just for testing
+    timestamp2index = {494: 5, 565: 4, 795: 3, 805: 2, 880: 1, 960: 0}
```
shankari, line 58: so these represent the timestamps of the trips in `test_july_22`? Does this still work for your ongoing testing?

emission/tests/pipelineTests/TestLabelInferencePipeline.py
```diff
@@ -0,0 +1,85 @@
...
+class TestLabelInferencePipeline(unittest.TestCase):
+    # It is important that these functions be deterministic
+    test_algorithms = {
+            ecwl.AlgorithmTypes.PLACEHOLDER_0: eacilp.placeholder_predictor_0,
+            ecwl.AlgorithmTypes.PLACEHOLDER_2: eacilp.placeholder_predictor_2
+    }
+
+    def setUp(self):
+        np.random.seed(61297777)
+        self.user_id = ecwu.User.fromEmail("test_july_22").uuid
```
shankari, line 26: This assumes that the `test_july_22` user is loaded into the database. For automated tests, it is better to load the data in the setup as well. Please see the functions in `emission/tests/common.py` and their usage, (e.g. `setupRealExample`) in other tests in `emission/tests` for libraries to load the data and run the pipeline.

emission/tests/pipelineTests/TestLabelInferencePipeline.py
```diff
@@ -0,0 +1,85 @@
...
+        self.reset_pipeline()
+        self.run_pipeline(self.test_algorithms)
+        time_range = estt.TimeQuery("metadata.write_ts", None, time.time())
+        self.cleaned_trips = esda.get_entries(esda.CLEANED_TRIP_KEY, self.user_id, time_query=time_range)
+        self.cleaned_id_to_trip = {trip.get_id(): trip for trip in self.cleaned_trips}
+        self.inferred_trips = esda.get_entries(esda.INFERRED_TRIP_KEY, self.user_id, time_query=time_range)
+
+    def tearDown(self):
+        self.reset_pipeline()
```
shankari, line 35: Delete the newly loaded data here.

emission/tests/pipelineTests/TestLabelInferencePipeline.py
```diff
@@ -0,0 +1,85 @@
...
+import emission.core.wrapper.user as ecwu
+import emission.pipeline.intake_stage as epi
+import emission.pipeline.reset as epr
+import emission.analysis.classification.inference.labels.pipeline as eacilp
+import emission.core.wrapper.labelprediction as ecwl
+import emission.storage.decorations.analysis_timeseries_queries as esda
+import emission.storage.timeseries.timequery as estt
+import emission.storage.decorations.trip_queries as esdt
+
+class TestLabelInferencePipeline(unittest.TestCase):
```
shankari, line 17: Please also add a `main` (similar to the other tests in `emission/tests`) to make it easier to run the tests in isolation. I'm not quite sure how you ran the tests without that.

shankari (changes requested):

emission/analysis/classification/inference/labels/pipeline.py
GabrielKS: It does — so far I've only been testing with `test_july_22`. This was the simplest way I could think of to remove `trip_n` and make the prediction a pure function of the trip, but if it's not flexible enough for me later on, I'll find a more complicated way.

emission/tests/pipelineTests/TestLabelInferencePipeline.py
GabrielKS: Will do. I ran with `./e-mission-py.bash -m unittest -v emission.tests.pipelineTests.TestLabelInferencePipeline`.

emission/tests/pipelineTests/TestLabelInferencePipeline.py
GabrielKS: Addressed in the most recent commit

emission/tests/pipelineTests/TestLabelInferencePipeline.py
GabrielKS: Addressed in the most recent commit

shankari (approved):
LGTM!

PR #826, opened 2021-06-30
check unit test code

corinne-hcr:
TestDataPreprocessing.py and TestGetUsers.py need to read trips. Also need help with setup and teardown function.

### DIFF
--- a/emission/analysis/modelling/tour_model/data_preprocessing.py
+++ b/emission/analysis/modelling/tour_model/data_preprocessing.py
@@ -1,0 +1,58 @@
+import emission.storage.decorations.analysis_timeseries_queries as esda
+import emission.analysis.modelling.tour_model.cluster_pipeline as pipeline
+import emission.analysis.modelling.tour_model.similarity as similarity
+import pandas as pd
+from sklearn.model_selection import KFold
+
+
+# read data that have user labels
+def read_data(user):
+    trips = pipeline.read_data(uuid=user, key=esda.CONFIRMED_TRIP_KEY)
+    return trips
+
+
+# - trips: all trips read from database
+# - filter_trips: valid trips that have user labels and are not points
+def filter_data(trips,radius):
+    non_empty_trips = [t for t in trips if t["data"]["user_input"] != {}]
+    non_empty_trips_df = pd.DataFrame(t["data"]["user_input"] for t in non_empty_trips)
+    valid_trips_df = non_empty_trips_df.dropna(axis=0, how='any', thresh=None, subset=None, inplace=False)
+    valid_trips_idx_ls = valid_trips_df.index.tolist()
+    valid_trips = [non_empty_trips[i]for i in valid_trips_idx_ls]
+
+    # similarity codes can filter out trips that are points in valid_trips
+    filter_trips = similarity.filter_too_short(valid_trips, radius)
+    return filter_trips
+
+
+# use KFold (n_splits=5) to split the data into 5 models (5 training sets, 5 test sets)
+def extract_features(filter_trips):
+    X = []
+    for trip in filter_trips:
+        start = trip.data.start_loc["coordinates"]
+        end = trip.data.end_loc["coordinates"]
+        distance = trip.data.distance
+        duration = trip.data.duration
+        X.append([start[0], start[1], end[0], end[1], distance, duration])
+    return X
+
+def split_data(filter_trips):
+    X = extract_features(filter_trips)
+    kf = KFold(n_splits=5, shuffle=True, random_state=3)
+    train_idx = []
+    test_idx = []
+    for train_index, test_index in kf.split(X):
+        train_idx.append(train_index)
+        test_idx.append(test_index)
+    return train_idx, test_idx
+
+
+# collect a set of data(training/test set) after splitting
+def get_subdata(filter_trips,train_test_set):
+    collect_sub_data = []
+    for train_test_subset in train_test_set:
+        sub_data = []
+        for idx in train_test_subset:
+            sub_data.append(filter_trips[idx])
+        collect_sub_data.append(sub_data)
+    return collect_sub_data
--- a/emission/analysis/modelling/tour_model/get_request_percentage.py
+++ b/emission/analysis/modelling/tour_model/get_request_percentage.py
@@ -1,0 +1,135 @@
+import label_processing as label_pro
+import copy
+import itertools
+
+
+# This function is to compare a trip with a group of trips to see if they happened in a same day
+def match_day(trip,bin,filter_trips):
+    if bin:
+        t = filter_trips[bin[0]]
+        if trip['data']['start_local_dt']['year']==t['data']['start_local_dt']['year']\
+                and trip['data']['start_local_dt']['month']==t['data']['start_local_dt']['month']\
+                and trip['data']['start_local_dt']['day']==t['data']['start_local_dt']['day']:
+            return True
+    return False
+
+
+# This function is to compare a trip with a group of trips to see if they happened in a same month
+def match_month(trip,bin,filter_trips):
+    if bin:
+        t = filter_trips[bin[0]]
+        if trip['data']['start_local_dt']['year']==t['data']['start_local_dt']['year']\
+                and trip['data']['start_local_dt']['month']==t['data']['start_local_dt']['month']:
+            return True
+    return False
+
+
+# This function bins trips according to ['start_local_dt']
+def bin_date(trip_ls,filter_trips,day=None,month=None):
+    bin_date = []
+    for trip_index in trip_ls:
+        added = False
+        trip = filter_trips[trip_index]
+
+        for bin in bin_date:
+            if day:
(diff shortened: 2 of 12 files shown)


shankari:
Please submit a separate PR for the code refactoring since the unit tests are not complete.

corinne-hcr:
Just for records.
1.I am currently reading in `shankari_2016-06-20.expected_confirmed_trips`. I can get 6 confirmed trips in total. But when the process goes to `similarity` code (`filter_too_short` function), the code cannot read `start_place`.
2.To be able to test the functions for analyzing common trips, I need to make common trips.
3.Here are codes for creating `start_place`(or `end_place`), suggested by @shankari 
```
import emission.core.wrapper.place as ecwp
start_place = ecwp.Place()
start_place.location = ct.start_loc
start_place['_id'] = ct.start_place
import emisson.storage.timeseries.abstract_timeseries as esta
esta.update(start_place)

P.S. ct is an instance of a confirmed trip for which you are trying to create a start place
```

corinne-hcr:
I am still blocked by this:
`start_place` is an instance of Place(), contains the location the id. But `esta.update(start_place)` doesn't work.
```
AttributeError: module 'emission.storage.timeseries.abstract_timeseries' has no attribute 'update'
```
The definition of `update()` is `pass`. 
I don't know how to solve that.

shankari:
`pass` is a static method in timeseries (`esta.Timeseries.update`) should work. Again, my suggestions are not intended to be runnable code - they give you an indication of the methods you should look at, but you need to actually understand the python and adapt as necessary.

corinne-hcr:
I run these codes to get `filter_trips`, but`esta.TimeSeries.update(start_place)` doesn't work. 
The error is 
```    
start_lon = start_place.data.location["coordinates"][0]
AttributeError: 'NoneType' object has no attribute 'data'
```
```
import emission.core.wrapper.place as ecwp
import emission.storage.timeseries.abstract_timeseries as esta
self.readAndStoreTripsFromFile("emission/tests/data/real_examples/shankari_2016-06-20.expected_confirmed_trips")
ts = esta.TimeSeries.get_time_series(self.testUUID)
user = self.testUUID
trips = preprocess.read_data(user)
for trip in trips:
    ct = trip
    start_place = ecwp.Place()
    start_place.location = ct.data.start_loc
    start_place['_id'] = ct.data.start_place
    esta.TimeSeries.update(start_place)
filter_trips = preprocess.filter_data(trips, 100)
print(filter_trips)
```

shankari:
***Again, my suggestions are not intended to be runnable code - they give you an indication of the methods you should look at, but you need to actually understand the code and adapt as necessary.***

You are trying to read `start_place.data.location`. You are setting `start_place.location`. Make them be consistent and it will work.

If I need to give you runnable code, I might as well write the code myself.

corinne-hcr:
To be consistent like this?
But it still doesn't work
```
user = self.testUUID
trips = preprocess.read_data(user)
for trip in trips:
  ct = trip
  ct.data.start_place = ecwp.Place()
  ct.data.start_place.location = ct.data.start_loc
  ct.data.start_place.data['_id'] = ct.data.start_place
  esta.TimeSeries.update(ct.data.start_place.data)
  test = similarity.filter_too_short([ct], 100)
```
For this line,
```
ct.data.start_place = ecwp.Place()
```
 the error is 
```
AttributeError: property start_place is read-only
```
If I use 
```
start_place.location = ct.start_loc
start_place.data['_id'] = ct.start_place
```
the error is
```
AttributeError: property start_loc is not defined for Entry
```
 Since ct is an instance of a confirmed trip, so it is in Entry class. Only `ct.data.start_loc` can access `start_loc`

shankari:
@corinne-hcr have you checked out the e-mission data model and the different wrapper classes? 
https://github.com/e-mission/e-mission-server/tree/master/emission/core/wrapper

They are linked from the timeseries notebook (`Timeseries_Sample.ipynb`) that you experimented with earlier. The notebook also outlines the various ways to access the timeline objects as entry objects, data tables, etc.

> To be consistent like this?

No. You don't want to modify the cleaned trip object. you want to read the cleaned trip object and modify the start place, since that is the object that you are creating.

Concretely. Entry objects should set data into the "data" sub-object.

```
    start_place["data"]["location"] = ct.data.start_loc
```

shankari:
Double-checking the code

```
def update(entry):
```

works with an entry.

The `Entry` class includes both `data` and `metadata`. `Place` is an example of `data`, so you need to call 

```
def update_data(user_id, key, obj_id, data):
```

instead.

shankari:
To clarify: my previous comment:

> Concretely. Entry objects should set data into the "data" sub-object.

Assumed that you were creating an `Entry` object. Since you are creating a `Place` object that corresponds to `data`, you can use `.location` directly.

shankari:
Again, looking at the implementation of `update_data` in `TimeSeries`, it is a static method without a concrete implementation. you need to call the method in `BuiltinTimeSeries`, which has the actual implementation instead. Note that static methods are not overridden by subclasses.

corinne-hcr:
I figured that I would get a different result if I changed `similarity` to extract coordinates directly from `trip.data.start_loc["coordinates"]`(for example)  instead of `start_place = esda.get_entry(esda.CLEANED_PLACE_KEY,  trip.data.start_place)`
Here is the result from `start_place = esda.get_entry(esda.CLEANED_PLACE_KEY,  trip.data.start_place)`
1st round
2nd round

Here is the result from` trip.data.start_loc["coordinates"]`
image
image

Also, the way I extract coordinates for the 2nd round now is just extracting from the confirmed trip, not from place. Should I change?

shankari:
I can't see the images, I am not sure that they were uploaded correctly.
Having said that:
- the existing results are not immutable. they are not every good (so that we need to strive to meet them) and the choice of using the place v/s the trip locations is based on a decision by an undergrad who worked on the project for a summer. I don't think there was a lot of thought put into it, at least, there's no documentation of such thinking.
- I am extremely surprised that the locations for the trip don't match the locations for the place
- This is why unit tests are important. If we had them, we would be able to see where the difference originated and see if it was meaningful, instead of looking at the final result and trying to make sense of it.

shankari:
@corinne-hcr I still can't see the images

shankari:
@corinne-hcr I can see the images now. I don't see a significant difference between the two results; while individual points are moved around, that falls within the bounds of normal probability differences. If you are familiar with generating boxplots, and can generate a boxplot with the results, that would make it more clear. But generating the boxplot is not a priority.

shankari:
At a high level, there should typically be multiple tests for each function to test various scenarios that might happen, not just ones that happened to occur while processing one particular user in one particular dataset and one possible method. (e.g. https://github.com/corinne-hcr/e-mission-server/pull/2/commits/b46a370cc656b4c28f38986375ba4a296ec05f60) However, this should hopefully catch most regressions while refactoring so it is a good start!

Can you list out what additional tests you plan to complete?

I will review this tomorrow, I have a presentation to finish tonight to meet my own deadlines.

corinne-hcr:
Currently, the similarity code is not significant changed. @shankari added new ways of accessing. But the old ways should still work. So the test of similarity code should still be fine.

shankari:
@corinne-hcr so do you have any additional questions? How many more tests do you plan to write and what is their ETA?

corinne-hcr:
I plan to write 3 more tests - get_score, second_round_of_clustering, evaluation_pipeline
They should be done by tomorrow's meeting.
I feel overwhelmed with so many changes at this point. I need some time tomorrow to digest the recent PR and changed ways. I don't know which way and what analysis I need to run and put in the paper. I think I will need to ask some questions after reading the codes. Please forgive me that I am probably not able to fully understand the new analysis in a very short time based on my knowledge and skillset.

shankari:
@corinne-hcr I would start with https://github.com/e-mission/e-mission-eval-private-data/commit/abf4f78b417de7190cd35b8e3e1877df938ee1be which compares the similarity code with different settings. In particular, the results around whether or not to filter and whether or not to use the cutoff. Note that I am now returning the labels from similarity (mapped to the original trip indices) so we don't need to maintain the "trip" data structure any more. 

Note also that I have different results for the `filtered_cutoff` case than you do. I am pretty sure this is because of 
https://github.com/e-mission/e-mission-eval-private-data/pull/28#issuecomment-886230514

I don't care about that for the pipeline changes, since my selected settings are `no_filter_no_cutoff`, but you can feel free to do so!

corinne-hcr:
> I don't care about that for the pipeline changes, since my selected settings are `no_filter_no_cutoff`, but you can feel free to do so to add additional analysis results to the paper!

I remember you mentioned `single level` in one of the PR. Do you mean that you just use 1 round of clustering instead of 2 rounds? If so, the evaluation_pipeline is not useful anymore?

shankari:
Not sure what you mean by `single_level` (link to code would be helpful). But yes, for the deployment on Monday night, I will use only the first round, and I will not filter or cutoff any trips, and I will not model only one split. I anticipate changing the evaluation pipeline significantly, as I said earlier.

PR #829, opened 2021-07-19
Modeling and functions

corinne-hcr:
Move codes from private data repo to here
These are what I have for now.
When I am done with refactoring, I will update the codes here.

### DIFF
--- a/bin/build_label_model.py
+++ b/bin/build_label_model.py
@@ -1,0 +1,67 @@
+"""
+Script to build and save the labeling model.
+"""
+import logging
+
+import argparse
+import uuid
+import copy
+
+import emission.pipeline.reset as epr
+import emission.core.get_database as edb
+import emission.core.wrapper.user as ecwu
+import emission.storage.timeseries.abstract_timeseries as esta
+import emission.analysis.modelling.tour_model_first_only.build_save_model as eamtb
+
+def _get_user_list(args):
+    if args.all:
+        return _find_all_users()
+    elif args.platform:
+        return _find_platform_users(args.platform)
+    elif args.email_list:
+        return _email_2_user_list(args.email_list)
+    else:
+        assert args.user_list is not None
+        return [uuid.UUID(u) for u in args.user_list]
+
+def _find_platform_users(platform):
+    # Since all new clients register a profile with the server, we don't have
+    # to run a 'distinct' query over the entire contents of the timeseries.
+    # Instead, we can simply query from the profile users, which is
+    # significantly faster
+    # Use the commented out line instead for better performance.
+    # Soon, we can move to the more performant option, because there will be
+    # no users that don't have a profile
+    # return edb.get_timeseries_db().find({'metadata.platform': platform}).distinct(
+    #    'user_id')
+   return edb.get_profile_db().find({"curr_platform": platform}).distinct("user_id")
+
+def _find_all_users():
+   return esta.TimeSeries.get_uuid_list()
+
+def _email_2_user_list(email_list):
+    return [ecwu.User.fromEmail(e).uuid for e in email_list]
+
+if __name__ == '__main__':
+    logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s',
+        level=logging.DEBUG)
+
+    parser = argparse.ArgumentParser()
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument("-a", "--all", action="store_true", default=False,
+        help="build the model for all users")
+    group.add_argument("-p", "--platform", choices = ['android', 'ios'],
+                        help="build the model for all on the specified platform")
+    group.add_argument("-u", "--user_list", nargs='+',
+        help="user ids to build the model for")
+    group.add_argument("-e", "--email_list", nargs='+',
+        help="email addresses to build the model for")
+
+    args = parser.parse_args()
+    print(args)
+
+    user_list = _get_user_list(args)
+    logging.info("received list with %s users" % user_list)
+    for user_id in user_list:
+        logging.info("building model for user %s" % user_id)
+        eamtb.build_user_model(user_id)
--- a/bin/debug/label_stats.py
+++ b/bin/debug/label_stats.py
@@ -1,0 +1,41 @@
+import emission.core.get_database as edb
+import uuid
+import argparse
+
+
+parser = argparse.ArgumentParser(prog="intake_single_user")
+group = parser.add_mutually_exclusive_group(required=True)
+group.add_argument("-e", "--user_email")
+group.add_argument("-u", "--user_uuid")
+
+args = parser.parse_args()
+if args.user_uuid:
+    sel_uuid = uuid.UUID(args.user_uuid)
+else:
+    sel_uuid = ecwu.User.fromEmail(args.user_email).uuid
+
+print("All inferred trips %s" % edb.get_analysis_timeseries_db().count_documents({"metadata.key": "analysis/inferred_trip", "user_id": sel_uuid}))
+
+print("Inferred trips with inferences %s" % edb.get_analysis_timeseries_db().find({"metadata.key": "analysis/inferred_trip", "user_id": sel_uuid, "data.inferred_labels": {"$ne": []}}).count())
+
+print("All expected trips %s" % edb.get_analysis_timeseries_db().count_documents({"metadata.key": "analysis/expected_trip", "user_id": sel_uuid}))
+
+print("Expected trips with inferences %s" % edb.get_analysis_timeseries_db().find({"metadata.key": "analysis/expected_trip", "user_id": sel_uuid, "data.expectation": {"$exists": True}}).count())
+
+for t in list(edb.get_analysis_timeseries_db().find({"metadata.key": "analysis/inferred_trip", "user_id": sel_uuid})):
+    if t["data"]["inferred_labels"] != []:
+        confirmed_trip = edb.get_analysis_timeseries_db().find_one({"user_id": t["user_id"],
+                "metadata.key": "analysis/confirmed_trip",
+                "data.start_ts": t["data"]["start_ts"]})
+        if confirmed_trip is None:
+            print("No matching confirmed trip for %s" % t["data"]["start_fmt_time"])
+            continue
+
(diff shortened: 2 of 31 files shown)


shankari:
@corinne-hcr can you clarify that this is the most recent version of the code? Note that this is different from https://github.com/e-mission/e-mission-server/pull/826

corinne-hcr:
Yes, the files in this PR are copied from the private data repo. They are what I have and am working on now.

shankari:
> When I am done with refactoring, I will update the codes here.

Note that I plan to significantly refactor this code as part of my cleanup. I would suggest importing directly from this repo instead of making a copy, because you will not be able to simply copy your changes over.

shankari:
You are reading all the user trips **twice** while determining the parameters.
First, in  `get_user_ls` (`emission/analysis/modelling/tour_model/get_users.py`) and second, in `emission/analysis/modelling/tour_model/evaluation_pipeline.py` as part of the `main` function. Not surprising that everything is super slow....

corinne-hcr:
I use `get_user_ls` in the notebook code, but it runs fast.
In the clustering code, around 8 mins for 1 user. Around 50 mins to run all users. The tuning part is slow.

shankari:
@corinne-hcr **please do not make any additional changes**. 

I am refactoring the code much more substantially, and if you push changes as well, I am going to get a bunch of merge conflicts.

I should have taken over the codebase a lot earlier, but I waited for you to push all your changes so that we would not be working on the same codebase at the same time.

At this point, I'm going to have to pull an all-nighter to clean up this code, so please don't make my life harder.

shankari:
@corinne-hcr Please work on the unit tests instead so I can know that my refactoring is correct.

shankari:
At the current point, everything seems to work end-to-end for the minipilot dataset.
@corinne-hcr congratulations! It looks like your code mostly worked; I only had to spend ~ 5 hours and some changes to integrate it. It still needs a lot of cleanup, both in terms of code structure and in terms of performance, but it does work in terms of functionality, so we can get the end-to-end demo working before the meeting.

corinne-hcr:
@shankari 
https://github.com/e-mission/e-mission-server/pull/829/commits/14ced91464d878b7e5733fe2a885f0a4b643f01f cannot pass all the checks
What to do now? Just leave them there?

shankari:
@corinne-hcr this really illustrates why unit tests that test various scenarios are important. Although my code for the random estimator worked with the stage dataset, it does not work in the corner case where the user does not have any user inputs.
This code https://github.com/e-mission/e-mission-server/pull/829/commits/f254af9eafe29be59e6ca4ac8e23d77807dccd8d#diff-eb55ab891674bbd78d9f4cac01c36da267e4138ec4b96c2a193cb7815ad5fa4aR137 fails with the error

```
    random_user_input = random.choice(unique_user_inputs) if random.randrange(0,10) > 0 else []
  File "/home/runner/miniconda-4.8.3/envs/emissiontest/lib/python3.7/random.py", line 261, in choice
    raise IndexError('Cannot choose from an empty sequence') from None
IndexError: Cannot choose from an empty sequence
```

Bad news: I did not add unit tests for the placeholder since I wrote it in 2 hours.
Good news: Gabriel's unit test did catch it

This is also an argument for automated tests since if we had merged this to master, it would have bitten us later.
We don't have automated tests on your fork, so my original PR didn't run the automated tests.
Fortunately, we do have automated tests as part of the e-mission-server repo, so we ran them now and caught before merging to master.

I will fix this as part of my next set of changes.

corinne-hcr:
In my previous work, I filtered out data that didn't have user inputs. Is it the reason that I didn't hit the check?
Is it because of not enough user inputs, my previous model building code only worked for 3 users?

shankari:
there are two questions here:
- why did you not hit the check? You would not have hit the check until you ran the tests with https://github.com/e-mission/e-mission-server/pull/829/commits/f254af9eafe29be59e6ca4ac8e23d77807dccd8d. Did you run the tests earlier?
- why did the model building only work for 3 users? Very likely because of low labeling %. Note that with the more complex expectation code that @GabrielKS wrote, we will eventually have a lower and lower percentage of trips manually labeled, since as we have higher confidence, we will stop requesting labels. @GabrielKS do I understand your proposal correctly?
So as part of the expectation pipeline changes, I will stop that validity checking and I will also include the trips below the cutoff in case they match better.

corinne-hcr:
No, I didn't. I just know the new design by reading that commit. I don't know how to run the test before merging. 
The new design looks completely different from previous one.

shankari:
ok so that explains why you haven't seen the failure earlier. Although the similarity code has been rewritten, it will still work the old way. If you see anything that is not backwards compatible, please flag it!

shankari:
@corinne-hcr are you saving the user labels as a dataframe? The saved output looks like

```
[{"0": [{"0": [{"labels": {"mode_confirm": "shared_ride", "purpose_confirm": "home", "replaced_mode": "bus"}, "p": {"py/reduce": [{"py/function": "numpy.core.multiarray.scalar"}, {"py/tuple": [{"py/reduce": [{"py/type": "numpy.dtype"}, {"py/tuple": ["f8", false, true]},
```

corinne-hcr:
Not dataframe, they are dict.
The first 0 is from the 1st round, the second 0 is from the 2nd round. Then return the value of the 2nd 0.

shankari:
> Not dataframe, they are dict.
The first 0 is from the 1st round, the second 0 is from the 2nd round. Then return the value of the 2nd 0.

That is not the way that dicts typically look like when they are serialized. For example, it looks like some functions are getting dumped too `{"py/reduce": [{"py/function": "numpy.core.multiarray.scalar"}`.

```
>>> import json
>>> json.dumps({"0": {"0": {"k1": "v1", "k2": "v2"}}})
'{"0": {"0": {"k1": "v1", "k2": "v2"}}}'
```

When I dump the a dict, it just shows up as a dict without serialized functions

corinne-hcr:
The results in the dict, (e.g. `p` is from a dataframe)
Before I save everything in a dict/list, I use dataframe to find unique labels and p

shankari:
@corinne-hcr I am not asking you this question because I don't know how to read the file. I am perfectly capable of reading files from disk. This is more to make sure that we are using the correct datastructures and methods to serialize/deserialize the file.

Have you tried opening any of the files that are created by the current pipeline? The `locations_*` files are plain text and are what I expect. The `models_*` files are pickled outputs, which is what I would expect as well. But the `user_labels_*` files are also what appear to be pickled outputs and I am not sure why they need to be pickled.

corinne-hcr:
@shankari it is not successful. Is it affected by the previous PR?

shankari:
If you look at the logs, all the 

```
    random_user_input = random.choice(unique_user_inputs) if random.randrange(0,10) > 0 else []
  File "/home/runner/miniconda-4.8.3/envs/emissiontest/lib/python3.7/random.py", line 261, in choice
    raise IndexError('Cannot choose from an empty sequence') from None
IndexError: Cannot choose from an empty sequence
```

errors are gone, both because of https://github.com/e-mission/e-mission-server/pull/829/commits/f2060ccfb3aecceef027031a42e0f8367d3c2618 which handled the corner case, and because of https://github.com/e-mission/e-mission-server/pull/829/commits/e666d2293b860ffcfa2d7d0676a7e71c4e299260 which removed the random placeholder.

There is still failure due to 

```
FAIL: testGetUUIDList (storageTests.TestTimeSeries.TestTimeSeries)
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/storageTests/TestTimeSeries.py", line 39, in testGetUUIDList
    self.assertIn(self.testUUID, uuid_list)
AssertionError: UUID('25c227ba-fefe-4c5f-9b01-f5bbcf1617e7') not found in []

```

Since that is in the storage tests, it must be because of https://github.com/e-mission/e-mission-server/pull/829/commits/7fac45254c602d50def8f17c0d06c04c15764c81

Let me take a look

shankari:
We currently have hard-coded radius and other parameters for the first round.

So the instructions for generating labels are:
- generate the model files: `./e-mission-py.bash emission/analysis/modelling/tour_model_first_only/build_save_model.py`
  -  run one time
  - only re-run if labels have changed
- reset the inference pipeline:  `./e-mission-py.bash bin/debug/reset_partial_label_testing.py -i`
- reset the confirmed trips pipeline:  `./e-mission-py.bash bin/debug/reset_partial_label_testing.py -c`
- re-run the intake pipeline: `./e-mission-py.bash bin/intake_multiprocess.py 3`

shankari:
@GabrielKS I have now enabled the `same_mode` to actual mode mapping to fix https://github.com/e-mission/e-mission-docs/issues/656#issuecomment-892201488

Can you see if this makes a difference?

shankari:
Deployed this to staging.
- Rebuilt the models (`./e-mission-py.bash bin/build_label_model.py -a > /var/log/better_model.log 2>&1`)
- Confirmed that all the model files had been updated (timestamp = `Aug  5 05:52`)
- Reset the inference trips
```
./e-mission-py.bash bin/debug/reset_partial_label_testing.py -i
```

Note that because of the large number of expected trips, the deletion is taking 2 minutes and involves multiple thousand expected trips. 

```
{'n': 5922, 'ok': 1.0}
{'n': 1094421, 'ok': 1.0}
{'n': 5922, 'ok': 1.0}
{'n': 5922, 'ok': 1.0}
{'n': 78, 'ok': 1.0}
```

- By this time, the regular pipeline run had started, so waited until it was complete, and re-ran the reset

```
./e-mission-py.bash bin/debug/reset_partial_label_testing.py -i
{'n': 3872, 'ok': 1.0}
{'n': 3872, 'ok': 1.0}
{'n': 3872, 'ok': 1.0}
{'n': 3872, 'ok': 1.0}
{'n': 38, 'ok': 1.0}
./e-mission-py.bash bin/debug/reset_partial_label_testing.py -c
{'n': 0, 'ok': 1.0}
{'n': 5922, 'ok': 1.0}
{'n': 39, 'ok': 1.0}
```

- Re-ran the pipeline

- Checking for one user; everything looks good

```
all inferred trips 264
all confirmed trips 264
confirmed trips with inferred labels 150
confirmed trips without inferred labels 114
confirmed trips with expectation 264
```

PR #830, opened 2021-08-09
Implementation for the survey_response syncing.

jruzekowicz:
Following the implementation referenced here: https://github.com/e-mission/e-mission-server/commit/ecd825282db86cf074645a0201d9bc39c636cfb0

I used the same process and testing will be done when the other end (phone side) is fully implemented.


emission/core/wrapper/entry.py
```diff
@@ -73,6 +73,8 @@ def _getData2Wrapper():
             "manual/mode_confirm": "userlabel",
             # user confirmation of the travel purpose
             "manual/purpose_confirm": "userlabel",
+            #user response to the survey questions
+            "manual/survey_response": "userlabel",
```
shankari, line 77: I think that @asiripanich uses `manual/survey_confirm` instead. Maybe we could come up with a standardized nomenclature. Also, should we have separate labels for the initial demographic survey and the trip-level surveys?

shankari (approved):
LGTM overall. Will merge after hearing from @asiripanich

emission/core/wrapper/entry.py
jruzekowicz: Okay, so the current nomenclature follows the "statement_confirm" pattern? I can switch to that. 
What do you mean by separate labels for this?

shankari: @jruzekowicz don't change to it until we hear from @asiripanich

>  What do you mean by separate labels for this?

Right now, we are only displaying a survey when the user starts the onboarding process. We might want to have trip-level surveys, similar to https://nextcloud.damajash.org/s/CRZg4zQY2gKWMnF as well, in which case, we may want to have a different key for that.

You don't need to make that different key right now, but we may want to call this `manual/demo_survey_confirm` or something to distinguish from trip-level surveys.

asiripanich: Hi Shankari, personally I don't mind `survery_response`. We do the initial survey differently than you do. We use an in-app browser to open an online survey form instead of using Enketo. However, for the trip-level surveys, we use Enketo because it allows us to capture the responses and use them to display on the UI and the responses to be edited without internet.

jruzekowicz: @shankari Would it be better suited to call this something such as "Itinerum_survey_response"?

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, user_id):
                 "manual/incident": self.timeseries_db,
                 "manual/mode_confirm": self.timeseries_db,
                 "manual/purpose_confirm": self.timeseries_db,
+                "manual/survey_response": self.timeseries_db,
```
asiripanich, line 62: @atton16, it seems like @shankari's team is going with `survey_response`. Maybe it is what we should use as well for our Enketo plugin?

atton16: There is a "manual/survey" in `gis-based-mode-detection` branch though. So ???

atton16: But this one works for me anyways.

PR #831, opened 2021-08-10
Get confidenceThreshold from config file

GabrielKS:
Fixes https://github.com/e-mission/e-mission-docs/issues/661. 

Put confidenceThreshold into expected trips
+ Add a new field to the expected and confirmed trips for it
 + Populate it when we process the expectations
 + Add a test to `TestExpectationPipeline` to check for the existence of `confidenceThreshold`
 + Fix sneaky time zone-related bug in `expectation_notification_config`
 + Edit `TestExpectationNotificationConfig` so that it would have caught that bug

End-to-end testing:
 1. Use server and client branches with only mode and purpose
 2. Enable placeholder predictor 3
 3. Use an expectations config that has a relaxed mode with a `confidenceThreshold` < 0.80 and an intensive mode with a `confidenceThreshold` >= 0.80 with a schedule beginning at `2015-07-22T16:00:00` local time
 2. Reset pipeline; load and intake the `shankari_2015-07-22` `test_july_22` test dataset
 4. Ensure on the client that the second most recent trip has all yellow labels and that the most recent trip has all red labels

### DIFF
--- a/emission/tests/pipelineTests/TestExpectationPipeline.py
+++ b/emission/tests/pipelineTests/TestExpectationPipeline.py
@@ -26,6 +26,11 @@ class TestExpectationPipeline(unittest.TestCase):
         960: arrow.get("2023-02-12T20:00:00.000", tzinfo=tz) 
     }
 
+    @staticmethod
+    def fingerprint(trip):
+        # See eacilp.placeholder_predictor_2 for an explanation of the "fingerprint" technique
+        return trip["data"]["start_local_dt"]["hour"]*60+trip["data"]["start_local_dt"]["minute"]
+
     def setUp(self):
         self.test_options_stash = eace._test_options
         eace._test_options = {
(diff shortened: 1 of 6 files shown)


emission/tests/pipelineTests/TestExpectationPipeline.py
```diff
@@ -100,13 +110,17 @@ def testProcessedAgainstRaw(self):
...
             elif raw_expectation["type"] == "all":
-                self.assertEqual(trip["data"]["expectation"], {"to_label": True})
+                self.assertEqual(trip["data"]["expectation"]["to_label"], True)
             else:
                 print("Expectation behavior for "+str(raw_expectation)+" has not been implemented yet; not testing. Value is "+str(trip["data"]["expectation"]))
             # TODO: implement tests for the other configurable expectation types once they've been implemented
 
+    def testConfidenceThreshold(self):
+        for trip in self.expected_trips:
+            self.assertTrue("confidence_threshold" in trip["data"])
```
shankari, line 122: Shouldn't you have a tighter check than this? You know what the `confidence_threshold` should be, why not check for it directly?

shankari (commented):
Makes sense overall, just a comment about a tighter check...

emission/tests/pipelineTests/TestExpectationPipeline.py
shankari: I would like to see some actual values even if they are all the same

shankari (approved):

shankari:
I just deployed this. I am not going to reset the pipeline this time, just allow the new trips to get the better labels.
Objections?

PR #832, opened 2021-08-17
Add documentation for new Label UI

GabrielKS:
Addresses https://github.com/e-mission/e-mission-docs/issues/666. Draft; don't merge yet.


shankari:
This looks great! As you can see, writing documentation takes significant time and effort. At least when I was in the tech industry (which is ~ 10 years ago now), we used to have tech writers who specialized in documentation. Scrappy small teams, however, don't have that luxury.

shankari:
I'm merging and deploying this now. Again, we can tweak this later if we really want to.

shankari:
Deployed to https://stage.canbikeco.org/faq/hybrid_labeling.html

PR #833, opened 2021-08-19
Add a very high-level technical overview + get ready to make the colu…

shankari:
…mns side-by-side

Add a very high-level technical overview - targeted at a program admin level -
along with a brief explanation of the "intensive" and "relaxed" modes and why
they are necessary.

+ get ready to put the columns side by side by adding `col` `<div`s within the
`row`. However, to actually make them appear side by side, you need an
additional `col` class.

e.g.

```
<div class="col-md-33 mx-auto" style="text-align: left;">
```

to

```
<div class="col col-md-33 mx-auto" style="text-align: left;">
```

Implementing that, however, makes the images appear side by side, and the
layout is not really responsive until the width is unreasonably small. It would
not convert to a vertical layout on an iPhone6, for example.

So we have backed out that change until we have a chance to look at settings
for this CSS template.


shankari:
@GabrielKS would you like to review this?

webapp/www/faq/hybrid_labeling.html
```diff
@@ -153,6 +159,16 @@ <h4>
                         The Diary screen has not changed at all. None of the new features (e.g., the yellow labels, the confirm button, only seeing trips you need to label) work on the Diary screen. So we recommend you do your labeling on the Label screen!
                     </h4>
                     <p></p>
+                    <h3>
+                        <strong>I love algorithms! Can I have more details?</strong>
+                    </h3>
+                    <h4>
+                        We group all your data so that trips with the same start and end location are in the same bin. We then look at the labels for the trips in each bin and how often they occur. If a trip has (shared_ride,work,drove_alone) labels 60% of the time, and (drove_alone,work,shared_ride) 40% of the time, we will display the highest probability label - e.g. (shared_ride,work,drove_alone) as a "yellow" "To Label". If it occurs 100% of the time, though, after the first few times, we don't even need to put it into "To Label" because we are pretty sure that is right.
```
GabrielKS, line 166:
```suggestion
                        We group all your data so that trips with the same start and end location are in the same group. We then look at the labels for the trips in each group and how often they occur. If a trip has (shared_ride,work,drove_alone) labels 60% of the time, and (drove_alone,work,shared_ride) 40% of the time, we will display the highest probability labels — i.e., (shared_ride,work,drove_alone) — as yellow in "To Label." If the highest probability labels occur closer to 100% of the time, after a while, we don't even need to put them into "To Label" because we are so sure that they are correct.
```

webapp/www/faq/hybrid_labeling.html
```diff
@@ -153,6 +159,16 @@ <h4>
...
+                        But what if you used to carpool to work consistently, but now you started riding your spiffy new ebike? To detect such changes, we will periodically go from the usual "relaxed" mode to an "intensive" mode, where all your trips will show up in "To Label" for a week.
```
GabrielKS, line 167:
```suggestion
                        But what if you used to carpool to work consistently, but now you started riding your spiffy new eBike? To detect such changes, we will periodically go from the usual "relaxed" mode to an "intensive" mode, where all your trips will show up in "To Label" for a week.
```
Nitpick: I've been using `e-bike`; this website seems to sometimes use `e-bike` and sometimes use `eBike`  — we should standardize on something but it doesn't seem to be `ebike`.

webapp/www/faq/hybrid_labeling.html
```diff
@@ -153,6 +159,16 @@ <h4>
...
+                        To summarize, as your travel patterns change - you switch jobs or modes or take up a new hobby - your trip locations and probabilities will change. But as long as we can get the correct labels periodically, the algorithm will update itself with the new probabilities and continue making your life easier.
```
GabrielKS, line 168:
```suggestion
                        To summarize, as your travel patterns change — you switch jobs or modes or take up a new hobby — your trip locations and probabilities will change. But as long as we can get the correct labels periodically, the algorithm will update itself with the new probabilities and continue making your life easier.
```
Em dashes! No pun intended. They appear the same in this monospaced font, but on the website they will appear longer.

GabrielKS (changes requested):
Looks good overall. I left some minor rewording and proofreading suggestions.

PR #834, opened 2021-08-19
Fix expected trip repetition bug, add test to prevent it happening again

GabrielKS:
Fixes https://github.com/e-mission/e-mission-docs/issues/654

### DIFF
--- a/emission/tests/pipelineTests/TestPipelineStageNonrepetition.py
+++ b/emission/tests/pipelineTests/TestPipelineStageNonrepetition.py
@@ -13,7 +13,7 @@ import emission.analysis.classification.inference.labels.inferrers as eacili
 import emission.analysis.configs.expectation_notification_config as eace
 import emission.analysis.userinput.expectations as eaue
 
-class TestExpectationPipeline(unittest.TestCase):
+class TestPipelineStageNonrepetition(unittest.TestCase):
     # The limitation of this approach is that one must manually add database keys here.
     # So when implementing a new pipeline stage, put the relevant key(s) here!
     # I've tried to go through the various existing pipeline stages and put some of their keys here, but this may already be incomplete.
@@ -63,7 +63,7 @@ class TestExpectationPipeline(unittest.TestCase):
         counts = {}
         db = edb.get_analysis_timeseries_db()
         for key in self.keys_to_track:
-            count = db.find({"metadata.key": key, "user_id": self.testUUID}).count()
+            count = db.count_documents({"metadata.key": key, "user_id": self.testUUID})
             counts[key] = count
         # print(counts)
         return counts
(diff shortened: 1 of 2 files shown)


emission/tests/pipelineTests/TestPipelineStageNonrepetition.py
```diff
@@ -0,0 +1,83 @@
...
+        eaue._test_options = test_options_stash
+
+    def reset_all(self):
+        etc.dropAllCollections(edb._get_current_db())
+    
+    def count_keys(self):
+        counts = {}
+        db = edb.get_analysis_timeseries_db()
+        for key in self.keys_to_track:
+            count = db.find({"metadata.key": key, "user_id": self.testUUID}).count()
```
shankari, line 66: I believe that `count()` is deprecated and replaced by `countDocuments()`

shankari, line 66:
```suggestion
            count = db.countDocuments({"metadata.key": key, "user_id": self.testUUID})
```

shankari (changes requested):
Can you add some information on what the problem as and how this fixes the it? From the change, it looks like we would have got `expected_trip = None`. So when we tried to access `expected_trip["data"]["end_ts"]`, we would have got an exception.

But that code is outside the `try/expect` block so it would not be handled. So we should have logged that exception and generated an error. But I didn't see any errors in the intake stdinout. What am I missing here?

GabrielKS:
> Can you add some information on what the problem as and how this fixes the it? From the change, it looks like we would have got `expected_trip = None`. So when we tried to access `expected_trip["data"]["end_ts"]`, we would have got an exception.
> 
> But that code is outside the `try/expect` block so it would not be handled. So we should have logged that exception and generated an error. But I didn't see any errors in the intake stdinout. What am I missing here?

The bug was that `_process_and_save_trip` did not return a trip. This meant that `expected_trip` from line 23 was always `None` and thus `last_trip_done` from line 17 was always `None`. So the check on line 24 always lazily evaluated to `True` and we never queried `expected_trip["data"]["end_ts"]` to get that error. When we passed `None` as a trip to `mark_expectation_population_done`, it internally handled that as a special case (per the boilerplate I copypasted from the other `mark_..._done`s), so we didn't get an error there either.

emission/tests/pipelineTests/TestPipelineStageNonrepetition.py
GabrielKS: You're right (though it's `count_documents()`). Fixed now.

shankari (approved):

shankari:
Restating https://github.com/e-mission/e-mission-server/pull/834#issuecomment-902341047 in my own words to help me recollect it later:

Without the fix, the code is essentially:

```
>>> last_trip_done = None
>>> expected_trip = None
>>> if last_trip_done is None or last_trip_done["data"]["end_ts"] < expected_trip["data"]["end_ts"]:
...     print(True)
True
```

As we can see, it does not error out due to the short circuit. Then, instead of the `print(True)`, `last_trip_done` remains None. So this code gets triggered:

```
        if last_trip_done is None:
            logging.debug("After run, last_trip_done == None, must be early return")
        epq.mark_expectation_population_done(user_id, last_trip_done)
```

PR #843, opened 2021-12-11
Fix median_speed call when returning confirmed trips

shankari:
In https://github.com/e-mission/e-mission-server/pull/841
we added support for returning metrics based on confirmed trips.
This works for all the metrics that we support, except for median_speed, since
the trips don't have a list of speeds.

I first tried to fix this by adding a list of speeds to each trip, but was
worried about breaking the `ground_truth` comparison tests.

So for now, if we are returning `confirmed_trips`, and there is no `speed`
field, we concatenate the speeds lists from the sections to generate the speed
list for the trip.

+ additional pandas test to check the concatenation
+ additional metrics test that retrieves all kinds of metrics and ensures that
    we don't have a similar issue in the future


shankari:
Fixes https://github.com/e-mission/e-mission-docs/issues/688#issuecomment-991392462

PR #848, opened 2022-03-22
Create a script that exports user travel diaries as geojson

shankari:
This reuses the same method as the the code to generate the diary for a
particular day for display in the UI. The related

For an example of working with the data, please see
https://github.com/e-mission/e-mission-eval-private-data/blob/master/OpenPATH_geojson_example.ipynb

@PatGendre in case this helps

### DIFF
--- a/bin/debug/extract_geojson_for_time_range_and_user.py
+++ b/bin/debug/extract_geojson_for_time_range_and_user.py
@@ -1,0 +1,90 @@
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+from __future__ import absolute_import
+# Exports all data for the particular group of users for the particular day range
+# Exports them as geojson using the `emission.analysis.plotting.geojson.geojson_feature_converter` method
+# Note that you probably want to edit `conf/analysis/debug.conf.json.sample` to
+# turn off `output.conversion.validityAssertions` to avoid running into asserts
+# while generating the geojson files
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *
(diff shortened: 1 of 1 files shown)


PatGendre:
@shankari thanks !   
just a question about this script : it uses [geojson_feature_converter](https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/plotting/geojson/geojson_feature_converter.py), and it is not completely clear which objects are exported.
I understand inferred trips and sections are exported, so the predicted mode / purpose computed later on the in pipeline is not exported, right ? Only the sensed_mode seems to be exported.      
Also, I don't see if (and where) the manual/mode_confirm and manual/purpose_confirm are exported.

I ask this because the same code is used by the getTrips API which we use for transferring data to personal cozy clouds.

shankari:
@PatGendre yes, this intentionally reuses the getTrips API. As part of a complete rewrite of the label display functionality in the UI, which included scalability improvements, I made server side changes to work off the confirmed trips instead. This means that the trip labels are indeed incorporated (at the trip level). The sensed modes continue to be exported at the sensed level.

https://github.com/e-mission/e-mission-server/commit/78a3ebb612077bea923d922e6cdfbca4ccfa1921

shankari:
Note that I didn't quite get to the point where the user input is a generic JSON structure, but the UI directive is much simplified

PatGendre:
> This means that the trip labels are indeed incorporated (at the trip level). The sensed modes continue to be exported at the sensed level.

Ok, fine, thanks !

PR #849, opened 2022-05-10
Fix indexes

aGuttman:
Change HASHED to ASCENDING and removed sparse from GEOSPHERE for documentdb compatibility


aGuttman:
Fixes one part of https://github.com/e-mission/e-mission-docs/issues/721

emission/core/get_database.py
```diff
@@ -21,8 +21,20 @@
...
+except:
+    print("URL not formatted, defaulting to \"Stage_database\"")
+    db_name = "Stage_database"
+else:
+    if parsed['database']:
+        db_name = parsed['database']
+    else:
+        print("URL does not specify a DB name, defaulting to \"Stage_database\"")
+        db_name = "Stage_database"
+
```
shankari, line 35: @aGuttman This looks like it will work, but there are also additional locations where we use `Stage_database`
https://github.com/e-mission/e-mission-docs/issues/721#issuecomment-1122684668

Did you have a chance to go through at least the important ones (primarily `abstract_timeseries.py`) to ensure that they don't need to be fixed?

aGuttman: `abstract_timeseries.py` includes "Stage_database" in `_reset_url()`, a function which seems to never be used:
```
grep -r -l "_reset_url" .
./emission/storage/timeseries/abstract_timeseries.py
./emission/storage/timeseries/__pycache__/abstract_timeseries.cpython-37.pyc
./htmlcov/d_9b9568736b288f26_abstract_timeseries_py.html
```
`_reset_url()` is defined in `abstract_timeseries.py` but never called inside of it. The other two are artifacts of some kind, from compiling and from coverage testing.

For the db_auth/TestMongodbAuth, these are stand alone tests not run in the unit tests that are hardcoded to localhost. I can update those as well to use the config url, but it seemed unnecessary to worry about them at the time, since the unit tests were passing.

Probably should have asked about these at the time. Is it worth it to update db_auth/TestMongodbAuth? Does not make sense to you that _reset_url() is never called, or am I missing something wrt it?

shankari: @aGuttman as you can see in the comment, I added _reset_url() as a hack to allow us to connect to multiple databases from a single client instance, so that we could perform analyses over multiple instances at the same time - see https://github.com/e-mission/e-mission-eval-private-data/blob/d9c8f3fa6d2d6c307e38edcfa4e123f4cb8a96ac/tour_model_eval/Federating%20and%20saving%20multiple%20datasets.ipynb

It turns out that we can just `mongorestore` all of them into the same DB instance. So this should not be needed.

At the same time, some level of federation, so that we can have a public dashboard that spans the various CEO locations, for example, would be good. But we should explore the standard mongodb federation functionality (if any) first.

So I am open to removing this code, as long as you file an issue to explore federation over the long term.

shankari: wrt `db_auth.py` and `TestMongodbAuth.py`, I thought you could modify them to support and test the new solution. 

@aGuttman We don't test DB auth as part of the unit tests, since we don't want to cause dev mongodb installs to become unreachable if the tests fail, but we should test them nevertheless.

aGuttman: Ok. I will open a new issue, remove the old code and update the tests.

shankari:
@aGuttman any ETA for this? I would like to merge https://github.com/e-mission/e-mission-server/pull/853
to finish up the server changes, and don't want to cause merge conflicts with your code.

aGuttman:
The name configuration is easy enough, I can finalize those changes today, but I'm first having trouble getting the test to pass without any changes. I get auth errors when it tries to create the first admin user, which as I understand, shouldn't be happening: https://www.mongodb.com/docs/manual/core/localhost-exception/#std-label-localhost-exception

The test itself seems to be written in a way were it expects this to work. I'm doing this on a fresh db without anything in it, so that shouldn't be the problem. And once this works, I can get the db name to be dynamic, but right now the only way I've been able to make it work is by manually adding a admin user via the mongo shell, connecting with that users credentials and then creating/using the new admin user the test creates. And this only works if you only run one test at a time, adding the admin user manually each time. If this is the intended use (I don't think it is), this test seems a little pointless to configure nicely, since there is so much manual work anyway.

shankari:
@aGuttman have you started mongodb with the `--auth` command?

```
# Implemented outside tests because it requires mongod to be started with the
# --auth opton to pass
# $ mongod --auth
# if mongodb is run without `--auth`, the readonly test is guaranteed to fail
```

aGuttman:
Yes. Without the auth option the users can be created but the `assertRaises()` fail in the tests since there is no access restriction. With auth I can only seem to create the first user in the mongo shell.

shankari:
Hm, interesting. Please file an issue for investigating that as well.

So the next tasks are:
- file an issue for the admin user creation issue
- add a test to `TestMongodbAuth` to test with a different DB as well (e.g. `testReadWriteUserCustomDB`)
  - no additional config, just copy-paste the previous test and make your changes
- file an issue for federating data
  - remove the `_reset_url` code from the timeseries

I think that's it!

shankari (approved):

PR #851, opened 2022-05-22
Remove the code path for the "old style" format

shankari:
We added the "new style" location format in 2015.
We retained the code to process the "old style" format for backwards
compatibility, since the clients would not necessarily be upgraded
immediately.

However:
- all clients should have been updated by now
- we have unpublished that client
- we are creating a new client to talk to the new server

Let's just remove the old style code path since we don't need it any
more, and there's no point in keeping bitrotted code.

This fixes https://github.com/e-mission/e-mission-docs/issues/721#issuecomment-1129249120

Changes:
- change the data from the old format to the new format
- ensure that we load the new format
- remove `format_location_raw` and always return `format_location_simple`

Testing done:
- `emission/tests//netTests/TestBuiltinUserCacheHandlerInput.py` and
- `emission/tests//netTests/TestBuiltinUserCacheHandlerOutput.py` both pass


shankari:
@aGuttman, can you review this fix to remove the obsolete code path, and pull it to your PR before I review it?

aGuttman (approved):
All makes sense to me.

Ran all tests on local mongodb server and TestBuiltinUserCacheHandlerInput.py/Output.py merged with db_compatibility branch on remote documentdb server with no problems.

PR #852, opened 2022-05-23
Rjf/incremental inference

robfitzgerald:
this draft PR is an update to make sure i'm going in the right direction. i'm about 8 hours in and expect 4-6 hours remain. the `tour_model_first_only` module is refactored into two modules with two core abstractions:

```
emission/
  analysis/
    modelling/
      user_label_model/user_label_prediction_model.py
      similarity/similarity_metric.py
```

the existing binning-based heuristic has been refactored into `user_label_model.GreedySimilarityBinning`. `user_label_model.run_model` contains methods to expose to the inference and build_model phases. a future dev would 1) create new derived instances of UserLabelPredictionModel and add their model's default initialization to `run_model._model_factory`.

remaining work:
- [x] write tests with Confirmedtrip data against the binning function, debug, confirm
- [x] read/write to the pipeline to store the incremental timestamp, use as TimeQuery parameter, test + confirm
- [x] wire `run_model.predict_labels_with_n` into the inference code
- [x] wire `run_model.update_user_label_model` into the `bin/build_label_model.py` script
- [x] see about setting various arguments (storage type, model type, distance radius, trip count threshold, model params) via e-mission config
- [x] fill out stubs for load_db + save_db in `user_label_model/util`

the objectives here were to
1. make it easy for future devs to add new prediction models
2. to make it simple for those models to also use incremental data reading
3. to support both file system and database storage (database implementation left to future work)
4. to de-duplicate and dis-ambiguate

i tried to balance what seemed important to keep with the desire to clean up the module. it took me a while to understand what was going on and so i hope i've improved on readability. some notable removals:
- i left in the "elbow heuristic" used for pruning bins, but, can delete that if it's no longer used
- some code in `tour_model.get_request_percentage.requested_trips_bl_cutoff` expects us to have held on to the bins that were removed in the elbow heuristic, but i delete them. if we need to support this, i can fix that


emission/analysis/modelling/user_label_model/run_model.py
```diff
@@ -0,0 +1,201 @@
...
+    model = _model_factory(model_type)
+    model_name = model_type.model_name
+
+    model_data = load(user_id, model_name, model_storage)
+    model.from_dict(model_data)
+
+    return model
+
+
+if __name__ == '__main__':
```
robfitzgerald, line 163: oh, these were the test cases at the bottom of the `tour_model_first_only/load_predict` file, though i'm not sure what dataset they were based on. at least not the dataset documented [here](https://github.com/e-mission/e-mission-server#loading-test-data)

emission/analysis/modelling/user_label_model/util.py
```diff
@@ -0,0 +1,130 @@
...
+    { "user_id": user_id, "data": model_data }
+
+    :param user_id: the user to store data for
+    :type user_id: object
+    :param table: the table name
+    :type table: str
+    :param model_data: the data row to store tagged by this user id
+    :type model_data: Dict
+    """
+    pass
```
shankari, line 98: @robfitzgerald alas, storing to the DB is also a super high priority task at this point. As we work on deploying this in the NREL cloud environment (https://github.com/e-mission/e-mission-docs/issues/721), cloud services would prefer to run the modeling/inference tasks as stateless scheduled tasks. So we cannot store the model to disk, since there is no persistent disk.

We did consider storing the models to S3, but that would then require an S3 bucket approved for moderate use, which would, in turn require additional cyber approval.

Since the database is already approved for moderate use, and as a NoSQL database, mongodb supports document storage, it seemed easiest to store in the database.

robfitzgerald: this is fine, i can work with Andrew on supporting db storage then. but i also assume we still want to support file system storage for non-NRELian users on less-restrictive deployment environments or for testing. i mean, why not, it's already in there 🤷

robfitzgerald: ok, sounds good, i will remove the FS-related utils and we can make getting the database read/write done within this PR. as long as you're ok with letting go of backwards compatibility here. i'll reach out to you if i have questions about writing to tables (and naming conventions and schemas, but, feel free to propose those here if you have ideas beyond my initial assumptions).

shankari: @robfitzgerald Backwards compatibility is not a huge concern right now.

At this point, the system regenerates the model from scratch every time. And when it is deployed on the NREL OpenPATH servers, it will not have any file based models to read from anyway. 

If we want to be extra nice, we could fallback to reading from file if there is no entry in the database, and then remove the fallback code after a year. That will ensure that if somebody else had been running this modeling pipeline (not aware of anybody other than CanBikeCO at this point), they can continue to use the old model to label incoming trips while the new model was being built. But there is no reason at this point to every write to the file system.

Again, not aware of anybody other than CanBikeCO who has enabled the new pipeline, so we could also just punt on the backwards compatibility if it is faster.

wrt naming conventions and schema, I have created two main collections and am storing all data in them. `Stage_timeseries` accessed via `emission.core.get_database.get_stage_timeseries()` and `Stage_analysis_timeseries` accessed via `emission.core.get_database.get_analysis_timeseries()`.

Thinking of creating similar `Stage_models_timeseries` and `Stage_results_timeseries`.

In the raw and analysis timeseries, the entries follow the same basic format with data and metadata entries, but with the fields in data being different based on `metadata.key`. This allows us to have generic metadata based queries based on `metadata.write_ts` across multiple types of data.

Not sure if the models need to be as time based, or whether they should just be UUID based. Thoughts?

robfitzgerald: so, given your convention of maintaining immutable database entries, we could do that with the label models, and it would just mean that we take the one with the most recent write_ts entry when we need to run inference. but i'm not sure how large different models can get. i'm guessing not huge, as the datasets for each user are not huge. if that sounds sensible, we could adopt the convention of immutable records and grab the latest each time.

there's probably not a lot of value in storing all of those models, 365 models per year per user. so, the other thing we could do is assume to store these rows as mutable ones, and really only store 1 row per UUID. in that case, `metadata.write_ts` becomes less a query key and more a housekeeping field. but that would assume that UUID-based indexing is not slow in `Stage_analysis_timeseries` (or maybe motivation to create another collection configured for UUID indexing and designed/labeled for mutable DB operations).

as for the fs discussion, i'm not trying to advocate for file saving i'm just being careful that's all. if it's not helping anyone we can remove that stuff. but it's already implemented so it's not "faster" or "slower" to keep that functionality.

robfitzgerald:
i was dreading this message, knowing full well i promised results by now. i would like to try and wrap this up next week, if that's not too late for you. i'm dealing with a few tasks at work that have been taking much more of my time than anticipated (and that i've had trouble estimating my time for because they are outside of my comfort zone, honestly).

emission/analysis/modelling/user_label_model/util.py
shankari: I think we should start with an immutable object, but just keep the last `n` entries, with `n=3`. This should strike a good balance between robustness and disk space. We will then need a cleanup cronjob/schedule script that will go through and delete older entries.

in the long run, we should remove it because it is obsolete and I don't want to have a bunch of vestigial code sitting around either being maintained or waiting to bitrot.

in the short term, reading from DB with fallback to file seems like a bit more work than reading only from the DB.

robfitzgerald:
@shankari i've spent a little time today reviewing the database queries. i'm looking for how best to record models and retrieve them. two ideas:
- when i need the model, i retrieve the one with the latest timestamp from the analysis timeseries with matching user_id/key
- when i need the model, i retrieve the one with the timestamp that _matches_ `last_ts_run` in the pipeline state with matching user_id/key

looking for your opinion there.

also, if i did want to implement the former (retrieving the _latest_ entry with a matching user_id/key), would we need to add that capability to BuiltinTimeSeries via some new method?

robfitzgerald:
update here. database read/write is wired into the save functions along with pipeline updates. what trip data is loaded depends on whether the Model type reports itself as "incremental" or not.

wanted to confirm how this should be wired in. at this point i have simply swapped in the relevant methods in inferrers.py and build_save_model.py. i haven't tested anything, wanted to confirm this design looks correct, hoping to make time to test the internals though i'm a little fuzzy on running e2e tests here.

emission/analysis/modelling/user_label_model/model_storage.py
```diff
@@ -0,0 +1,129 @@
...
+        filename = create_filename(user_id, model_type)
+        model_data = util.load_fs(filename)
+        return model_data
+    elif model_storage == ModelStorage.DATABASE:
+        
+        # retrieve stored model with timestamp that matches/exceeds the most
+        # recent PipelineState.USER_LABEL_MODEL entry        
+        ts = esda.get_timeseries_for_user(user_id)
+        if not isinstance(ts, BuiltinTimeSeries):
+            raise Exception('user model storage requires BuiltInTimeSeries')
```
shankari, line 45: as discussed, we don't want to have the filesystem code around for the long term.
It is simply an artifact of the implementer not knowing how to work with the database.
We will always have a database to store the raw objects, so the file system support is unnecessary complexity.

You should remove the filesystem option and remove all the related config flags.
If you really want to support backwards compat, you should fallback to reading from the file in the one implementation.
https://github.com/e-mission/e-mission-server/pull/852#discussion_r895077533

emission/analysis/modelling/user_label_model/util.py
```diff
@@ -0,0 +1,96 @@
...
+    line from the first to the last point of the curve and then find the data point
+    that is farthest away from that line."
+
+    :param values: list of values from which to select a cut-off
+    :type values: List[float]
+    :return: the index and value to use as a cutoff
+    :rtype: Tuple[int, float]
+    """
+    N = len(values)
+    x = list(range(N))
```
shankari, line 81: Am I right in assuming that this implementation is basically a copy-paste of the original implementation? If so, please add a permalink to the original code since copy/paste loses the commit history, and we may want to go back and figure out the reason for particular design decisions

shankari (commented):
First round of comments below.

Two high level questions:
- it looks like you copied over the original code into new locations and changed the command line signatures. Did you change the implementation as well? If not, please indicate so, and add a permalink to the original code. That helps with review (specially since we don't have dedicated unit tests for this code section), and also helps us find the original commit history if we want to trace back the reason for a particular section of code.
- I am not sure we should call this a user mode. As you may recall from the project with Venu last year, a user model is typically a model of user behavior based on their preferences - e.g. what weight do they place on cost vs. time vs. carbon? This kind of model of their common trips is called a tour model.
It's a small thing, but just fixing the naming may help others navigate the code in the future.

emission/analysis/classification/inference/labels/inferrers.py
```diff
@@ -140,7 +143,9 @@ def n_to_confidence_coeff(n, max_confidence=None, first_confidence=None, confide
 
 # predict_two_stage_bin_cluster but with the above reduction in confidence
 def predict_cluster_confidence_discounting(trip, max_confidence=None, first_confidence=None, confidence_multiplier=None):
-    labels, n = lp.predict_labels_with_n(trip)
+    model_type = ModelType.GREEDY_SIMILARITY_BINNING
+    model_storage = ModelStorage.DATABASE
+    labels, n = eamur.predict_labels_with_n(trip, model_type, model_storage)
```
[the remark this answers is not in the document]
robfitzgerald: the "n" is the number of rows of data that are associated with this prediction. that was the name and signature of the prediction function in the original code [here](https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/modelling/tour_model_first_only/load_predict.py#L66). because you told me not to mess with anything outside of tour_model_first_only, i matched the signature and only changed which function was being called here.

user_label_model_greedy_c46b0f38-fc47-4e09-a8de-f15f382f7121
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
robfitzgerald: i don't know, i like the UUID a lot, might name my next child "a8de f15zg3ra7d" in fact

emission/analysis/modelling/similarity/confirmed_trip_feature_extraction.py
```diff
@@ -0,0 +1,28 @@
...
+def od_features(trip: Confirmedtrip) -> List[float]:
+    o_lat, o_lon = origin_features(trip)
+    d_lat, d_lon = destination_features(trip)
+    return [o_lat, o_lon, d_lat, d_lon]
+
+def distance_feature(trip: Confirmedtrip) -> List[float]:
+    return [trip.data.distance]
+
+def duration_feature(trip: Confirmedtrip) -> List[float]:
+    return [trip.data.duration]
```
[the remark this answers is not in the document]
robfitzgerald: sure thing.

emission/analysis/modelling/user_label_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,210 @@
...
+        features = self.metric.extract_features(trip)
+        return features
+
+    def _assign_bins(self, trips: List[Confirmedtrip]):
+        """
+        assigns each trip to a bin by greedy similarity search
+
+        :param data: trips to assign to bins
+        :type data: List[Confirmedtrip]
+        """
```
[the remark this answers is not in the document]
robfitzgerald: i'm new enough to this game to ask for clarification. is it simply that you would like a code comment with a link to the location in the original code? in this case, adding a line like:

```python
# https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/modelling/tour_model/similarity.py#L118
```

emission/analysis/modelling/user_label_model/util.py
robfitzgerald: sure thing. 

but, actually, this is actually code that was completely deprecated in the current version (the branch that would activate this code was hard-coded to false). i could simply delete it, but i wanted to keep it around in case you wanted to ever reactivate this heuristic.

robfitzgerald:
> it looks like you copied over the original code into new locations and changed the command line signatures. 

i did, if you mean function signatures here.

> Did you change the implementation as well? 

yes in some parts. in order to create an abstraction for future devs to work with, i needed to fit the existing solution to that abstraction, so a few changes were made for that. also, while porting it, i found duplicate entries for its similarity metric and decided it should be extracted into it's own module and made available to future model implementations. i also found that the existing solution work with lists and then converted to dicts for saving, and changed it so that it simply created the model as a dict in the first place. so, slightly different data structure stored in GreedySimilarityBinning than in tour_model/similarity.

> If not, please indicate so, and add a permalink to the original code. That helps with review (specially since we don't have dedicated unit tests for this code section), and also helps us find the original commit history if we want to trace back the reason for a particular section of code.

i'll tag those copied code segments with URLs to the latest commit (not master), which i think is what you mean by permalinking them.

shankari:
> but i also don't know what's deprecated and what's important to keep in the "tour_model" directory so i'm staying out of it, and i don't think the name "tour_model_first_only" is a good name for the proposed refactor here, as the name disambiguates from some approach to clustering that appears deprecated as well. 

for the record, `tour_model_first_only` is the currently used model. It is called `first_only` because is only uses the first round of clustering. It is `tour_model` that is currently deprecated 😄 `tour_model_first_only` is a copy of `tour_model` with the code around second round of clustering removed.

> i found duplicate entries for its similarity metric and decided it should be extracted into it's own module and 

Duplicate entries? Where?

> i also found that the existing solution work with lists and then converted to dicts for saving, and changed it so that it simply created the model as a dict in the first place.

Again, some comments around this would help with review, again because we don't have any unit tests for this code.

bin/build_label_model.py
```diff
@@ -64,4 +66,7 @@ def _email_2_user_list(email_list):
     logging.info("received list with %s users" % user_list)
     for user_id in user_list:
         logging.info("building model for user %s" % user_id)
-        eamtb.build_user_model(user_id)
+        model_type = ModelType.GREEDY_SIMILARITY_BINNING
+        model_storage = ModelStorage.DATABASE
+        min_trips = 14
```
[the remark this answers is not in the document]
shankari: ugh!

emission/analysis/modelling/user_label_model/run_model.py
```diff
@@ -0,0 +1,194 @@
...
+    user_id = trip['user_id']
+    model = _load_user_label_model(user_id, model_type, model_storage)
+    if model is None:
+        return [], -1
+    else:
+        predictions, n = model.predict(trip)
+        return predictions, n
+
+
+def _get_trips_for_user(user_id, time_query: Optional[TimeQuery]=None, min_trips: int=14):
```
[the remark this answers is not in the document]
shankari: agreed

emission/analysis/modelling/user_label_model/greedy_similarity_binning.py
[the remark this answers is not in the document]
shankari: Yes that is what I had in mind.

emission/analysis/modelling/user_label_model/util.py
shankari: Yes, it was completely deprecated for now because we don't care about focusing on common trips for this implementation. 

However, if we want to show people their common trips and ask them even more details, we probably do want to focus on them. So it is good to keep it around.

bin/build_label_model.py
```diff
@@ -6,12 +6,14 @@
 import argparse
 import uuid
 import copy
+from emission.analysis.modelling.user_label_model.model_storage import ModelStorage
+from emission.analysis.modelling.user_label_model.model_type import ModelType
```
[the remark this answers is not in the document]
shankari: that looks super tedious to implement, but I'm glad you did it! I don't think you got all of them though - this particular location is still unchanged.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, user_id):
                 "metrics/daily_mean_median_speed": self.analysis_timeseries_db,
                 "inference/prediction": self.analysis_timeseries_db,
                 "inference/labels": self.analysis_timeseries_db,
+                "inference/user_label_model": self.analysis_timeseries_db,
```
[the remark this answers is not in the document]
shankari: I was hoping to get this done now instead of having to deal with migrations in the future. The changes should be fairly straightforward since mongodb doesn't have any schemas. You literally need to add a new function and a new collection in `emission.core.get_database`. LMK if you don't want to tackle it and I will.

shankari:
@robfitzgerald let me know when you have finished addressing all the review comments so I can re-review

bin/build_label_model.py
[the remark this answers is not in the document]
robfitzgerald: oop. got it.

robfitzgerald:
> Duplicate entries? Where?

there was duplication of the feature extraction + similarity metric, appearing for both prediction (see [in_bin](https://github.com/e-mission/e-mission-server/blob/03fa115f1211b1c36b7672c8f44be1dffdb33c83/emission/analysis/modelling/tour_model_first_only/load_predict.py#L28)) and model building (see [filter_too_short](https://github.com/e-mission/e-mission-server/blob/03fa115f1211b1c36b7672c8f44be1dffdb33c83/emission/analysis/modelling/tour_model/similarity.py#L44)).

there's a tight coupling between a mapping to features and a similarity metric. hence i combined these into emission.analysis.modelling.similarity.SimilarityMetric so it can be used by future similarity-based models.

robfitzgerald:
> Again, some comments [on how the data structures changed] would help with review, again because we don't have any unit tests for this code.

cool, i'll get on it.

robfitzgerald:
@shankari addressed your additional concerns here. i'm going to now work on writing some tests for this code. things look mostly like e2e tests; do unit tests also go the tests directory? and, should these tests go in `emission.tests.modellingTests` instead of following the same namespacing directories like in java land? and all camelCase?

robfitzgerald:
i added some testing now as well. it includes mocked Confirmedtrip entries. i don't get the impression that you do a lot of mocking in this repo, but, i went ahead with it to confirm at least for myself that the logic is sound (and to catch a few bugs!)

noting here some additional requirements beyond running install documented [here](https://github.com/e-mission/e-mission-docs/blob/master/docs/install/manual_install.md#python-distribution):
```
$ pip install polyline
$ pip install geopy
$ pip install geocoder
```

shankari:
> i'm going to now work on writing some tests for this code.

I am so glad you are doing this. It is such a relief to work with an experienced developer who can do this on his own initiative.

> i don't get the impression that you do a lot of mocking in this repo, 

We do use some mocking - e.g. in the most recent tests that I added in `emission/tests/storageTests/TestTripQueries.py`
But it is true that we also do a large amount of testing on real-world day snippets that I used for tuning the algorithms to begin with.

> noting here some additional requirements beyond running install documented [here](https://github.com/e-mission/e-mission-docs/blob/master/docs/install/manual_install.md#python-distribution):

Please add these to `setup/environment36.yml` so that they are also installed as part of setup and the automated tests run. I am on PTO tomorrow, but will review (and hopefully merge) on Monday.

robfitzgerald:
i seem to be getting something wrong about pipeline states. after training my model and writing it to the database, i attempt to set this pipeline stage as done (`mark_stage_done(user_id, MY_NEW_PIPELINE_STATE, my_timestamp)`). but using a test database with no pipeline data in it, there is no "current pipeline state" and so `mark_stage_done` blows up with `assert(curr_state is not None)`.

is there a one-time pipeline state initialization i need to execute? or am i in some other way misunderstanding what's going on?

shankari:
@robfitzgerald that is because before marking the stage as done, the pipeline framework expects you to `get_time_range_for_stage` to retrieve the set of relevant trips.

Also, for my edification, is this change only refactoring/writing to a different location, or also supporting incremental updates to the model?

robfitzgerald:
> the pipeline framework expects you to `get_time_range_for_stage` to retrieve the set of relevant trips

ah, ok, i'll make changes accordingly

> Also, for my edification, is this change only refactoring/writing to a different location, or also supporting incremental updates to the model?

also supports incremental updates. all of the code is set so that when someone creates a TripModel they need to implement the field [is_incremental](https://github.com/e-mission/e-mission-server/pull/852/files#diff-af7638f59d4ddf79282a90055b614b24817d7e234979395da1b30a1b833b41a7R58). the [code](https://github.com/e-mission/e-mission-server/pull/852/files#diff-fc0a28e68b9e0da800db533678393f809e570e3c5afe05979ccf6cb227bf59aeR102) in `run_model.py` will submit a TimeQuery for the Confirmedtrips if it is an incremental TripModel, otherwise it will set `timequery=None`.

robfitzgerald:
> Did you have to change the model building algorithm, or was it as simple as reading the old model and running the same binning algorithm? Since the new trips will either fit into one of the existing bins or create a new bin anyway...

oh man, no, i wasn't doing anything of the sort, i thought this task ended at making it _possible_ for other future models to be implemented that are incremental, as i stated early on [here](https://github.com/e-mission/e-mission-docs/issues/709#issuecomment-1069412354), when i said "my understanding was that this ticket was only to write the model building process; that the integration/implementation of incremental clustering algorithms themselves was left to future work from someone else this summer."

i didn't see any value in refactoring the existing model to run incrementally, since it doesn't store any condensed representation of what's going on, so this will lead to very large model objects stored every time this pipeline is executed. but if that's something you would like, it won't require anything more than flipping `is_incremental = False` to `True`.

robfitzgerald:
got my e2e test to work. quick questions:
1. are tests run in series or parallel? wondering due to concurrent access to database from tests. i'm running mine using python's built-in test runner because the run*Test*.sh scripts didn't work for me
2. is there a convention for configuration in OpenPATH, like, using some config library, or, use environment variables? there's a few things like the similarity threshold of 500 meters that i'd like to push up to that level of concern

robfitzgerald:
ok, pulling the model configuration from /conf/analysis/... which should hopefully give an example to future model developers for that.

i'm not convinced my incremental code is failsafe but wanted to check in on timeline and task hours bandwidth here before i schedule more work. 

to test incremental, i would implement a mock model implementation which simply counts trips, and then write a test that adds a seed set of training trips and builds/saves a stored model once (ever). then, each time the test is run, a few more trips are added and we check that the model 1) loads correctly from the test start time, 2) adds the correct count to the test counter.

shankari:
@robfitzgerald not `/conf/analysis/` but `conf/analysis` in the `e-mission-server` repo.
Here's how I read it now:
https://github.com/e-mission/e-mission-server/blob/2177582fdea9a52faa8c1a238e62b111dbbcedba/emission/analysis/config.py

robfitzgerald:
oh yes, sorry, my mistake with the leading '/', i [did set this up](https://github.com/e-mission/e-mission-server/pull/852/commits/1320e96096cee80e15b925a3c25868ac8ae5b6cf#diff-182aad7d7f499e940769b6dbafbe6720df1366aee4e251c0242e2bff708b8c20R13) following your pattern there, pointing at conf/analysis/trip_model.conf.json for configuration with a default/sample config provided.

shankari:
> i'm not convinced my incremental code is failsafe but wanted to check in on timeline and task hours bandwidth here before i schedule more work.
> to test incremental, i would implement a mock model implementation which simply counts trips, and then write a test that adds a seed set of training trips and builds/saves a stored model once (ever). then, each time the test is run, a few more trips are added and we check that the model 1) loads correctly from the test start time, 2) adds the correct count to the test counter.

This sounds like a good test and I think you should spend the time to implement it. But I think that is the last step remaining, right?

To make it easier, if you like, you may want to you can use the `real_examples` in `emission/tests/data` as the training set. Since the model works primarily on trip data, you can also randomly generate only the trip level information and not any of the additional detail.

shankari:
@robfitzgerald note also that your new automated tests are failing...

```
ERROR: testBinning (modellingTests.TestGreedySimilarityBinning.TestGreedySimilarityBinning)
when $should_be_grouped trips are the same, they should appear in a bin
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/TestGreedySimilarityBinning.py", line 34, in testBinning
    threshold=0.001,  # ~ 111 meters in degrees WGS84
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/modellingTestAssets.py", line 169, in generate_mock_trips
    purpose_weights=label_data.get('purpose_weights')
Found pc_trip 2016-12-12T09:19:09.784000-08:00
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/modellingTestAssets.py", line 79, in sample_trip_labels
    if len(labels) > 0:
TypeError: object of type 'NoneType' has no len()
ERROR: testNoPrediction (modellingTests.TestGreedySimilarityBinning.TestGreedySimilarityBinning)
when trained on trips in Colorado, shouldn't have a prediction for a trip in Alaska
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/TestGreedySimilarityBinning.py", line 99, in testNoPrediction
    threshold=0.001,  # ~ 111 meters in degrees WGS84
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/modellingTestAssets.py", line 169, in generate_mock_trips
    purpose_weights=label_data.get('purpose_weights')
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/modellingTestAssets.py", line 79, in sample_trip_labels
    if len(labels) > 0:
TypeError: object of type 'NoneType' has no len()
ERROR: testPrediction (modellingTests.TestGreedySimilarityBinning.TestGreedySimilarityBinning)
training and testing with similar trips should lead to a positive bin match
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/TestGreedySimilarityBinning.py", line 65, in testPrediction
    threshold=0.001,  # ~ 111 meters in degrees WGS84
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/modellingTestAssets.py", line 169, in generate_mock_trips
    purpose_weights=label_data.get('purpose_weights')
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/modellingTestAssets.py", line 79, in sample_trip_labels
    if len(labels) > 0:
TypeError: object of type 'NoneType' has no len()
FAIL: testIncrementalRun (modellingTests.TestRunGreedyIncrementalModel.TestRunGreedyModel)
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/TestRunGreedyIncrementalModel.py", line 161, in testIncrementalRun
    'expected number of trips stored in bin')
AssertionError: 3 != 8 : expected number of trips stored in bin
```

robfitzgerald:
> were you also planning to move the model storage to a new collection called something like model_db?

as i said above, we hadn't talked about what it takes to add a new collection to OpenPATH, and above in the thread (somewhere) i had asked if this was in scope, and i don't think we talked about it since then, so i assumed it wasn't. i am on vacation starting tomorrow (Tues July 11th) so perhaps we just create an issue for that.

> definitely prefer the second approach

great, i'll get on this today.

> note also that your new automated tests are failing...

right. i wrote those tests and then started working on e2e tests. somewhere along the line, i think i made some API changes to the trip model, so i have to hop back over to those to update them.

hoping to get this all in the bag today!

robfitzgerald:
@shankari e2e test finished after sorting out some timestamp issues and figuring out more about how the pipeline works. also updated the tests that were failing. if all looks ok to you, then this should be good to go.

robfitzgerald:
sounds good. 

the database queries related to model load/save actions are in [model_storage.py](https://github.com/e-mission/e-mission-server/pull/852/files#diff-797387a6786c978cb38c4028c4eb9aa98a4d0185a523e71529ef0d34f23e7fd9). note that i modified [BuiltinTimeSeries](https://github.com/e-mission/e-mission-server/pull/852/files#diff-0edb91df02d56c15d6cb618e37c9439c6423824908c47f2207836945b974e8b8) to break apart the method `get_first_value_for_field`. the original just returned the field based on the time query, so i broke out the logic that accessed the complete entry into `get_first_entry`. i didn't push the definition up to AbstractTimeSeries because i didn't want to make assumptions about what should be available to all db timeseries interfaces.

i'm around through the end of the day and will check email tomorrow before we leave for a week and a half if other things come up; but also, since this is tied to my personal github account, i'll get email notifications if stuff does come up between now and the 22nd.

robfitzgerald:
@shankari i've returned from vacation. noticing this hasn't been merged. is there anything you are waiting for from me?

shankari:
@robfitzgerald I just haven't had time to review; been busy with intern workload and TRB updates.
I started a review around a month ago and didn't even have time to complete it.
Picking it up now...

emission/tests/modellingTests/TestRunGreedyIncrementalModel.py
```diff
@@ -0,0 +1,186 @@
...
+        updated_model = eamur._load_stored_trip_model(
+            self.user_id,
+            model_type=self.model_type,
+            model_storage=self.model_storage,
+            model_config=self.greedy_model_config
+        )
+
+        self.assertEqual(len(updated_model.bins), 2, 
+            'there should be two bins, one with similar trips, one with an outlier')
+
```
shankari, line 178: Where the outlier is the 6th trip from the setup, right? Can you please add a comment here or in the following assert to clarify?

emission/tests/modellingTests/TestRunGreedyModel.py
```diff
@@ -0,0 +1,164 @@
...
+                logging.debug(f'found {self.total_trips} trips in database')
+
+    def tearDown(self):
+        """
+        delete entries for user self.user_id in the database, not
+        yet implemented in database operations, so these test entries will
+        have to stick around for now.
+        """
+        pass
+
```
shankari, line 83: If you look at the other tests, we delete these by accessing the database directly using `emission.core.get_database`

emission/tests/modellingTests/TestRunGreedyModel.py
```diff
@@ -0,0 +1,164 @@
...
+        logging.debug(f'~~~~ do nothing ~~~~')
+        eamur.update_trip_model(
+            user_id=unused_user_id,
+            model_type=eamumt.ModelType.GREEDY_SIMILARITY_BINNING,
+            model_storage=eamums.ModelStorage.DOCUMENT_DATABASE,
+            min_trips=self.min_trips,
+            model_config=greedy_model_config
+        )
+
+        # todo: check the pipeline for this user to confirm they don't have a current timestamp
```
shankari, line 119: Can you please fix this todo?

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -0,0 +1,128 @@
...
+            "similarity_threshold_meters": 500,      # meters,
+            "apply_cutoff": False,
+            "incremental_evaluation": False
+        }
+        model = eamtg.GreedySimilarityBinning(model_config)
+        
+        model.fit(trips)
+
+        # $should_be_grouped trip features should appear together in one bin
+        at_least_one_large_bin = any(map(lambda b: len(b['features']) >= should_be_grouped, model.bins.values()))
```
shankari, line 48: why is this `>=` instead of `=`? It looks like `testPrediction` assumes that if we generate trips within a threshold they will all end up in the same bin...

shankari (changes requested):
Whoo! that was a pretty heavy lift even for the review, so definitely a heavy lift for the code. Thanks for writing such well-documented and well-tested code.

I do have a lot of comments, but I also want to get this deployed sooner rather than later. It has already been pushed out because of my focus on the TRB papers, and I really don't want to wait too much longer.

So I'll go ahead and address the big comments, but hope you can address the others when you get a chance. 

Big issues I will address:
- any vs. all for matching
- moving the model into a separate model database

setup/environment36.yml
```diff
@@ -8,11 +8,14 @@ dependencies:
...
 - future=0.18.0
 - geojson=2.4.1
+- geocoder=1.38.1
+- geopy=2.2.0
 - google-auth=1.20.1
 - jsonpickle=1.4.1
 - numpy=1.19.1
 - pandas=1.1.0
 - pip=20.2.2
+- polyline=1.4.0
```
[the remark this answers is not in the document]
shankari: We don't use them.
Reverting them passes all tests
https://github.com/e-mission/e-mission-server/pull/872/commits/0b7649f564733117d4218694c21f9747adc694ce

emission/analysis/modelling/similarity/od_similarity.py
```diff
@@ -0,0 +1,21 @@
...
+    in meters.
+    """
+    
+    def extract_features(self, trip: ecwc.Confirmedtrip) -> List[float]:
+        return ctfe.od_features(trip)
+
+    def similarity(self, a: List[float], b: List[float]) -> List[float]:
+        o_dist = ecc.calDistance([a[0], a[1]], [b[0], b[1]])
+        d_dist = ecc.calDistance([a[2], a[3]], [b[2], b[3]])
+        return [o_dist, d_dist]
```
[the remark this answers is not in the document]
robfitzgerald: right, see comment on `confirmed_trip_feature_extraction`. i agree, it creates friction to require flattening here but i think it fits a more general idea of the shape of a "row" of features.

emission/analysis/modelling/similarity/similarity_metric.py
```diff
@@ -0,0 +1,41 @@
+from abc import ABCMeta, abstractmethod
+from typing import List
+import logging
+
+import emission.core.wrapper.confirmedtrip as ecwc
+
```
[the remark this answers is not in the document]
robfitzgerald: i agree. i think a single abstraction that provides an interface for 

```python
def predict(self, data: List[float]) -> Tuple[List[Dict], int]
```

and

```python
def predict(self, data: List[List[float]]) -> List[List[Dict]]
```

would end up being pretty cumbersome to implement. really, if we want to support both types of models (and if it's helpful to have interfaces for them) then we would likely build another base class like `MatrixTripModel` for ones that use gradients or other matrix-based approaches.

emission/tests/modellingTests/TestRunGreedyModel.py
robfitzgerald: ah, `edb.get_pipeline_state_db().delete_many({'user_id': self.testUUID})`. thanks, will do.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
robfitzgerald: you're right. fixed.

emission/tests/modellingTests/TestRunGreedyModel.py
robfitzgerald: updated with the correct pipeline test here (which i just didn't know how to do yet when i wrote this originally, i think)

robfitzgerald:
@shankari applied the requested changes and responded to your questions above.

i tried to review the `testAnyVsAllWhilePredicting` test to see why it fails. one thing i noticed is that it's a particularly skewed input. it only "works" in the binning case because of the order we are reading the input. we train on them exactly in that input order; if we shuffled the input, it would fail, since non-consecutive rows are not "similar". is this the behavior you're looking to test?

regardless, it is weird that the old similarity approach fails with this input, but on a quick review i couldn't see why. same with `testRandomTripsWithinTheSameThreshold`.

shankari:
> i tried to review the testAnyVsAllWhilePredicting test to see why it fails. one thing i noticed is that it's a particularly skewed input. it only "works" in the binning case because of the order we are reading the input. we train on them exactly in that input order; if we shuffled the input, it would fail, since non-consecutive rows are not "similar". is this the behavior you're looking to test?

This is the any vs. all issue that I have been highlighting.
Please see https://github.com/e-mission/e-mission-server/pull/852#discussion_r939611260

robfitzgerald:
> This is the any vs. all issue that I have been highlighting
> Please see https://github.com/e-mission/e-mission-server/pull/852#discussion_r939611260

wow, sorry, i missed this and a whole bunch of other comments strung along with it, starting [here](https://github.com/e-mission/e-mission-server/pull/852#discussion_r939533327). i'm sorry for the lapse. i just wrapped up 4 hours of meetings and some mep work so i can switch back to this right now to address this feedback (and, btw, i totally got this wrong, you are right, i thought that "all" was the bug, not "any" :-O)

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+        selected_record = None
+        
+        for bin_id, bin_record in self.bins.items():
+            for bin_features in bin_record['features']:
+                if self.metric.similar(trip_features, bin_features, self.sim_thresh):
+                    logging.debug(f"found nearest bin id {bin_id}")
+                    logging.debug(f"similar: {trip_features}, {bin_features}")
+                    selected_bin = bin_id
+                    selected_record = bin_record
+                    break
```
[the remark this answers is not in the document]
robfitzgerald: eek, sorry, i had misread the original code's intent here. refactoring it so an incoming trip must match all features in a bin in order to assign it to that bin.

shankari:
> wow, sorry, i missed this and a whole bunch of other comments strung along with it, starting https://github.com/e-mission/e-mission-server/pull/852#discussion_r939533327. i'm sorry for the lapse. i just wrapped up 4 hours of meetings and some mep work so i can switch back to this right now to address this feedback (and, btw, i totally got this wrong, you are right, i thought that "all" was the bug, not "any" :-O)

No worries, I had so many review comments that GitHub actually hid some of the ones in the middle :)
So this is not surprising.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+        unlabeled = list(filter(lambda t: len(t['data']['user_input']) == 0, trips))
+        if len(unlabeled) > 0:
+            msg = f'model.fit cannot be called with unlabeled trips, found {len(unlabeled)}'
+            raise Exception(msg)
+        self._assign_bins(trips)
+        if len(self.bins) > 1 and self.apply_cutoff:
+            self._apply_cutoff()
+        self._generate_predictions()
+
+        binned_features = sum([len(b['features']) for b in self.bins.values() ])
```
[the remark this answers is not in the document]
robfitzgerald: this may have been part of a logging stub. removed.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+        corresponds to a label at the matching index of the label input
+
+        :param trips: 2D array of features to train from
+        """
+        
+        logging.debug(f'fit called with {len(trips)} trips')
+        unlabeled = list(filter(lambda t: len(t['data']['user_input']) == 0, trips))
+        if len(unlabeled) > 0:
+            msg = f'model.fit cannot be called with unlabeled trips, found {len(unlabeled)}'
+            raise Exception(msg)
```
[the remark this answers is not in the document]
robfitzgerald: right, i agree, but i also want to prevent more obtuse error messages from happening in case things change in the calling scope. perhaps this is just a compulsion of someone who relies heavily on the type system :-) happy to remove if you think it's excessive or if you're concerned about the runtime hit.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+        [see https://github.com/e-mission/e-mission-server/blob/5b9e608154de15e32df4f70a07a5b95477e7dbf5/emission/analysis/modelling/tour_model/similarity.py#L118]
+
+        :param data: trips to assign to bins
+        :type data: List[Confirmedtrip]
+        """
+        logging.debug(f"_assign_bins called with {len(trips)} trips")
+        for trip in trips:
+            trip_features = self.extract_features(trip)
+            trip_labels = trip['data']['user_input']
+            bin_id, bin_record = self._nearest_bin(trip)
```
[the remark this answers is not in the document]
robfitzgerald: created a new function called `_find_matching_bin_id` which has this "all" matching behavior

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+        bin_sizes = [len(bin_rec['features']) for bin_rec in self.bins.values()]
+        _, cutoff_bin_size = util.find_knee_point(bin_sizes)
+        logging.debug(
+            "bins = %s, elbow distance = %s" % (num_bins, cutoff_bin_size)
+        )
+
+        updated_bins = {bin_id: bin_rec 
+                        for bin_id, bin_rec in self.bins.items() 
+                        if len(bin_rec['features']) >= cutoff_bin_size}
+
```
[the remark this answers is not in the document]
robfitzgerald: the "sum" was used for the logging only:
```python
logging.debug('the new number of trips is %d' % sum)
```
i'm assuming that doesn't need a code comment.

emission/analysis/modelling/trip_model/run_model.py
```diff
@@ -0,0 +1,172 @@
...
+        return model
+    
+
+def _latest_timestamp(trips: List[ecwc.Confirmedtrip]) -> float:
+    """extract the latest timestamp observed from a list of trips
+
+    :param trips: the trips to review
+    :return: the latest timestamp
+    """
+    ts = sorted(trips, key=lambda r: r['data']['end_ts'], reverse=True)[0]['data']['end_ts']
```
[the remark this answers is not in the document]
robfitzgerald: i don't know, is that a known thing, that copying to pandas from python List and then sorting, and copying back to list, is faster than running python's internal sort? i don't know enough about python to know if that's true. would it be true in the incremental case, where the number of trips is much smaller?

emission/analysis/modelling/trip_model/config.py
```diff
@@ -0,0 +1,79 @@
...
+    return config_data
+
+def get_optional_config_value(key) -> Optional[str]:
+    """
+    get a config value at the provided path/key
+
+    :param key: a key name or a dot-delimited path to some key within the config object
+    :return: the value at the key, or, None if not found
+    """
+    cursor = config_data
```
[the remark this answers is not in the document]
robfitzgerald: i think i was at first.. 🤔  i think i copied that logic from another config.py though? sorry, don't remember.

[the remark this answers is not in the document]
robfitzgerald: there is `pkg_resource.resource_filename` which takes dot-delimited paths too

emission/tests/modellingTests/TestBackwardsCompat.py
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        n = 60
+        trips = etmm.generate_mock_trips(
+            user_id="joe", 
+            trips=n, 
+            origin=(0, 0), 
+            destination=(1, 1), 
+            label_data=label_data, 
+            threshold=0.1,  # Much bigger than the 500m threshold, so we will get multiple bins
```
robfitzgerald, line 160: @shankari 

this mock trip generator will randomly sample origins near the params for origin (0,0) and destination (1,1). it will _guarantee_ that some number of them are _within_threshold_, which by default is `None`. the remaining are guaranteed to be outside of that threshold. 

_BUT_, for those outside of the threshold, there is nothing preventing them from still being near each other. looking at the feature bins, i found this one:

```json
"30": {
        "features": [
            [
                0.007234893836876131,
                0.0016395550304311184,
                0.007234893836876131,
                0.0016395550304311184
            ],
            [
                0.003869586654049201,
                0.0017138992558370781,
                0.003869586654049201,
                0.0017138992558370781
            ]
        ],
        "labels": [ ... ],
        "predictions": [ ... ]
    },
```

i hopped over to jupyter and converted these linestrings to web mercator which is in a near-metric system projection, and these rows indeed appear to be within 500 meters of each other:


```
LINESTRING (805.385 182.514, 805.385 182.514)
LINESTRING (430.760 190.790, 430.760 190.790)
```

so now i'm not sure why the original algorithm didn't get this result.

robfitzgerald: we could revise this mocked data in a similar way to your first backwards compat test building a matrix grid that we know has grid distance > similarity threshold

robfitzgerald:
whoa i really do think github is having trouble tracking the thread state, i just had a few more comments show up, and i also have one that appears twice (from 7 minutes ago:

Screen Shot 2022-08-09 at 4 38 01 PM

beware of wackiness

emission/tests/modellingTests/TestBackwardsCompat.py
shankari: > BUT, for those outside of the threshold, there is nothing preventing them from still being near each other. looking at the feature bins, i found this one:

True. As you can see from the comment, I expect to get multiple bins with multiple trips in each bin.
Not sure what the linestrings represent, but at least some of them are > 500m and some are on the cusp (430m) so maybe the near-metric is the problem?

After you re-run with the fix, how many bins do you get using the new method?

shankari: ok so I re-ran with the fix and I get the same result

```
2022-08-09 16:37:01,436:DEBUG:old bins = 50 but new_bins = 35
```

which seems to indicate that maybe the fix didn't work...

shankari: Re-tested the any versus all test and it does work

```
./e-mission-py.bash emission/tests/modellingTests/TestBackwardsCompat.py TestBackwardsCompat.testAnyVsAllWhilePredicting
Ran 1 test in 0.021s

OK
```

shankari: The monkeytest should still work if the implementation is the same since we are theoretically running the same algorithm on the same dataset. Let's add additional logging to see the differences...

shankari: @robfitzgerald I added logs to the old similarity detection code

```
diff --git a/emission/analysis/modelling/tour_model/similarity.py b/emission/analysis/modelling/tour_model/similarity.py
index 58e1fecb..0ac2ddb0 100644
--- a/emission/analysis/modelling/tour_model/similarity.py
+++ b/emission/analysis/modelling/tour_model/similarity.py
@@ -39,6 +39,7 @@ This is called by cluster_pipeline.py.
 #Determine whether two points are within a cutoff distance threshold
 def within_radius(lat1, lon1, lat2, lon2, radius):
     dist = ecc.calDistance([lon1, lat1], [lon2, lat2])
+    print(f"Comparing distance {dist} to radius {radius}, returning {dist <= radius}")
     return dist <= radius

 def filter_too_short(all_trips, radius):
@@ -124,8 +125,8 @@ class similarity(object):
                         bin.append(a)
                         added = True
                         break
-                except:
-                    print(f"Got exception while matching trip {a}, creating new bin")
+                except Exception as e:
+                    print(f"Got exception {e} while matching trip {a}, creating new bin")
                     added = False
             if not added:
                 self.bins.append([a])
@@ -320,6 +321,7 @@ class similarity(object):
         enda = tripa.end_loc["coordinates"]
         endb = tripb.end_loc["coordinates"]

+        print(f"Comparing trip at index {a} and {b}")
         # Flip indices because points are in geojson (i.e. lon, lat)
         start = within_radius(starta[1], starta[0], startb[1], startb[0], self.radius)
         end = within_radius(enda[1], enda[0], endb[1], endb[0], self.radius)
```

And then ran the test (result in monkeytest_aug_9.log.gz).
And it turns out that there are some trips for which the start **or** the end are within the 500m radius, but in every case, the other side is outside the radius (start_end_true_aug_9.log.gz). Note that in the original implementation, we only return true from `withinDistance` if both start *and* end are within the radius.

[start_end_true_aug_9.log.gz](https://github.com/e-mission/e-mission-server/files/9295500/start_end_true_aug_9.log.gz)
[monkeytest_aug_9.log.gz](https://github.com/e-mission/e-mission-server/files/9295501/monkeytest_aug_9.log.gz)

shankari: The regression in the new code is because the feature extraction is broken.

I added some additional logging to the new similarity metric

```
$ git diff emission/analysis/modelling/similarity/
diff --git a/emission/analysis/modelling/similarity/similarity_metric.py b/emission/analysis/modelling/similarity/similarity_metric.py
index 6be00216..59b8d70d 100644
--- a/emission/analysis/modelling/similarity/similarity_metric.py
+++ b/emission/analysis/modelling/similarity/similarity_metric.py
@@ -37,5 +37,6 @@ class SimilarityMetric(metaclass=ABCMeta):
         :return: true if the feature similarity is within some threshold
         """
         similarity_values = self.similarity(a, b)
+        print(f"while comparing features {a} and {b}, values = {similarity_values}")
         is_similar = all(map(lambda sim: sim <= thresh, similarity_values))
         return is_similar
```

And the start and end features are always identical, leading to identical start and end distances.

```
while comparing features [-0.0026629996435594883, 0.03876364351091465, -0.0026629996435594883, 0.0387636435109
1465] and [-0.02794919620160515, 0.016667362364481252, -0.02794919620160515, 0.016667362364481252], values = [3733.9601065577103, 3733.9601065577103]
while comparing features [0.008169998441481105, 0.007399881502856938, 0.008169998441481105, 0.0073998815028569
38] and [-0.02794919620160515, 0.016667362364481252, -0.02794919620160515, 0.016667362364481252], values = [4146.366766570032, 4146.366766570032]
while comparing features [0.008169998441481105, 0.007399881502856938, 0.008169998441481105, 0.0073998815028569
38] and [-0.0026629996435594883, 0.03876364351091465, -0.0026629996435594883, 0.03876364351091465], values = [3689.6604531600005, 3689.6604531600005]
...
```

robfitzgerald:
I can take a look in the next hour. Is it bad logic there or was the
similarity threshold 4147 or greater?

robfitzgerald:
Another question: is the inference logic also supposed to be all, or was
that also "any"? Sorry not at my laptop to check.

robfitzgerald:
Just had a thought. If a map is truthy, it might be testing that and always
getting True. Perhaps we need to make the result into a list via
all(list(map(...)))

emission/tests/modellingTests/TestBackwardsCompat.py
robfitzgerald: indeed you are right

```python
def destination_features(trip: ecwc.Confirmedtrip) -> List[float]:
    """extract the trip destination coordinates. 

    :param trip: trip to extract features from
    :return: destination coordinates
    """
    try:
        destination = trip['data']['start_loc']["coordinates"]
        return destination
    except KeyError as e:
        msg = 'Confirmedtrip expected to have path data.start_loc.coordinates'
        raise KeyError(msg) from e
```

fixing it now

shankari:
Hi Rob,

The problem is not with the map or all or truthy. The problem is that the start and end locations are always identical, so the start and end distances are always identical.

while comparing features [
-0.0026629996435594883, 0.03876364351091465,
-0.0026629996435594883, 0.03876364351091465] and
[-0.02794919620160515, 0.016667362364481252,
-0.02794919620160515, 0.016667362364481252],
values = [3733.9601065577103, 3733.9601065577103]

Shankari

From: Rob Fitzgerald ***@***.***>
Date: Tuesday, August 9, 2022 at 6:35 PM
To: e-mission/e-mission-server ***@***.***>
Cc: shankari ***@***.***>, Mention ***@***.***>
Subject: Re: [e-mission/e-mission-server] Rjf/incremental inference (PR #852)

CAUTION: This email originated from outside of NREL. Do not click links or open attachments unless you recognize the sender and know the content is safe.
Just had a thought. If a map is truthy, it might be testing that and always
getting True. Perhaps we need to make the result into a list via
all(list(map(...)))

robfitzgerald:
@shankari 

fixed the typo breaking the feature extraction for od features (see [comment](https://github.com/e-mission/e-mission-server/pull/852#discussion_r941940119)).

robfitzgerald:
still fails though because it's moved onto a new problem. looking into it.

shankari:
> still fails though because it's moved onto a new problem. looking into it.

That test passes for me now.

wrt other failure, I wonder if it is related to 

```
--- a/emission/analysis/modelling/trip_model/greedy_similarity_binning.py
+++ b/emission/analysis/modelling/trip_model/greedy_similarity_binning.py
@@ -121,7 +121,7 @@ class GreedySimilarityBinning(eamuu.TripModel):
         predicted_bin, bin_record = self._nearest_bin(trip)
         if predicted_bin is None:
             logging.debug(f"unable to predict bin for trip {trip}")
-            return [], -1
+            return [], 0
         else:
```

I had to make that change locally when I was testing earlier 
https://github.com/e-mission/e-mission-server/pull/872#issuecomment-1207480908

robfitzgerald:
> wonder if it is related to

that was it. i wrongly set this as a flag value for when no prediction occurred. i've updated it to return 0 now.

shankari:
@robfitzgerald  you now need to change `testNoPrediction` to expect 0 and not -1
I think that is it wrt unit tests

robfitzgerald:
found another that failed too, it was expecting 2 bins with one singleton outlier bin, but it really should have been two. added a comment to explain, in TestRunGreedyIncrementalModel.py:

```
        #        0      1      2      3      4      5  labels?
        # 0   True   True   True   True  False  False     True
        # 1   True   True   True   True  False  False    False
        # 2   True   True   True   True  False  False    False
        # 3   True   True   True   True  False  False     True
        # 4  False  False  False  False   True  False     True
        # 5  False  False  False  False  False   True     True
        
        # trip 0 and 3 are similar and will form bin 0
        # trip 1 and 2 have no labels and will be ignored
        # trips 4 and 5 are both dis-similar from the rest and will form singleton bins
```

also, realized i needed to "delete_many" on the pipeline state as well for test cleanup.

confirmed all tests in the modellingTests directory are OK this time:

```
Ran 20 tests in 2.341s

OK
```

shankari (commented):
Couple of code changes

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,267 @@
...
+        for bin_id, bin_record in self.bins.items():
+                matches_bin = all([self.metric.similar(trip_features, bin_sample, self.sim_thresh)
+                    for bin_sample in bin_record['features']])
+                if matches_bin:
+                    return bin_id
+        return None
+
+    def _nearest_bin(self, trip: ecwc.Confirmedtrip) -> Tuple[Optional[int], Optional[Dict]]:
+        """
+        finds a bin which contains at least one matching feature. the
```
shankari, line 188:
```suggestion
        finds a bin which contains at least all matching features. the 
```

shankari (commented):
One more code change

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,267 @@
...
+        return None, None
+
+    def _apply_cutoff(self):
+        """
+        removes small clusters by an "elbow search" heuristic. see
+        https://stackoverflow.com/a/2022348/4803266.
+        """
+        # the cutoff point is an index along the sorted bins. any bin with a gte
+        # index value is removed, as that bin has been found to be smaller than the cutoff.
+        bins_sorted =  self.bins.sort(key=lambda bin: len(bin['features']), reverse=True)
```
shankari, line 217:
```suggestion
        # This was the last line of calc_cutoff_bins in the old code, and is moved to the equivalent of delete_bins in the new code
        bins_sorted =  self.bins.sort(key=lambda bin: len(bin['features']), reverse=True)
```

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,267 @@
...
+                    logging.debug(f"found nearest bin id {bin_id}")
+                    logging.debug(f"similar: {trip_features}, {bin_features}")
+                    return bin_id, bin_record
+
+        return None, None
+
+    def _apply_cutoff(self):
+        """
+        removes small clusters by an "elbow search" heuristic. see
+        https://stackoverflow.com/a/2022348/4803266.
```
shankari, line 213:
```suggestion
        removes small clusters by an "elbow search" heuristic. see
        https://stackoverflow.com/a/2022348/4803266.
        Copied over from https://github.com/e-mission/e-mission-server/blob/5b9e608154de15e32df4f70a07a5b95477e7dbf5/emission/analysis/modelling/tour_model/similarity.py#L158
```

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+            else:
+                # create new bin
+                new_bin_id = str(len(self.bins))
+                new_bin_record = {
+                    "features": [trip_features],
+                    "labels": [trip_labels],
+                    "predictions": []
+                }
+                logging.debug(f"creating new bin {new_bin_id} at location {trip_features}")
+                self.bins[new_bin_id] = new_bin_record
```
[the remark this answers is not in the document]
shankari: you actually put that in as the first line in `apply_cutoff`

shankari (commented):
More minor code changes

emission/analysis/modelling/trip_model/config.py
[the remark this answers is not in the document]
shankari: Not sure what you mean by this

```
$  grep -r resource_filename emission/analysis | wc -l
       0
```

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,269 @@
...
+        Copied over from https://github.com/e-mission/e-mission-server/blob/5b9e608154de15e32df4f70a07a5b95477e7dbf5/emission/analysis/modelling/tour_model/similarity.py#L158
+        """
+        # the cutoff point is an index along the sorted bins. any bin with a gte
+        # index value is removed, as that bin has been found to be smaller than the cutoff.
+        # This was the last line of calc_cutoff_bins in the old code, and is moved to the equivalent of delete_bins in the new code
+        bins_sorted =  self.bins.sort(key=lambda bin: len(bin['features']), reverse=True)
+        
+        num_bins = len(bins_sorted)
+        bin_sizes = [len(bin_rec['features']) for bin_rec in bins_sorted.values()]
+        _, cutoff_bin_size = util.find_knee_point(bin_sizes)
```
shankari, line 223:
```suggestion


# The first two lines below correspond to the original lines below in the original elbow_distance
#         y = [0] * len(self.bins)
#         for i in range(len(self.bins)):
#             y[i] = len(self.bins[i])
        num_bins = len(bins_sorted)
        bin_sizes = [len(bin_rec['features']) for bin_rec in bins_sorted.values()]
        _, cutoff_bin_size = util.find_knee_point(bin_sizes)
```

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+        return selected_bin, selected_record
+
+    def _apply_cutoff(self):
+        """
+        removes small clusters by an "elbow search" heuristic. see
+        https://stackoverflow.com/a/2022348/4803266.
+        """
+        num_bins = len(self.bins)
+        bin_sizes = [len(bin_rec['features']) for bin_rec in self.bins.values()]
+        _, cutoff_bin_size = util.find_knee_point(bin_sizes)
```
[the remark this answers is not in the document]
shankari: Re-added in 
https://github.com/e-mission/e-mission-server/pull/852/files#r943092075

https://github.com/e-mission/e-mission-server/pull/852/commits/c3837401e5419fef9507c2f657bb8e0482dbc197

emission/analysis/modelling/trip_model/run_model.py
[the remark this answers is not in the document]
shankari: It is definitely faster to sort in pandas/numpy but not sure about the overhead of copying back and forth.
Would probably have to actually try it out.
I'm happy with holding off on this given the other major changes to the pipeline.

shankari:
@robfitzgerald I assume that you are not planning on making the changes to move the model to a separate model DB and I should not wait for them.

Can you address the two or three remaining comments, mainly around naming and comments?

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,281 @@
...
+        bins_sorted =  self.bins.sort(key=lambda bin: len(bin['features']), reverse=True)
+        
+
+
+# The first two lines below correspond to the original lines below in the original elbow_distance
+#         y = [0] * len(self.bins)
+#         for i in range(len(self.bins)):
+#             y[i] = len(self.bins[i])
+
+
```
shankari, line 228: Remove duplicated comment

```suggestion
```

emission/core/wrapper/pipelinestate.py
```diff
@@ -19,6 +19,7 @@ class PipelineStages(enum.Enum):
     CLEAN_RESAMPLING = 11
     MODE_INFERENCE = 4
     LABEL_INFERENCE = 14
+    TRIP_MODEL = 16
```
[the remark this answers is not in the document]
shankari: Will handle in separate update to avoid blowing up this PR even further.

emission/storage/pipeline_queries.py
```diff
@@ -121,6 +121,22 @@ def mark_mode_inference_complete(user_id):
...
+        # key off of Confirmedtrip end timestamp for the provided time range
+        tq.timeType = 'data.end_ts'
+        return tq
+
+def mark_trip_model_done(user_id, last_ts):
+    mark_stage_done(user_id, ps.PipelineStages.TRIP_MODEL, last_ts + END_FUZZ_AVOID_LTE)
+
+def mark_trip_model_failed(user_id):
+    mark_stage_failed(user_id, ps.PipelineStages.TRIP_MODEL)
+
```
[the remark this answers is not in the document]
shankari: Will handle in separate update to avoid blowing up this PR even further.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, user_id):
                 "metrics/daily_mean_median_speed": self.analysis_timeseries_db,
                 "inference/prediction": self.analysis_timeseries_db,
                 "inference/labels": self.analysis_timeseries_db,
+                "inference/trip_model": self.analysis_timeseries_db,
```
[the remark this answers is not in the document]
shankari: Will handle in separate update to avoid blowing up this PR even further.

emission/storage/decorations/analysis_timeseries_queries.py
```diff
@@ -37,6 +37,7 @@
 METRICS_DAILY_USER_MEDIAN_SPEED = "metrics/daily_user_median_speed"
 METRICS_DAILY_MEAN_MEDIAN_SPEED = "metrics/daily_mean_median_speed"
 INFERRED_LABELS_KEY = "inference/labels"
+TRIP_MODEL_STORE_KEY = "inference/trip_model"
```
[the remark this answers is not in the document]
shankari: Will handle in separate update to avoid blowing up this PR even further.

emission/analysis/modelling/trip_model/config.py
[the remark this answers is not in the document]
robfitzgerald: ooh no, pkg_resource is a nifty built-in python library that lets one reference files that are in a python package via dot notation. it gets around os-specific pathing. 

for example, if you have a file

```
e-mission/
  __init__.py
  resources/
    __init__.py
    config/
      __init__.py
      my_conf.json
```

then you can reference this in code via

```python
from pkg_resource import resource_filename

file = resource_filename("emission.resources.config", "my_conf.json")
```

note, we need all directories in the path to be "python modules" aka have an __init__.py file in them for this to work.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,252 @@
...
+                    [f1, f2, .., fn],
+                    ...
+                ],
+                "labels": [
+                    { label1: value1, ... }
+                ],
+                "predictions": [
+                    { "labels": { label1: value1, ... }, 'p': p_val }
+                ]
+            }
```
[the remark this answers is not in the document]
robfitzgerald: > why do we have "features" instead of "trips"

1. since this object is also the one we will store in the database, it is smaller if it only includes the features we require for classification
2. trips are domain-specific, i thought for most folks just working on classification they would prefer working in the domain of real number vectors
3. this way, we extract trip features at most once

> why do we save the labels and predictions separately? I double-checked the current file-based storage, and

labels are the input and are extracted when we extract the features. predictions are tacked on once we finish fitting the model. if optimizing memory/storage size is important, this could be re-written so that we remove "labels" once we finish creating "predictions". but
1. when re-running an incremental model, we can apply the same code that trains from "labels" if we still have them (as opposed to reverse-engineering the labels from the previous predictions)
2. to a smaller degree, leaving the labels does help with debugging and explaining the model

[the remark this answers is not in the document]
robfitzgerald: > it should definitely be explained better in the comments

sure. but, i mean, the same could be said about the previous implementation.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -0,0 +1,275 @@
...
+        for bin_id, bin_record in self.bins.items():
+                matches_bin = all([self.metric.similar(trip_features, bin_sample, self.sim_thresh)
+                    for bin_sample in bin_record['features']])
+                if matches_bin:
+                    return bin_id
+        return None
+
+    def _nearest_bin(self, trip: ecwc.Confirmedtrip) -> Tuple[Optional[int], Optional[Dict]]:
+        """
+        finds a bin which contains at least all matching features. the
```
robfitzgerald, line 188: ah, sorry, i did mistakenly read the behavior of find_bin was also "any". we just need one of these bin search methods, let me know if you would like me to clean this up.

shankari:
> since this object is also the one we will store in the database, it is smaller if it only includes the features we require for classification

I understand, but calling it only `features` makes it confusing when we are iterating. It looks like we are iterating over individual features of a trip but we are actually iterating over rows of features/trips.

> sure. but, i mean, the same could be said about the previous implementation.

Sure, but the previous implementation was by an undergrad intern who went on to work on the "foundations of cryptography". Now that we have this shiny new refactored implementation, I want to make sure that it is readable and maintainable for the long term.

shankari (approved):
Changes look good conceptually.
I will go ahead and merge once tests pass.

emission/tests/modellingTests/TestRunGreedyIncrementalModel.py
```diff
@@ -0,0 +1,198 @@
...
+            model_storage=self.model_storage,
+            model_config=self.greedy_model_config
+        )
+
+        # the 5th and 6th trip in the original dataset were outliers and should form their own cluster
+        self.assertEqual(len(updated_model.bins), 3, 
+            'there should be three bins, one with 2 similar trips, and two singleton bins')
+
+        trips_in_bin = len(updated_model.bins['0']['feature_rows'])
+        print(f'trips in bins: {[len(x["feature_vectors"]) for x in updated_model.bins.values()]}')
```
shankari, line 190:
```suggestion
        print(f'trips in bins: {[len(x["feature_rows"]) for x in updated_model.bins.values()]}')
```

PR #854, opened 2022-06-05
Remove the webapp from the server

shankari:
As part of modularizing the platform, let's delete the webapp from the server code
We should create a separate webapp microservice, presumably using node.js so
that we can iterate on it separately.

This should also significantly lower the space taken by the e-mission-server
repo, simplify its dependencies, and stop generating dependabot alerts.


shankari:
We now have a 404 when we access the index.html

Screen Shot 2022-06-04 at 6 09 35 PM

shankari:
Using the error redirect functionality to redirect all unknown pages to the openpath page
- error handling: https://stackoverflow.com/questions/20085978/how-to-throw-custom-404-messages-in-python-bottle

However, the `redirect` method didn't work

Used a manual redirect instead and it works
-  https://techoverflow.net/2019/07/03/python-bottle-minimal-redirect-example/

Redirecting to the openpath page "https://www.nrel.gov/transportation/openpath.html" on 404....

shankari:
Given that we don't have an entry for index.html, and we redirect all missing URLs to the OpenPATH site, maybe we are done wrt "change the default index.html for the container". We can even remove the `index.html` from the docker repository, but not sure we need to do that right now.

PR #857, opened 2022-06-07
change relative imports to absolute path

hlu109:
fix imports so that these python source files can be called from e-mission-eval-private-data or other repositories.

### DIFF
--- a/emission/analysis/modelling/tour_model/get_plot.py
+++ b/emission/analysis/modelling/tour_model/get_plot.py
@@ -4,7 +4,7 @@ import pandas as pd
 from matplotlib import cm
 import folium
 import branca.colormap as clm
-import load_predict as predict
+import emission.analysis.modelling.tour_model.load_predict as predict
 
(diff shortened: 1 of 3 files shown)


shankari:
@hlu109 this is a good change, but unnecessary because we don't currently use any of the files in `emission/analysis/modelling/tour_model`. As part of a flurry of last minute refactoring, I pulled out the first round only into `tour_model_first_only`, and that is being further refactored https://github.com/e-mission/e-mission-server/pull/852

**don't use the existing notebooks for the tour model; they will only confuse you and cause you to reinvent the wheel**

Use https://github.com/e-mission/e-mission-eval-private-data/pull/28 instead

shankari:
I might have some pending changes for that PR on my laptop; let me clean them up and submit them as well

shankari:
Closed since I merged https://github.com/e-mission/e-mission-eval-private-data/pull/28
But thanks for the PR!

PR #858, opened 2022-06-08
Pull from master

shankari:
Let's see the conflicts


shankari:
entry: keep both
webapp: keep both (mostly) except for the line with `@post("/datastreams/find_entries/<time_type>", method=['OPTIONS', 'POST']) keep the zephyr version
timeseries: keep both
`

shankari:
can you also file an issue to merge the emevalzephyr data types to master so we don't have to keep doing this

PR #865, opened 2022-06-22
Disable aggregate access to the spatio-temporal data

shankari:
- Disable aggregate access to the spatio-temporal data for now until we can
  ensure that we don't share trajectories when there are too few people
  participating in the program.
- This was not an issue until now, because we:
  - required gmail to login
  - required a program token to login

But now that we open up enrollment via anonymous token, we can run into the
issue where malicious users sign up for studies, don't contribute anything, and
simply pull aggregate spatio-temporal data at will using an automated script.
If we don't have a lot of users in the study, this can run into issues with
tracking trajectories through repeated queries.

We can consider multiple potential fixes for the future:
- apply k-modularity for the trajectories (which will require map matching)
- require a client secret for access to the spatio-temporal data so that only
  the phone app can retrieve the data
- change the granularity of the data to ensure privacy

But let's wait and pick one later.

For now, we disable the aggregate spatio-temporal queries and remove the heatmap
from the UI
https://github.nrel.gov/kshankar/openpath-phone/issues/2#issuecomment-44111

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -105,7 +105,13 @@ def _fill_aggregate_backward_compat(request):
 @post("/result/heatmap/pop.route/<time_type>")
 def getPopRoute(time_type):
   _fill_aggregate_backward_compat(request)
-  user_uuid = get_user_or_aggregate_auth(request)
+  # Disable aggregate access for the spatio-temporal data temporarily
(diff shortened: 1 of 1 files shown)


shankari:
Another potential fix: condition access to aggregate data on data contribution - e.g. contribute one day, get 5 queries.
But it seems like that would also be fairly easy to game by pushing fake data, or installing a mock location app on a phone

PR #870, opened 2022-07-20
Add Error_Rates get method

allenmichael099:
Since we are making a new collection called Error_Rates, I added a get_database function to get the Error_Rates collection.
get_Error_Rates_db()


shankari:
This is pulling from the `ceo_ebike_project`, so is pulling in a bunch of unwanted changes.
So I will not merge this.

@allenmichael099 also, please use the modifiable model storage wrapper
https://github.com/e-mission/e-mission-server/tree/07e01eab5accb9b4f8204ecab86c0c79caf3dfa9/emission/storage/modifiable
instead of creating/using a database in mongodb directly

PR #872, opened 2022-08-07
Restructure clustering codebase

shankari:
This is an add-on to https://github.com/e-mission/e-mission-server/pull/852
that addresses several of the requested changes.

Please see https://github.com/e-mission/e-mission-server/pull/852 for the discussion around the changes.


shankari:
As expected, the backwards compat test failed with 

```
FAIL: testAnyVsAllWhilePredicting (modellingTests.TestBackwardsCompat.TestBackwardsCompat)
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/TestBackwardsCompat.py", line 67, in testAnyVsAllWhilePredicting
    f"old bins = {old_bins} but new_bins = {new_bins}")
AssertionError: 2 != 1 : old bins = [[0, 1], [2, 3]] but new_bins = {'0': {'features': [[0.8, 0.8, 0.8, 0.8], [0.9, 0.9, 0.9, 0.9], [1.0, 1.0, 1.0, 1.0], [1.1, 1.1, 1.1, 1.1]], 'labels': [{'mode_confirm': 'walk', 'purpose_confirm': 'exercise'}, {'mode_confirm': 'walk', 'purpose_confirm': 'exercise'}, {'mode_confirm': 'walk', 'purpose_confirm': 'exercise'}, {'mode_confirm': 'walk', 'purpose_confirm': 'exercise'}], 'predictions': [{'labels': {'mode_confirm': 'walk', 'purpose_confirm': 'exercise'}, 'p': 1.0}]}}

Ran 256 tests in 653.003s
```

shankari:
Now, let's see if we can create more of a monkeytest (that should also fail) so that we can make sure to catch other failures that I didn't notice in the code review.

shankari:
Wrote a random test generator with a small threshold that put everything into the same bin; it passed

```
./e-mission-py.bash emission/tests/modellingTests/TestBackwardsCompat.py TestBackwardsCompat.testRandomTripsWithinTheSameThreshold
Ran 1 test in 0.203s

OK
```

Then, used random trips with a large threshold which would fit into multiple bins.

- Failure # 1: old version returns 0 for the size, new version returns 

```
    if pred_bin == -1:
        logging.info(f"No match found for {trip['data']['start_fmt_time']} early return")
        return [], 0
```

vs.

```
        if predicted_bin is None:
            logging.debug(f"unable to predict bin for trip {trip}")
            return [], -1
        else:
```

Changed code to match...

- Issue 2: setting the threshold to 10000 causes trips that are way too far apart and results in 50 different clusters.
Since `threshold=0.001` corresponds to 111 meters, let's set this to 1 instead.

Ah! that fails, but still with old_bins = 50 and new_bins = 49. Let's tighten a bit more to 0.1

Ah now we have a real difference

```
2022-08-07 13:34:10,141:INFO:greedy binning model fit to 50 rows of trip data
2022-08-07 13:34:10,283:DEBUG:old bins = 50 but new_bins = 35
```

Let's stick with this for now and commit it. Again, the new test should fail.

robfitzgerald:
@shankari what is the purpose of this separate PR? should i be merging it into my PR? should i be applying the fixes you requested to my original PR, or here?

shankari:
@robfitzgerald I created this new PR since I wasn't sure about your availability and I wanted to fix at least the big ticket items in the original PR:
- fixing all vs any
- model DB

If you do have availability, I can change this to point to your repo and you can merge + fix.
If not, I can make the changes here, and when I merge, the original PR will automatically show as "merged"

PR #874, opened 2022-08-12
Store the model in a new model_db instead of in the analysis_db

shankari:
Unlike the user inputs, which are read-only from an analysis perspective, or
analysis results, which are write once (except for confirmed trips), the models
are updated frequently, every time they are re-calibrated.

So we create a new interface to match this new access pattern.
Unlike the timeseries, which focuses on reading data for a particular range,
the new interface focuses on updating models and retrieving the most recent model

We still need to write code to purge obsolete models so that the database does
not grow unboundedly.

Changes:
- new database
- new abstract interface in `emission/storage/modifiable`
- new concrete implmentation in `emission/storage/modifiable`
- change the `model_storage` to use the new interface instead of the timeseries
    - bonus, don't need to use the concrete implementation any more since this
      implementation is all about the most recent version
- change the test to delete the model db in addition to the analysis DB
    - we need to now delete both since the trips are in the analysis DB while
      the models are in the separate model DB

### DIFF
--- a/emission/analysis/modelling/trip_model/model_storage.py
+++ b/emission/analysis/modelling/trip_model/model_storage.py
@@ -6,9 +6,7 @@ import json
 import emission.analysis.modelling.trip_model.model_type as eamum
 import emission.core.wrapper.tripmodel as ecwu
 import emission.storage.decorations.analysis_timeseries_queries as esda
-import emission.storage.timeseries.abstract_timeseries as esta
-import emission.storage.timeseries.builtin_timeseries as estb
-import pymongo
+import emission.storage.modifiable.abstract_model_storage as esma
 
 
 class ModelStorage(Enum):
@@ -51,14 +49,8 @@ def load_model(user_id, model_type: eamum.ModelType, model_storage: ModelStorage
         
         # retrieve stored model with timestamp that matches/exceeds the most
         # recent PipelineState.TRIP_MODEL entry        
-        ts = esda.get_timeseries_for_user(user_id)
-        if not isinstance(ts, estb.BuiltinTimeSeries):
-            raise Exception('user model storage requires BuiltInTimeSeries')
-        latest_model_entry = ts.get_first_entry(
-            key=esda.TRIP_MODEL_STORE_KEY,
-            field='metadata.write_ts',
-            sort_order=pymongo.DESCENDING
-        )
+        ms = esma.ModelStorage.get_model_storage(user_id)
+        latest_model_entry = ms.get_current_model(key=esda.TRIP_MODEL_STORE_KEY)
 
         if latest_model_entry is None:
             logging.debug(f'no {model_type.name} model found for user {user_id}')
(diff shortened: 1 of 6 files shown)


shankari:
@sebastianbarry since you wanted to review my changes.
This is the first step in https://github.com/e-mission/e-mission-docs/issues/709#issuecomment-1212703570

shankari:
Failure:

```
FAIL: testIncrementalRun (modellingTests.TestRunGreedyIncrementalModel.TestRunGreedyModel)
incremental trip models train from Confirmedtrip entries at most
  File "/home/runner/work/e-mission-server/e-mission-server/emission/tests/modellingTests/TestRunGreedyIncrementalModel.py", line 193, in testIncrementalRun
    'expected number of trips stored in bin')
AssertionError: 2 != 5 : expected number of trips stored in bin

```


When I run it locally, I first got

```
FAIL: testIncrementalRun (emission.tests.modellingTests.TestRunGreedyIncrementalModel.TestRunGreedyModel)
incremental trip models train from Confirmedtrip entries at most
  File "/Users/kshankar/e-mission/gis_branch_tests/emission/tests/modellingTests/TestRunGreedyIncrementalModel.py", line 188, in testIncrementalRun
    'there should be three bins, one with 2 similar trips, and two singleton bins')
AssertionError: 1 != 3 : there should be three bins, one with 2 similar trips, and two singleton bins

```

And then the same error as above.

shankari:
ok, so comparing a working vs non-working use case, when the model is first created, we have the same number of bins

Screen Shot 2022-08-11 at 9 56 28 PM

but after the update, we have fewer

Screen Shot 2022-08-11 at 9 57 33 PM

shankari:
ok, this is because when we have two records, we are retrieving the older record and not the newer one.

```

:DEBUG:upsert_doc called with key inference/trip_model
:DEBUG:retrieved latest trip model recorded at timestamp 1660279601.0412931
:DEBUG:{'user_id': UUID('aa9fdec9-2944-446c-8ee2-50d79b3044d3'), 'metadata':
{'key': 'inference/trip_model', 'platform': 'server', 'write_ts':
1660279601.0412931, 'time_zone': 'America/Los_Angeles', 'write_local_dt':
{'year': 2022, 'month': 8, 'day': 11, 'hour': 21, 'minute': 46, 'second': 41,
'weekday': 3, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time':
'2022-08-11T21:46:41.041293-07:00'}, 'data': {'model_ts': 1466467959.767,
'model_type': 'GREEDY', 'model': {'0': {'feature_rows': [2 entries], '1': {'feature_rows': [1 entry]}}}}}


:DEBUG:upsert_doc called with key inference/trip_model
:DEBUG:Inserting entry Entry({'_id': ObjectId('62f5db31ee41eacb500a20f9'),
'user_id': UUID('aa9fdec9-2944-446c-8ee2-50d79b3044d3'), 'metadata':
Metadata({'key': 'inference/trip_model', 'platform': 'server', 'write_ts':
1660279601.104193, 'time_zone': 'America/Los_Angeles', 'write_local_dt':
LocalDate({'year': 2022, 'month': 8, 'day': 11, 'hour': 21, 'minute': 46,
'second': 41, 'weekday': 3, 'timezone': 'America/Los_Angeles'}),
'write_fmt_time': '2022-08-11T21:46:41.104193-07:00'}), 'data':
Tripmodel({'model_ts': 1660279591.0647948, 'model_type': 'GREEDY', 'model':
{'0': {'feature_rows': [5 entries], '1': {'feature_rows': [1 entry],
'2': {'feature_rows': [1 entry]}) into model DB


:DEBUG:retrieved latest trip model recorded at timestamp 1660279601.0412931
```

PR #876, opened 2022-08-31
Update dependencies

aGuttman:
Use a newer version of conda and update dependencies so they solve.

Will require new image to be built/pushed and small changes in dashboard repo to use the updated image.

See issue https://github.com/e-mission/em-public-dashboard/issues/37

### DIFF
--- a/setup/environment36.notebook.additions.yml
+++ b/setup/environment36.notebook.additions.yml
@@ -3,9 +3,9 @@ channels:
 - conda-forge
 - defaults
 dependencies:
-- branca=0.4.1
-- folium=0.11.0
-- ipython=7.17.0
-- jedi=0.17.2
+- branca=0.5.0
+- folium=0.12.1.post1
+- ipython=7.33.0
+- jedi=0.18.1
 - jupyter=1.0.0
-- matplotlib=3.2.2
+- matplotlib=3.5.1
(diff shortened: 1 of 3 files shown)


shankari:
> Will require new image to be built/pushed and small changes in dashboard repo to use the updated image.

Actually, it does not.
Using `emission/e-mission-server.dev.server-only` and rebuilding the containers will fix it.

Can you please make the change and confirm that it works before I merge this?

shankari:
Testing done:
After the change, urllib has the fix

```
urllib3                   1.26.11            pyhd8ed1ab_0    conda-forge
```

aGuttman:
Replace `conda clean -p` with `find /root/miniconda-*/pkgs -wholename \*info/test\* -type d | xargs rm -rf` as suggested in https://github.com/e-mission/em-public-dashboard/issues/37

Image builds, container runs, test files are removed.

Question on how to best merge this. This branch was forked off of master before the PR that added the Dockerfile and associated files was merged. Would it be appropriate to make these changes to the supplemental docker script from a new branch, create a PR, merge those changes in and then merge this PR in with those changes already in place? The alternative seems to be copying and pasting the docker stuff into this branch, making the change and submitting it all as one PR, but that feels off to me. Neither seems right, actually. If there a good way to do this?

shankari:
Good point wrt committing.
I am fine with committing this as such since these changes fix the hanging issue.
The `find | xargs` removes false positives for the automated vulnerability checker and can be its own commit.
I just wanted to make sure that we were ready to go with that change.

PR #877, opened 2022-09-21
Gis based mode detection

asiripanich:


asiripanich:
Whoops. Sorry @shankari. :)

PR #880, opened 2022-10-06
Create a script to automatically reset the pipeline to the correct va…

shankari:
…lue for each user

This replaces the current, cumbersome method in which I get a list of users with an invalid pipeline state, and then reset them manually.

This automates that manual algorithm, while also making it more sophisticated.

The algorithm essentially finds all invalid states for a current user, groups them, and then resets to the oldest timestamp. This will ensure that we don't skip some data by moving the pointer ahead to a time that is not yet processed.

This fixes https://github.com/e-mission/e-mission-docs/issues/805

### DIFF
--- a/bin/monitor/reset_invalid_pipeline_states.py
+++ b/bin/monitor/reset_invalid_pipeline_states.py
@@ -1,0 +1,25 @@
+import arrow
+import logging
+import argparse
+import pandas as pd
+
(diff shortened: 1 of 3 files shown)


shankari:
Testing done:
The newly updated pipeline tests pass

```
Ran 13 tests in 185.486s

OK
```

PR #881, opened 2022-10-06
Wrap a mongodb cursor in a list before converting it to a dataframe

shankari:
Using `pd.json_normalize` directly on a cursor seems to skip the first line. Wrapping the cursor in a list seems to fix it.
Not sure if this is expected behavior or whether it has been fixed now in pandas

This change:
- adds a new test that explicitly tests this functionality - the test passes if the first line is missing in the cursor but not in the list. This does not call any functions in the reset.
- adds a new check to the existing mock test by adding a new user (`no_missing`) to the front of the test data and ensuring that the user is also reset correctly. Without this change, the first entry had multiple repetitions, so even if the first entry was skipped, the user would still be reset
- fixes the issue in the reset code

Testing done:
- tests pass

This re-fixes https://github.com/e-mission/e-mission-docs/issues/805

### DIFF
--- a/emission/tests/pipelineTests/TestPipelineReset.py
+++ b/emission/tests/pipelineTests/TestPipelineReset.py
@@ -676,10 +676,7 @@ class TestPipelineReset(unittest.TestCase):
              arrow.get("2022-09-21").shift(hours=-3).timestamp, # user3, only one entry
              arrow.get("2022-09-22").shift(hours=-3).timestamp]) # user4, only one entry
 
-    def testAutoResetMock(self):
-        # The expectation is that user_5 and user_6 will be stripped out since:
-        # user5 does not have a curr_run_ts
-        # user6 is an output_gen state
+    def testNormalizeWithACursor(self):
         invalid_states_mixed = pd.DataFrame({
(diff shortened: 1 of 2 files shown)


emission/tests/pipelineTests/TestPipelineReset.py
```diff
@@ -676,23 +676,50 @@ def testGetAllResets(self):
...
+
+        df_from_cursor = pd.json_normalize(edb.get_pipeline_state_db().find())
+        df_from_list = pd.json_normalize(list(edb.get_pipeline_state_db().find()))
+
+        # This is actually incorrect because we saved everything, so we should read back everything
+        # but it is the current behavior, so let's flag if it changes
+        self.assertEqual(len(df_from_cursor), len(invalid_states_mixed) - 1)
+
+        # This is actually incorrect because we saved everything, so we should read back everything
+        # but it is the current behavior, so let's flag if it changes
```
shankari, line 701:
```suggestion
        # This is the expected behavior in all cases, but let's make sure that it stays as we move through versions of pandas
```

PR #885, opened 2022-10-27
add token collection

aGuttman:
DRAFT:
Add token collection to database, draft basic access methods, draft usage in auth


emission/storage/decorations/token_queries.py
```diff
@@ -0,0 +1,21 @@
...
+standard_library.install_aliases()
+from builtins import *
+import logging
+
+import emission.core.get_database as edb
+import emission.storage.timeseries.aggregate_timeseries as estag
+from uuid import UUID
+
+def get_all_tokens():
+    tokens_list = [e for e in edb.get_token_db().find()]
```
shankari, line 16: this can just be `list(edb.get_token_db().find())`

emission/net/auth/token_list.py
```diff
@@ -13,6 +13,10 @@
 
 class TokenListMethod(object):
     def __init__(self):
+
+        # import emission.storage.decorations.token_queiries as tq
+        # key_file = tq.get_all_tokens()
+
```
shankari, line 19:
```suggestion

        # import emission.storage.decorations.token_queiries as tq
        # self.token_list = tq.get_all_tokens()

```

emission/core/get_database.py
```diff
@@ -40,6 +40,17 @@
 def _get_current_db():
     return _current_db
 
+def get_token_db():
+    Tokens= _get_current_db().Stage_Tokens
+    print("tokens")
+    mydict = { "name": "John", "address": "Highway 37" }
+
+    x = Tokens.insert_one(mydict)
+    print(_get_current_db().list_collection_names())
```
shankari, line 49:
```suggestion
```

emission/storage/decorations/token_queries.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+# Utility class to return useful user queries
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *
```
shankari, line 8:
```suggestion
```

These are from python2 -> python3 upgrade; we can just remove this since we are writing directly in python3 now.

shankari (changes requested):
this looks good so far, and is almost done.
only other big change is to add a unit test to emission/tests`

aGuttman:
Unit test currently has three tests.

1. Add a single entry via the wrapper, make sure that it is there.
2. Add several entries via the wrapper using a json file pointing to a list of tokens. Make sure that they are there.
3. Init the TokenListMethod in token_list.py, which loads in entries using the wrapper/json tested above. Check that a specific entry is present.

I added these tests to storageTests because when starting with just loading into the entries into the DB, that's where I thought it made sense. Now that it includes the auth check, it might make more sense to move that part of the testing to netTests, since token_list.py exists in net/auth

I included a json and simple text file in a new dir in storageTest to provide a token list to load from. I'm unsure if we want to keep this format of loading in the tokens since we are moving away from using a json to using the DB, but they need to come from somewhere.

aGuttman:
This PR is addressing https://github.com/e-mission/e-mission-docs/issues/790

emission/net/auth/token_list.py
```diff
@@ -10,21 +10,13 @@
...
-        key_file = open('conf/net/auth/token_list.json')
-        key_data = json.load(key_file)
-        key_file.close()
-        self.token_list_file = key_data["token_list"]
-        with open(self.token_list_file) as tlf:
-            raw_token_list = tlf.readlines()
-        self.token_list = [t.strip() for t in raw_token_list]
-        raw_token_list = None
+        tq.insert_from_json(source, userid)
+        self.token_list = tq.get_all_tokens()
```
shankari, line 19: @aGuttman this is not what we want.
Recall that the original problem from https://github.com/e-mission/e-mission-docs/issues/790 was that
> Container image has the files baked in, so we would need to rebuild it every time there is a new deployment

Our goal behind storing it in the database is so that the containers can all be identical
If you load the data in the init for `emission/net/auth/`:
(a) it will be loaded every time we try to validate a token
(b) it will attempt to load from the container, which means that the file needs to be on the container, which means that the containers will need to be different anyway

The load needs to happen in a separate script as indicated in
https://github.com/e-mission/e-mission-docs/issues/790#issuecomment-1247238718

The code here should only *read* from the database.

emission/storage/decorations/token_queries.py
```diff
@@ -1,21 +1,28 @@
...
+    token_list_file = key_data["token_list"]
+    with open(token_list_file) as tlf:
+        raw_token_list = tlf.readlines()
+    token_list = [t.strip() for t in raw_token_list]
+    # raw_token_list = None
+
+    for e in token_list:
+        ent = {"user_id":userid, "token" : e}
+        insert(ent)
+
```
shankari, line 28: When this is moved into a separate script, we don't need two config files, the token list file can be passed in directly as an argument.

emission/tests/storageTests/token_test.py
```diff
@@ -0,0 +1,33 @@
...
+import uuid
+import json
+
+import emission.core.get_database as edb
+
+import emission.storage.decorations.token_queries as esdt
+import emission.net.auth.token_list as tl
+
+
+class TestPlaceQueries(unittest.TestCase):
```
shankari, line 12: the class should match the file name and the test names should follow the existing naming convention

emission/tests/storageTests/token_test_files/token_list.json
```diff
@@ -0,0 +1,3 @@
+{
+  "token_list": "emission/tests/storageTests/token_test_files/tokens.txt"
+}
```
shankari, line 3: This should be removed when the `load` function is refactored.

runSingleTestDebug.sh
```diff
@@ -7,4 +7,4 @@ set -e
 # rm: missing operand
 # Try `rm --help' for more information.
 # PYTHONPATH=. python -m unittest discover -s emission/tests -p Test*;
-PYTHONPATH=. python emission/tests/netTests/TestPush.py
+PYTHONPATH=. python emission/tests/storageTests/token_test.py
```
shankari, line 10: Revert since it is not necessary.

shankari (changes requested):

aGuttman:
Got it. These should be easy changes since the functionality is all there now, but needs to be moved around to serve the correct purpose in the correct places. I will come back to it at the end of the day.

emission/net/auth/token_list.py
aGuttman: Where in the directory structure should this new script live?

shankari: scripts are, per convention, in `bin`
you should be able to find plenty of examples there that also use `argparse` for high quality command line processing.

emission/storage/decorations/token_queries.py
shankari: using `argparse` as I suggested earlier.

shankari:
> I included a json and simple text file in a new dir in storageTest to provide a token list to load from. I'm unsure if we want to keep this format of loading in the tokens since we are moving away from using a json to using the DB, but they need to come from somewhere.

As I indicated, the json is just a pointer to the text file, so we can pass in the text file location to the script directly and avoid the JSON completely. you should simplify the test as well after changing this.

shankari:
@aGuttman any updates on this so we can finish it off?

aGuttman:
I have the script written and tested, but not thoroughly. Needed to clear up the unit testing and commit. Been away from my computer since Friday evening but I was planning on finishing up tonight.

aGuttman:
New script `bin/auth/insert_tokens.py` loads tokens in DB from a simple text file. Auth in `token_list.py` only reads. `token_queries` no longer has insert from json functionality.

Tests added to `storageTest/TokenQueriesTest` for `token_queries` insert and retrieve functionality, `token_list` auth success and failure, `insert_tokens` script's individual functions, full script run, and full script run with auth success and failure.

I feel like the tests are super messy and should probably be split up, but I hope that they adequately cover the behavior and that the behavior is the desired one this time.

bin/auth/insert_tokens.py
```diff
@@ -0,0 +1,25 @@
...
+
+def get_tokens_from_file(file):
+    with open(file) as tlf:
+        raw_token_list = tlf.readlines()
+    token_list = [t.strip() for t in raw_token_list]
+    return(token_list)
+
+def load_tokens_into_db(token_list, userid):
+    for t in token_list:
+        esdt.insert({'user_id':userid,"token":t})
```
shankari, line 13: Why do we need to associate a `user_id` with the token, particularly given that the `user_id`, as passed in, is a single value for the entire list?

emission/storage/decorations/token_queries.py
```diff
@@ -11,18 +11,3 @@ def get_all_tokens():
 
 def insert(entry):
     edb.get_token_db().insert_one(entry)
-
-def insert_from_json(jsonfile, userid=0):
```
shankari: nit: I would probably define a `bulk_insert` here as well and call that from the script, since the bulk insert is the common case.

shankari (changes requested):
Almost there: My only question is around why the tokens need to be associated with a UUID, given that you don't use the UUID either for retrieval or after retrieval.

emission/tests/netTests/TestAuthSelection.py
```diff
@@ -20,6 +20,7 @@
 import emission.net.auth.skip as enas
 import emission.net.auth.token_list as enat
 
+import bin.auth.insert_tokens as bai
```
shankari, line 23: Moving the load/insert code into the wrapper will ensure that we don't need to import code from bin.

bin/auth/insert_tokens.py
aGuttman: This was something motivated by following examples I saw in the tests. The user_id is used to delete DB entries between tests, is what I thought I noticed. I don't know everything that goes on in the unit tests, I assume that they are meant to be kept as independent and contained as possible, but if there was reason to keep some things in the DB between tests, you needed to clear out some but not the others, and it looked like that was done via lines like `edb.get_analysis_timeseries_db().delete_many({'user_id': self.testUserId})`. At some point I changed my test to just `delete_many({})`, I don't remember when, and it worked fine, but I forgot to do away with the user_ids everywhere else.

I followed that pattern when building out the functions, and when I went to write the script, I think I just saw that there was room to provide this, so I put it in without thinking too much about it. I wasn't even sure I needed it when following the test examples in the first place, I can easily get rid of it.

shankari: I see. In general, the data in the timeseries and analysis timeseries collection are tied to an individual who is represented by the UUID.  That is why we insert for a particular UUID and then delete for the same UUID. In the case of the tokens, I don't see a specific need for them to be associated with a user.

In general, I would prefer for the DB design to not be driven solely by the unit tests; as you pointed out, you are already deleting everything `delete_many({})` anyway. If you wanted to be a bit more careful about that, so that inadvertently running the tests against an existing database would not break it, you could make sure to only delete the values that you insert.

This would have the same function as the current UUID-based deletion for the other unit tests.

emission/tests/storageTests/TestTokenQueries.py
```diff
@@ -0,0 +1,91 @@
...
+import emission.core.get_database as edb
+
+import emission.storage.decorations.token_queries as esdt
+import emission.net.auth.token_list as enat
+
+import bin.auth.insert_tokens as bai
+
+class TestTokenQueries(unittest.TestCase):
+    def setUp(self):
+        self.testUserId = uuid.uuid3(uuid.NAMESPACE_URL, "mailto:test@test.me")
```
shankari, line 17: We don't need the `testUserId` any more

```suggestion
```

emission/tests/storageTests/TestTokenQueries.py
```diff
@@ -0,0 +1,91 @@
...
+        edb.get_token_db().delete_many({})
```
shankari, line 18: This should really be in `teardown`

```suggestion

def tearDown(self):
    # delete all the tokens added here so that if we inadvertently run the tests against a real database, we don't
    # lose real data. There doesn't seem to be a pre-defined list of tokens - there is one file but there are also tests
    # that add `1`, `2` etc. So let's delete all single character tokens
    # so something like https://stackoverflow.com/questions/29577713/string-field-value-length-in-mongodb
    pass
```

shankari (changes requested):
@aGuttman Just a few small changes. I have added code for most of them; you just need to implement the replacement to `delete_many({})`

emission/tests/storageTests/TestTokenQueries.py
```diff
@@ -10,12 +10,15 @@
...
-    def setUp(self):
-        self.testUserId = uuid.uuid3(uuid.NAMESPACE_URL, "mailto:test@test.me")
-        edb.get_token_db().delete_many({})
+
+    def tearDown(self):
+    #All tests insert tokens of length one. Delete them once the test is done.
+    #Shouldn't affect "real" tokens with greater length that might be in the DB
+        for t in edb.get_token_db().find({}):
+            if len(t["token"]) == 1:
+                edb.get_token_db().delete_one(t)
```
shankari, line 21: I am merging this right now, but would like to see some description of how you tested that this works.

shankari (approved):

PR #886, opened 2022-11-03
Support dynamic conf auth method selection

aGuttman:
Support a webapp.conf:server:auth = dynamic case by resolving dynamic into skip or token_list by reading from the online dynamic conf of the given deployment

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -59,6 +59,7 @@ except:
     config_file = open('conf/net/api/webserver.conf.sample')
 
 OPENPATH_URL="https://www.nrel.gov/transportation/openpath.html"
+STUDY_CONFIG = os.getenv('STUDY_CONFIG', "stage-program")
 
 config_data = json.load(config_file)
 config_file.close()
@@ -482,6 +483,25 @@ def getUUID(request, inHeader=False):
         traceback.print_exc()
         abort(403, e)
 
+def resolve_auth(auth_method):
+    if auth_method == "dynamic":
+        download_url = "https://raw.githubusercontent.com/e-mission/nrel-openpath-deploy-configs/main/configs/" + STUDY_CONFIG + ".nrel-op.json"
+        print("About to download config from %s" % download_url)
+        r = requests.get(download_url)
+        if r.status_code is not 200:
+            print(f"Unable to download study config, status code: {r.status_code}")
+            sys.exit(1)
+        else:
+            dynamic_config = json.loads(r.text)
+            print(f"Successfully downloaded config with version {dynamic_config['version']} "\
+                f"for {dynamic_config['intro']['translated_text']['en']['deployment_name']} "\
+                f"and data collection URL {dynamic_config['server']['connectUrl']}")
+        if dynamic_config["intro"]["program_or_study"] == "program":
+            return "token_list"
+        else:
+            return "skip"
+    else:
+        return auth_method
 # Auth helpers END
 
 if __name__ == '__main__':
@@ -489,6 +509,8 @@ if __name__ == '__main__':
         webserver_log_config = json.load(open("conf/log/webserver.conf", "r"))
     except:
         webserver_log_config = json.load(open("conf/log/webserver.conf.sample", "r"))
+
+    auth_method = resolve_auth(auth_method)
 
     print(f"Using auth method {auth_method}")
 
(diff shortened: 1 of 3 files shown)


aGuttman:
addresses https://github.com/e-mission/e-mission-docs/issues/790

aGuttman:
Needs to be end-to-end tested by creating image, deploying with image, copying existing tokens in token collection, attempting to access through phone app with a valid token and an invalid token.

emission/net/api/cfc_webapp.py
```diff
@@ -59,6 +59,7 @@
     config_file = open('conf/net/api/webserver.conf.sample')
 
 OPENPATH_URL="https://www.nrel.gov/transportation/openpath.html"
+STUDY_CONFIG = os.getenv('STUDY_CONFIG', "stage-program")
```
shankari, line 62: You should clarify how the `STUDY_CONFIG` is set and add a sample.

emission/net/api/cfc_webapp.py
```diff
@@ -482,6 +483,25 @@ def getUUID(request, inHeader=False):
         traceback.print_exc()
         abort(403, e)
 
+def resolve_auth(auth_method):
```
shankari, line 486: Seems like you can add a unit test for `auth_method` that uses one of the study configs that always exist - e.g. devapp/stage.

shankari (changes requested):
The code itself looks fine; we just need some samples and unit tests.

emission/net/api/cfc_webapp.py
shankari: You should document what happens if the study config is not set, so that we can check for it when we deploy and fix ASAP if it is broken.

shankari:
- need to get the case where the environment variable is set
- need to configure the environment variable in `docker-compose` and/or `Dockerfile` to indicate that we expect this environment variable to be passed in

emission/tests/netTests/TestWebserver.py
```diff
@@ -58,7 +59,7 @@ def tearDown(self):
 
     def test404Redirect(self):
         from emission.net.api.bottle import response
-        import emission.net.api.cfc_webapp as enacw
+        importlib.reload(enacw)
```
shankari, line 62: why is this an `importlib.reload` instead of a direct `import`?

emission/tests/netTests/TestWebserver.py
```diff
@@ -67,6 +68,20 @@ def test404Redirect(self):
         self.assertEqual(response.status_code, 301)
         self.assertEqual(response.get_header("Location"), "http://somewhere.else")
 
+    def testResolveAuth(self):
+        import emission.net.api.cfc_webapp as enacw
```
shankari, line 72: given that we are doing a direct import here?

shankari (changes requested):
Should make the import vs. importlib be consisent.

emission/tests/netTests/TestWebserver.py
aGuttman: This reload is needed. Though the imported module isn't scoped, it seems the name is, (I'm not sure about this, but just through use that seems to be the case) so the import of `emission.net.api.cfc_webapp` goes at the top so that it can be used both here and in `testResolveAuthWithEnvVar`.

It needs to be reloaded for this test since the initial import was done before the config was set.

emission/tests/netTests/TestWebserver.py
aGuttman: This import probably isn't needed. I don't think this test is sensitive to the import and env var that is left unset. I'll try getting rid of it and see what happens.

aGuttman: Yeah, this import can be deleted. It isn't doing anything since the import happened globally and there isn't a need to have anything reloaded for this test.

shankari (changes requested):
What about adding the ENV to the Dockerfile?

aGuttman:
> What about adding the ENV to the Dockerfile?

I feel like it would be correct place to put this is in the cloud-computing openpath docker-compose as well. We could add it to the Dockerfile here as a reminder to those using this repo, but the deploy happens with the internal repo so should it be indicated there somehow.

shankari:
> I feel like it would be correct place to put this is in the cloud-computing openpath docker-compose as well. We could add it to the Dockerfile here as a reminder to those using this repo, but the deploy happens with the internal repo so should it be indicated there somehow.

Agreed in general. There are two separate files:
- Dockerfile: where we indicate that this variable is available to set. that should be in this repo since the code that uses it is in this repo
- `docker-compose`: there is no docker-compose in this repo. The example docker-compose is in `e-mission-docker` and in the internal openpath repo. Both of those docker-compose examples should also set this environment variable; again as an example of the set of values.

shankari (approved):
The changes to the `docker-compose` are in other repos, so this can be merged.
@aGuttman I assume you will make those changes as well.

PR #888, opened 2022-11-10
Update cfc_webapp.py

aGuttman:
Replace prints with logs around auth_method resolution


shankari (approved):

aGuttman:
Tested locally with standard running of cfc_webapp.py

PR #889, opened 2022-11-11
Update cfc_webapp.py

aGuttman:
If on staging, set auth_method to token_list for testing purposes.

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -497,7 +497,7 @@ def resolve_auth(auth_method):
             logging.debug(f"Successfully downloaded config with version {dynamic_config['version']} "\
                 f"for {dynamic_config['intro']['translated_text']['en']['deployment_name']} "\
                 f"and data collection URL {dynamic_config['server']['connectUrl']}")
-            
+
             if STUDY_CONFIG.startswith('stage-'):
                 logging.debug("On staging, use token_list for testing purposes")
                 return "token_list"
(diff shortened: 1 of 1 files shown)


shankari:
FYI, the reason that your logs were not showing up is because you have them before the logging is configured (in line 570).

```
564     logging.debug("attempting to resolve auth_method")
565     auth_method = resolve_auth(auth_method)
566
567     logging.debug(f"Using auth method {auth_method}")
568     print(f"Using auth method {auth_method}")
569
570     logging.config.dictConfig(webserver_log_config)
571     logging.debug("This should go to the log file")
```

shankari:
Please change the ordering so that your logs do show up and then I will merge.

emission/net/api/cfc_webapp.py
```diff
@@ -514,15 +519,13 @@ def resolve_auth(auth_method):
...
 
+    logging.config.dictConfig(webserver_log_config)
     logging.debug("attempting to resolve auth_method")
     auth_method = resolve_auth(auth_method)
 
     logging.debug(f"Using auth method {auth_method}")
     print(f"Using auth method {auth_method}")
 
-    logging.config.dictConfig(webserver_log_config)
-    logging.debug("This should go to the log file")
```
shankari, line 524: Is there a reason you removed this line?

aGuttman: Didn't really think about it too much. Just seemed unneeded, `logging.debug("attempting to resolve auth_method")` follows the configuration, so that showing up would indicate it instead.

shankari:
As a nit: it would be good have a more detailed commit message than using "Update cfc_webapp.py" every time.

PR #890, opened 2022-12-01
Replacement model

zackAemmer:
Implementation of a replacement mode model in the same style as GreedySimilarityBinning and previous TripModels. Model comparisons are in a notebook here:
https://github.com/e-mission/em-public-dashboard/pull/76

This code is the implementation of the best model from that notebook. The functionality of this code is tested through unit testing in tests/modellingtests, but should be tested with the actual server code separately (todo).


emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,122 @@
...
+    def extract_features(self, trips: ecwc.Confirmedtrip) -> List[float]:
+        # TODO: need to enable generic paths other than just user input for features
+        X = pd.DataFrame(
+            [[trip['data']['user_input'][x] for x in self.feature_list] for trip in trips],
+            columns=self.feature_list
+        )
+        y = pd.DataFrame(
+            [trip['data']['user_input'][self.dependent_var] for trip in trips],
+            columns=[self.dependent_var]
+        )
```
shankari: This can be a series, right? That will conform more closely to the standard ML formulation of an X matrix and a y vector.

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def from_dict(self, model: Dict):
+        self.gbdt = model
+
+    def extract_features(self, trips: ecwc.Confirmedtrip) -> List[float]:
+        # TODO: need to enable generic paths other than just user input for features
+        X = pd.DataFrame(
+            [[trip['data']['user_input'][x] for x in self.feature_list] for trip in trips],
+            columns=self.feature_list
+        )
```
shankari: Aren't some (most?) of your input features demographic? How can you read them from the user input in that case?

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,122 @@
...
+    def _process_data(self, X, y):
+        """
+        helper function to transform binned features and labels.
+        """
+        # Any non-numeric dtype must be one-hot encoded (if unordered) or numerically coded (if ordered)
+        dummies = []
+        for col in X:
+            if X[col].dtype=='object':
+                dummies.append(pd.get_dummies(X[col], prefix=col))
+        X = pd.concat(dummies, axis=1)
```
shankari: Interesting, I did not know about `pandas.get_dummies` before. 
What are the columns that are affected by this?

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,122 @@
...
+        # The outcome must be a single categorical column; recode to numeric
+        for col in y:
+            cat_list = list(pd.unique(y[col])).sort()
+            if y[col].dtype=='object':
+                y[col] = pd.Categorical(y[col], ordered=True, categories=cat_list)
+                y[col] = y[col].cat.codes
```
shankari: Why not use dummies here as well? It is because it will end up as a dataframe instead of a vector? Adding some comments here with your rationale will help future maintainers.

emission/tests/modellingTests/TestGradientBoostedDecisionTree.py
```diff
@@ -0,0 +1,76 @@
...
+        # pass in a test configuration
+        model_config = {
+            "incremental_evaluation": False,
+            "feature_list": ['mode_confirm','purpose_confirm'],
+            "dependent_var": 'replaced_mode'
+        }
+        model = eamtg.GradientBoostedDecisionTree(model_config)
+        model.fit(trips)
+        model.predict(trips)
+
```
shankari, line 48: in unit testing, you want to check that the results are consistent with what you expect. Since you have fixed input, if you use a fixed algorithm and a fixed random seed, you should get the same results every time.

emission/tests/modellingTests/TestGradientBoostedDecisionTree.py
```diff
@@ -0,0 +1,76 @@
...
+    def testUnseenTrainingClasses(self):
+        """
+        if a new class is added the model should re-train
+        """
```
shankari, line 104: don't see where you are actually adding the new class?

shankari (commented):
@zackAemmer this is a great start and you are definitely on the right track. Please see my high level review below.

emission/analysis/modelling/trip_model/util.py
```diff
@@ -1,7 +1,49 @@
...
+    :param trips: all trips to extract features from
+    :type trips: List[ecwc.Confirmedtrip]
+    :return: the training X features and y for the replacement mode model
+    :rtype: Tuple[List[List[float]], List[]]
+    """
+    # get dataframe from json trips; fill in calculated columns
+    trips_df = pd.json_normalize(trips)
+    # distance
+    trips_coords = trips_df[['data.start_loc.coordinates','data.end_loc.coordinates']]
+    trips_df['distance_miles'] = trips_coords.apply(lambda row : haversine(row[0],row[1]), axis=1)
```
shankari, line 32: each trip already has its distance in meters. you should just use it directly instead of calculating the haversine, which is more intensive, and which does not even take the trajectory into count.

shankari: Also, this may be a pet peeve, but why are we using `miles` internally? Sure, the American public is stuck in the imperial system, but when we do science under the hood, surely we can use SI units.

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,122 @@
+import logging
+from tokenize import group
+from typing import Dict, List, Optional, Tuple
+
+import numpy as np
+import pandas as pd
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.ensemble import GradientBoostingClassifier
```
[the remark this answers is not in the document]
zackAemmer: Updated to use see

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,122 @@
...
+        https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingClassifier.html
+
+        Replacement modes are considered to be the second-best choice for
+        a given trip (i.e., what mode would have been chosen if the actual
+        choice wasn't available).
+
+        The model is currently trained on data from all users.
+        """
+        if config is None:
+            config = eamtc.get_config_value_or_raise('model_parameters.gbdt')
```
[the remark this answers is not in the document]
zackAemmer: Added config file

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
zackAemmer: Is now a Series

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
zackAemmer: My thought is that this should work on strings, dates, mixed columns, anything unordinary but especially columns that are categorical/strings, which will not work as features for sklearn. I automatically convert them to dummy variables, assuming that if they were part of the specification, they were put there intentionally. I could see this going wrong if a column was specified that had many different responses such as a user input which would lead to a very large, sparse set of dummy variables. However the model would still train and not break. It may also lead to the same effect if there were a numeric column that had one string in it or something along those lines.

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
zackAemmer: Yes, added comments here and also changed this implementation a bit. Sklearn will accept strings for categorical y variables but not for X features.

emission/tests/modellingTests/TestGradientBoostedDecisionTree.py
zackAemmer: Good point I didn't think of that, added tests that check the actual model outputs

emission/tests/modellingTests/TestGradientBoostedDecisionTree.py
zackAemmer: Changed this and other code to use a config file for feature inputs

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -71,11 +63,12 @@ def fit(self, trips: List[ecwc.Confirmedtrip]):
         X_train, y_train = self.extract_features(trips)
         self.gbdt.fit(X_train, y_train)
         logging.info(f"gradient boosted decision tree model fit to {len(X_train)} rows of trip data")
+        logging.info(f"training features were {X_train.columns}")
 
-    def predict(self, trip: ecwc.Confirmedtrip) -> Tuple[List[Dict], int]:
+    def predict(self, trip: ecwc.Confirmedtrip) -> List:
```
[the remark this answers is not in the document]
zackAemmer: A list of trips can be passed (so that predict can be called on a batch of unlabeled trips all at once). So maybe this should be trip: List[ecwc.Confirmedtrip] -> List[string] ?

emission/analysis/modelling/trip_model/support_vector_machine.py
```diff
@@ -0,0 +1,117 @@
...
+        For anyone looking to implement a differnt online learning model in the 
+        future, here is a list of sklearn models that implement "partial_fit" 
+        and would be candidates for the online learning approach implemented 
+        here:
+
+        https://scikit-learn.org/0.15/modules/scaling_strategies.html
+
+        Replacement modes are considered to be the second-best choice for
+        a given trip (i.e., what mode would have been chosen if the actual
+        choice wasn't available).
```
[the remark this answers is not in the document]
zackAemmer: Updated intro descriptions for svm and gbdt

emission/analysis/modelling/trip_model/support_vector_machine.py
```diff
@@ -0,0 +1,117 @@
...
+        elif self.is_incremental and self.incremental_classes is not None:
+            logging.debug(f'updating incremental model fit')
+            self.svm.partial_fit(X_train, y_train)
+        # If not incremental, train regularly
+        else:
+            self.svm.fit(X_train, y_train)
+        logging.info(f"support vector machine model fit to {len(X_train)} rows of trip data")
+        logging.info(f"training features were {X_train.columns}")
+
+    def predict(self, trip: ecwc.Confirmedtrip) -> List[int]:
```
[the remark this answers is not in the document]
zackAemmer: After finding that sklearn can predict on dependent variables with strings this is List[string]

emission/analysis/modelling/trip_model/util.py
zackAemmer: Changed, you have my full and earnest agreement.

emission/analysis/modelling/trip_model/util.py
```diff
@@ -39,3 +81,26 @@ def find_knee_point(values: List[float]) -> Tuple[float, int]:
             index = i
     value = values[index]
     return [index, value]
+
+
+# if the non-mock trips have distance calculated then this can be removed
+# https://stackoverflow.com/questions/4913349/haversine-formula-in-python-bearing-and-distance-between-two-gps-points
+def haversine(coord1, coord2):
```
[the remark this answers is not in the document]
zackAemmer: Switched to distance formula from common, also only calling from test code now

emission/tests/modellingTests/TestGradientBoostedDecisionTree.py
```diff
@@ -34,16 +36,71 @@ def testSmoke(self):
...
         }
         model = eamtg.GradientBoostedDecisionTree(model_config)
         model.fit(trips)
         model.predict(trips)
 
-    def testUnseenTrainingClasses(self):
+
+    def testUnseenFeatures(self):
+        """
+        if the input classes for a feature change throw sklearn error
```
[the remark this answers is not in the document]
zackAemmer: Made change

emission/tests/modellingTests/TestGradientBoostedDecisionTree.py
```diff
@@ -65,12 +122,55 @@ def testUnseenTrainingClasses(self):
...
+            "feature_list": [
+                'data.user_input.mode_confirm',
+                'data.user_input.purpose_confirm',
+                'distance_miles'
+            ],
+            "dependent_var": 'data.user_input.replaced_mode'
         }
         model = eamtg.GradientBoostedDecisionTree(model_config)
         model.fit(trips)
         model.predict(trips)
```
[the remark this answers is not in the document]
zackAemmer: Added test for actual recoding

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
```diff
@@ -0,0 +1,87 @@
...
+            'incremental_evaluation',
+            'feature_list',
+            'dependent_var'
+        ]
+        for k in expected_keys:
+            if config.get(k) is None:
+                msg = f"gbdt trip model config missing expected key {k}"
+                raise KeyError(msg)
+        self.is_incremental = config['incremental_evaluation']
+        # use the sklearn implementation of a GBDT
```
shankari, line 48: for my own edification, the GBDT is not incremental, right?

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
shankari: This seems to be unchanged; I assume you will tackle it next?

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
shankari: Good job adding a related comment in the code.

emission/tests/modellingTests/TestSupportVectorMachine.py
```diff
@@ -0,0 +1,460 @@
...
+import pandas as pd
+import random
+
+
+class TestSupportVectorMachine(unittest.TestCase):
+
+    def setUp(self) -> None:
+        logging.basicConfig(format='%(asctime)s:%(levelname)s:%(message)s',
+        level=logging.DEBUG)
+
```
shankari, line 14: Aren't these very similar to the `TestGradientBoostedDecisionTree.py`? Can't you just refactor into one set of tests where you pass in the model? Except for the partial fit (which makes sense as a separate test), it seems like both of them have the same inputs and outputs.

emission/analysis/modelling/trip_model/support_vector_machine.py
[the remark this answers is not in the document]
shankari: I still don't understand, even with the updated description.
Let's discuss today.

emission/analysis/modelling/trip_model/gradient_boosted_decision_tree.py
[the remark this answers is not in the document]
shankari: Sure, processing a batch of trips at a time is definitely going to be more efficient than processing them one at a time.

conf/analysis/trip_model.conf.json.sample
```diff
@@ -8,6 +8,56 @@
...
             "incremental_evaluation": false
+        },
+        "gbdt": {
+            "feature_list": {
+                "data.user_input.mode_confirm": [
+                    "drive",
+                    "bike",
+                    "walk",
+                    "transit"
+                ],
```
shankari, line 19: As we discussed, we should not be using `mode_confirm` and `purpose_confirm` to infer the replaced mode because in practice people either label all or none. So we are unlikely to have the `mode_confirm` and `purpose_confirm` for trips that we want to predict the replaced mode for.

emission/analysis/modelling/trip_model/support_vector_machine.py
```diff
@@ -0,0 +1,117 @@
...
+        corresponds to a label at the matching index of the label input.
+
+        If using an incremental model, the initial call to fit will store 
+        the list of unique classes in y. If additional classes are added in the 
+        future, the model will need to be re-initialized on data that 
+        contains all of those classes, otherwise sklearn will throw an 
+        error. Additionally, any categorical variables used as features 
+        must have all classes present in the initial training data. If not, 
+        they will throw an sklearn error due to differences in the number of 
+        OHE variables in the data.
```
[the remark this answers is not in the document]
shankari: Initial training set: [walk, bike]
new data comes in with: [drive], so we retrain, and new training set has [walk, bike, drive]
new data comes in with: [walk], but because the current model has [walk, bike, drive], it again throws an error

zackAemmer:
Major Todo:
-Combine the SVM and GBDT tests since there is so much repeated code
-Switch over to "divergent" modeling approach which stores a dictionary of features/classes in the model instance (not the config)
-Return to replacement modeling notebooks, and switch to models that do not rely on any user labels (then update GBDT/SVM tests to reflect this)
-Write some more tests for the demographics; what if user didn't answer a question, also write documentation for new function used to extract demographic features.

shankari:
@zackAemmer I am reviewing the changes but note that your tests are failing due to an invalid config.

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -189,22 +200,23 @@ def generate_mock_trips(
...
-def add_trip_demographics(trip):
-    trip['data']['survey'] = {}
-    survey_features = {
-        'hhinc':['0-24999','25000-49000','50000-99999','100000+'],
-        'age':[x for x in range(0, 70)],
-        'veh':['0','1','2','3','4+']
-    }
+def add_trip_demographics(trip, survey_features):
+    response_id = ''.join(random.choices(string.ascii_uppercase + string.ascii_lowercase, k=22))
+    trip['data']['jsonDocResponse'] = {response_id: {'group_yk8eb99': {}, 'group_hg4zz25': {}, 'group_pa5ah98': {}}}
```
shankari, line 213: again, not sure why the demographic responses are associated with trips? You don't seem to be `copy()`ing the data, so presumably we are only storing links to the same demographic information, not actually creating duplicates that blow up the memory usage

shankari (commented):
High level comment: not sure why the demographic data needs to be part of the trip information. Detailed comments below...

zackAemmer:
**Remaining Todo:**
-The model tests for the GBDT/SVM are combined, but are not set up to use 'inferred_labels', instead looking for input from 'user_input' (e.g., data.user_input.mode_confirm).
-The demographics look for a UUID from the database, (which was a mistake, since there are no users in the test database). So fake data needs to be inserted into the test database then queried from the util.py "get_survey_df" function. Most likely this should be split into 2 functions, one untested which reads the data, the other tested which turns it into sklearn-compatible features.

shankari:
@zackAemmer have you also pushed your changes to the public dashboard notebook?

PR #891, opened 2022-12-08
Build out token script

aGuttman:
Built out token script to include adding tokens by file, one at a time and by uuids 'user_email' field (as was done on staging). Also added a show argument to list tokens currently in DB.

Created and ran unit tests for new script functionality.

### DIFF
--- a/bin/auth/insert_tokens.py
+++ b/bin/auth/insert_tokens.py
@@ -1,12 +1,31 @@
 import argparse
 import emission.storage.decorations.token_queries as esdt
+import emission.core.get_database as edb
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(prog="insert_tokens")
 
-    parser.add_argument("file")
+    parser.add_argument("--file", help="add tokens to DB from file")
+    parser.add_argument("--single", help="add a single token")
+    parser.add_argument("--uuid", help="add tokens to the DB off of the email field of UUIDS", action="store_true")
+    parser.add_argument("--show", help="show tokens that are in the DB", action="store_true")
 
     args = parser.parse_args()
 
-    token_list = (esdt.get_tokens_from_file(args.file))
-    esdt.insert_many_tokens(token_list)
+    if args.file:
+        token_list = (esdt.get_tokens_from_file(args.file))
+        esdt.insert_many_tokens(token_list)
+
+    if args.single:
+        token = args.single
+        esdt.insert({"token":token})
+
+    if args.uuid:
+        uuid_list = edb.get_uuid_db().find()
+        for u in uuid_list:
+            esdt.insert({"token":u["user_email"]})
+
+    if args.show:
+        token_list = esdt.get_all_tokens()
+        for t in token_list:
+            print(t)
--- a/bin/auth/insert_tokens.py
+++ b/bin/auth/insert_tokens.py
@@ -4,11 +4,11 @@ import emission.core.get_database as edb
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(prog="insert_tokens")
-
-    parser.add_argument("--file", help="add tokens to DB from file")
-    parser.add_argument("--single", help="add a single token")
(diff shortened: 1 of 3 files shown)


aGuttman:
Tested locally with new unit tests on a fresh mongoDB container.

bin/auth/insert_tokens.py
```diff
@@ -1,12 +1,31 @@
...
+import emission.core.get_database as edb
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(prog="insert_tokens")
 
-    parser.add_argument("file")
+    parser.add_argument("--file", help="add tokens to DB from file")
+    parser.add_argument("--single", help="add a single token")
+    parser.add_argument("--uuid", help="add tokens to the DB off of the email field of UUIDS", action="store_true")
+    parser.add_argument("--show", help="show tokens that are in the DB", action="store_true")
```
shankari, line 11: what happens if I call this with no arguments?
what happens if I call it with multiple arguments?

You should add unit tests for those cases as well.

Hint: Have you considered using a mutually exclusive group?

emission/tests/storageTests/TestTokenQueries.py
```diff
@@ -73,21 +76,73 @@ def test_auth_fail_file_insert(self):
...
 
-    def test_run_script(self):
-        os.system("python3 bin/auth/insert_tokens.py emission/tests/storageTests/token_test_files/tokens.txt")
+    def test_run_script_file(self):
+        os.system("python3 bin/auth/insert_tokens.py --file emission/tests/storageTests/token_test_files/tokens.txt")
         self.assertEqual(esdt.get_all_tokens(), ['a','b','c','d'])
 
-    def test_run_script_auth(self):
-        os.system("python3 bin/auth/insert_tokens.py emission/tests/storageTests/token_test_files/tokens.txt")
+    def test_run_script_file_auth(self):
```
shankari, line 83: what is the difference between this and `test_run_script_file`? you might want to add some comments explaining this...

emission/tests/storageTests/TestTokenQueries.py
```diff
@@ -73,21 +76,73 @@ def test_auth_fail_file_insert(self):
...
+        os.system("python3 bin/auth/insert_tokens.py --file emission/tests/storageTests/token_test_files/tokens.txt")
         tlm = enat.TokenListMethod()
         self.assertEqual(tlm.verifyUserToken('b'), 'b')
 
-    def test_run_script_auth_fail(self):
-        os.system("python3 bin/auth/insert_tokens.py emission/tests/storageTests/token_test_files/tokens.txt")
+    def test_run_script_file_auth_fail(self):
```
shankari, line 88: Ditto.

emission/tests/storageTests/TestTokenQueries.py
```diff
@@ -73,21 +76,73 @@ def test_auth_fail_file_insert(self):
...
+        with self.assertRaises(ValueError):
+            tlm = enat.TokenListMethod()
+            tlm.verifyUserToken('z')
+
+    def test_run_script_show(self):
+        esdt.insert({'token':'x'})
+        esdt.insert({'token':'y'})
+        esdt.insert({'token':'z'})
+        os.system("python3 bin/auth/insert_tokens.py --show")
+        self.assertEqual(True, True)
```
shankari, line 144: Nit: you should be able to read the script result by opening the stdout of the process as a file and reading from it.
https://docs.python.org/3/library/subprocess.html#subprocess.run

This may be too overkill for the current code, but I would encourage you to figure it out as a learning experience in your spare time.

shankari (changes requested):

emission/tests/storageTests/TestTokenQueries.py
aGuttman: It's the same, I just renamed it to specify what argument it is testing since I added new options.

shankari (approved):

PR #892, opened 2022-12-23
Replace mode pipeline

aGuttman:


aGuttman:
Addressing https://github.com/e-mission/e-mission-docs/issues/841

Start building out infrastructure to allow replace mode model to run in pipeline. Functions made to use gbdt model through trip_model interface, create storage methods for model.

aGuttman:
todo:
continue building/testing infrustructure components
write save/update functions for model
create unit tests

emission/analysis/classification/inference/labels/pipeline_replace_mode.py
```diff
@@ -0,0 +1,102 @@
...
+}
+
+# ensemble specifies which algorithm in eacile to run.
+# This makes it easy to test various ways of combining various algorithms.
+ensemble = eacile.ensemble_first_prediction
+
+
+# Does all the work necessary for a given user
+def infer_labels(user_id):
+    time_query = epq.get_time_range_for_label_inference(user_id)
```
shankari, line 30: again, this is not the time range to query for because that will return the time range for the label inference algorithm. You are your own algorithm and you need your own time range

shankari: This will break the pipeline unless changed.

emission/analysis/classification/inference/labels/pipeline_replace_mode.py
```diff
@@ -0,0 +1,102 @@
...
+        if lip.last_trip_done is None:
+            logging.debug("After run, last_trip_done == None, must be early return")
+        epq.mark_label_inference_done(user_id, lip.last_trip_done)
+    except:
+        logging.exception("Error while inferring labels, timestamp is unchanged")
+        epq.mark_label_inference_failed(user_id)
+
+# Code structure based on emission.analysis.classification.inference.mode.pipeline
+# and emission.analysis.classification.inference.mode.rule_engine
+class LabelInferencePipeline:
```
shankari, line 44: again, this needs to change for clarity

emission/analysis/classification/inference/labels/pipeline_replace_mode.py
```diff
@@ -0,0 +1,102 @@
...
+    # to the database, and records progress
+    def run_prediction_pipeline(self, user_id, time_range):
+        self.ts = esta.TimeSeries.get_time_series(user_id)
+        self.toPredictTrips = esda.get_entries(
+            esda.CLEANED_TRIP_KEY, user_id, time_query=time_range)
+        for cleaned_trip in self.toPredictTrips:
+            # Create an inferred trip
+            cleaned_trip_dict = copy.copy(cleaned_trip)["data"]
+            inferred_trip = ecwe.Entry.create_entry(user_id, "analysis/inferred_trip", cleaned_trip_dict)
+
```
shankari, line 62: you have basically copy-pasted the other pipeline.py, you need to understand how it works and adapt it to be a separate step.

emission/analysis/modelling/trip_model/run_model.py
```diff
@@ -118,6 +118,27 @@ def predict_labels_with_n(
         predictions, n = model.predict(trip)
         return predictions, n
 
+def predict_labels_with_gbdt(
```
shankari, line 121: where is this called from?

shankari (changes requested):
At a high level, I also don't see this new algorithm called from anywhere in the pipeline.

emission/analysis/modelling/trip_model/run_model.py
shankari: From the list of algorithms (but `pipeline_replace_mode.py` will do it later).

PR #893, opened 2022-12-31
insert_tokens: fix default message use

shankari:
Originally checked in by @aGuttman but not submitted.
PR created by @shankari for the record

Default message will no longer display when using options other than show

small typo in comment fixed


shankari:
without the change, when I inserted entries from a file, I kept getting an error

```
./e-mission-py.bash bin/auth/insert_tokens.py --file ~/in-house/openpath-qrcode-gen/smart_commute_ebike_token_list.txt
Please provide the script with an argument. Use the "--help" option for more details

```

it looks like the insert succeeded in spite of the error, so when I retried after merging, I ended up with multiple copies of the same token

PR #900, opened 2023-02-10
Updating Conda version

swastis10:
Updating Conda version


README.md
```diff
@@ -161,30 +161,6 @@ them into utility functions. Original versions of the notebooks can be obtained
 by looking at other notebooks with the same name, or by looking at the history
 of the notebooks.
 
-## JS Testing ##
-
-From the webapp directory
-
-    $ npm install karma --save-dev
-    $ npm install karma-jasmine karma-chrome-launcher --save-dev
-
```
shankari, line 170: I'm glad you took care of this, although in the future, I would prefer to have smaller commits, each of which focuses on one particular action. So I would have put this README into another commit albeit in the same PR.

setup/setup_conda.sh
```diff
@@ -7,12 +7,12 @@ if [[ -z $EXP_CONDA_VER || -z $PLATFORM ]]; then
...
     echo "   Assumes that the EXP_CONDA_VER variable is set"
     echo "   Platform options are Linux-x86_64, MacOSX-x86_64"
-    echo "   For Windows, manually download and install https://repo.anaconda.com/miniconda/Miniconda3-$EXP_CONDA_VER-Windows-x86_64.exe"
+    echo "   For Windows, manually download and install https://repo.anaconda.com/miniconda/Miniconda3-py39_$EXP_CONDA_VER-1-Windows-x86_64.exe"
 else
     INSTALL_PREFIX=$HOME/miniconda-$EXP_CONDA_VER
     SOURCE_SCRIPT="$HOME/miniconda-$EXP_CONDA_VER/etc/profile.d/conda.sh"
     
-    curl -o miniconda.sh -L https://repo.continuum.io/miniconda/Miniconda3-py38_$EXP_CONDA_VER-$PLATFORM.sh;
+    curl -o miniconda.sh -L https://repo.continuum.io/miniconda/Miniconda3-py39_$EXP_CONDA_VER-1-$PLATFORM.sh;
```
shankari, line 15: so the miniconda file format has now changed to include `-1`? Are there also `-2, -3`, etc?
What do they represent?

shankari (changes requested):
Looks good overall - please see minor comments below.
Main change is that we do not want to check in `miniconda.sh` - the reason that we have `setup_conda.sh` is so that we can download miniconda on the fly. I want to avoid checking in ginormous files unless we have to.

Just to reiterate, you *do not* have to move the README to a separate commit this time - that was just feedback for next time.

README.md
swastis10: I just removed miniconda.sh - Committed it by mistake

shankari:
@swastis10 why would you delete the `.gitignore`?

emission/tests/storageTests/TestMongoGeoJSONQueries.py
```diff
@@ -112,11 +112,11 @@ def getTestNegPolygon(self):
             [-123,38],[-123,37]]
 
   def testGeoWithNegativeValues(self):
-    test = {'track_location': {'type':'Point', 'coordinates': [-122.5,37.5]} }
+    test = {'track_location': {'type':'Point', 'coordinates': [122.5,37.5]} }
```
shankari, line 115: @swastis10 what is the motivation for this change?
I intentionally had it as a negative value to make sure that negative values worked. Note that all locations in the US have negative longitude. If this is breaking, we need to fix the code, not the test

shankari (commented):
Looks good to me so far. Glad to see that you reverted the negative geojson check
Tests are still failing though it looks like you have a pretty good handle on the fixes.

emission/tests/netTests/TestFormatters.py
```diff
@@ -13,6 +13,9 @@
 from uuid import UUID
 import attrdict as ad
 import geojson
+# This change should be removed in the next server update, by which time hopefully the new geojson version will incorporate the long-term fix for their default precision
+# See - jazzband/geojson#177
+geojson.geometry.Geometry.__init__.__defaults__ = (None, False, 15)
```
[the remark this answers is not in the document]
swastis10: done

swastis10:
@shankari This PR is ready for review

emission/analysis/intake/cleaning/filter_accuracy.py
```diff
@@ -100,7 +100,8 @@ def filter_accuracy(user_id):
                 logging.debug("Partial filtered points %d found" % len(filtered_points_df))
                 filtered_points_df = filtered_points_df[SEL_FIELDS_FOR_DUP]
                 filtered_points_df["msts"] = filtered_points_df.ts.apply(lambda x: int(x * 10**3))
-                matched_points_df = filtered_from_unfiltered_df.merge(filtered_points_df, on="msts", right_index=True)
+                matched_points_df = filtered_from_unfiltered_df.reset_index().merge(filtered_points_df, on="msts")
+                matched_points_df.set_index('index', inplace=True)
```
shankari, line 104: I'm a little confused about this. The issue comment that you linked to says that "on" can be removed,
> Therefore on can be removed from here.

but you have actually removed the `right_index`

shankari (commented):
@swastis10 just one requested clarification
otherwise LGTM!

shankari:
@swastis10 actually, one more question: there were a lot of failures of the type

> AttributeError: property sensed_mode is not defined for Confirmedtrip

e.g. https://github.com/e-mission/e-mission-docs/issues/856#issuecomment-1439372467

what was their cause and how did you resolve it?

emission/analysis/intake/cleaning/filter_accuracy.py
swastis10: @shankari 
According to the new version of Pandas, we cannot use 'left_index/right_index' and 'on' together so we can define the column on which we want to merge the dataframe and set the indexes using functions like set_index or reset_index.

In the testcase below in '[TestFilterAccuracy.py](https://github.com/e-mission/e-mission-server/pull/900/commits/a760b55cc7af16d95d6f9428a4fe0d7631b9c39d#diff-e9aaf12a6c1fcc5821ea6c95cf67a5223998ba7896dd25cfcdce1998983cd7e7)' there was only one column 'ts' therefore removing 'on' was also giving us the desired result but that won't be a good solution as we can have multiple columns in a dataframe.

swastis10:
> @swastis10 actually, one more question: there were a lot of failures of the type
> 
> > AttributeError: property sensed_mode is not defined for Confirmedtrip
> 
> e.g. [e-mission/e-mission-docs#856 (comment)](https://github.com/e-mission/e-mission-docs/issues/856#issuecomment-1439372467) as commented on further in [e-mission/e-mission-docs#856 (comment)](https://github.com/e-mission/e-mission-docs/issues/856#issuecomment-1441196286)
> 
> what was their cause and how did you resolve it?

@shankari, I did not do anything to resolve it. In [this](https://github.com/e-mission/e-mission-server/pull/900/commits/5ee9e5424eb60395ace49d0bde6d5694dad62b1e#diff-2d18a73b1f12befa9b206fe50ec9f3d3f89ff955e80a65c98b065c865f502fb6) commit, I was deleting and updating only one row from the database as opposed to multiple deletion and updating. Therefore, I am assuming I was getting this error. Once I resolved it, these testcases stopped failing.

PR #902, opened 2023-03-11
Configuring PYTHON_LEGACY UUID representation in cfc_webapp.py

swastis10:
We cannot encode native uuid.UUID with UuidRepresentation.UNSPECIFIED therefore defining uuid_representation as PYTHON_LEGACY fixes the issue 
Contains fix for - https://github.com/e-mission/e-mission-docs/issues/856#issuecomment-1464579523

### DIFF
--- a/emission/net/api/bottle.py
+++ b/emission/net/api/bottle.py
@@ -14,6 +14,8 @@ License: MIT (see LICENSE for details)
 """
 
 import sys
+import bson.json_util as bju
+from bson.binary import UuidRepresentation
 
 __author__ = 'Marcel Hellkamp'
 __version__ = '0.13-dev'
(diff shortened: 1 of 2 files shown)


swastis10:
The test data is now being loaded fine by the local server on the UI.
Screen Shot 2023-03-10 at 7 48 07 PM

shankari:
@swastis10 I don't think we should edit `bottle.py` directly since it is a dependency and not our code
I would use the approach outlined in https://github.com/bottlepy/bottle/issues/287

Note that I already do this in `cfc_webapp.py`, similar to the suggestions in the issue above

```
    for plugin in app.plugins:
        if isinstance(plugin, JSONPlugin):
            print("Replaced json_dumps in plugin with the one from bson")
            plugin.json_dumps = bson.json_util.dumps
```

You just need to change it to `plugin.json_dumps = lambda s: bju.dumps(...)`

swastis10:
The data is still loading fine, just like the screenshot attached above.

shankari (approved):
LGTM! Will merge and deploy after the tests finish running

PR #904, opened 2023-03-30
Purge timeseries (with csv export)

TTalex:
This PR is built atop PR https://github.com/e-mission/e-mission-server/pull/899

The idea is to offer a tool to purge entries from the timeseries database after they have been used by the pipeline process.
The goal is to reduce database footprint of active users, and improve overall pipeline performances.

This PR differs from PR https://github.com/e-mission/e-mission-server/pull/899 by:
- Exporting data before deletion to a configurable CSV file, to avoid losing precious information (default file being `/tmp/old_timeseries_<userUUID>.csv`)
- Adding a unit test file to confirm proper export and deletion using real data
- Allowing the method to be called outside the command line
- Adding the following command line parameters (and the corresponding method args) :

| Short args | Long arg | Description | Default value |
|---|---|---|---|
| -d | --dir_name | Target directory for exported csv data | "/tmp" |
|  | --file_prefix | File prefix for exported csv data | "old_timeseries_" |
|  | --unsafe_ignore_save | Ignore csv export of deleted data (not recommended, this operation is definitive) | False (unset) |

As a reminder, existing command args were:
| Short args | Long arg | Description | Default value |
|---|---|---|---|
| -h | | Help command |  |
| -e | --user_email | targeted user email (mutually exclusive with -u) | (required) |
| -u | --user_uuid | targeted user uuid (mutually exclusive with -e) | (required) |

**Some dev decision, subject to change:**
- Since the script targets specific user data, I opted to keep it dissociated from `bin/debug/purge_user.py`
- The script was kept inside the bin folder, in practice its use might fit in a cronjob, which I would not characterize as a "debug" tool
- The script only acts on the specific user, maybe an extension could be considered to retrieve all active users and run the method on each.
- I have placed the unit test file in a new `emission/tests/binTests` folder, since I couldn't find any tests matching this one.
- The default target directory is set to `/tmp`, which might break on non-unix machines
- File prefix and target directory params are probably a bit too sensitive to leading and ending slashes respectively, but this was deemed acceptable

Open to advices and suggestions.

### DIFF
--- a/bin/purge_user_timeseries.py
+++ b/bin/purge_user_timeseries.py
@@ -7,21 +7,24 @@ import emission.core.get_database as edb
 import emission.core.wrapper.pipelinestate as ecwp
 import emission.core.wrapper.pipelinestate as ecwp
 import emission.storage.pipeline_queries as esp
+import pandas as pd
 
-if __name__ == '__main__':
-    logging.basicConfig(level=logging.DEBUG)
 
-    parser = argparse.ArgumentParser(prog="purge_user_timeseries")
-    group = parser.add_mutually_exclusive_group(required=True)
-    group.add_argument("-e", "--user_email")
-    group.add_argument("-u", "--user_uuid")
+DEFAULT_DIR_NAME = "/tmp"
+DEFAULT_FILE_PREFIX = "old_timeseries_"
 
-    args = parser.parse_args()
+def exportOldTimeseriesAsCsv(user_id, last_ts_run, dir_name, file_prefix):
+    filename = dir_name + "/" + file_prefix + str(user_id) + ".csv"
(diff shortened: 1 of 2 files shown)


shankari:
@MukuFlash03 this is the next task for you to work on. You should:
- check to see if this works
- add tests for it
- discuss with cloud services where we can dump this data from an AWS perspective, (e.g. arctic storage), and then
- actually write the script and put it into our internal repo

I anticipate this will take ~ 3 weeks to complete 😄

MukuFlash03:
This is working, as it is. CSV file is storing the required data from timeseries db.
Currently it's stored as a temporary file but I'm keeping it on a permanent basis for now for testing purposes.
Will work on adding more tests and thinking of mongodump next.

PR #909, opened 2023-04-24
set end_confirmed_place on untracked composite objects

JGreenlee:
(Original PR was https://github.com/shankari/e-mission-server/pull/16)

Given https://github.com/e-mission/e-mission-docs/issues/887, the end place of untracked time is not actually zero duration. We do want to include it in the composite object. Thus, we can remove `isUntrackedTime` and the conditional check.


shankari (approved):
LGTM!

PR #917, opened 2023-05-20
✨ Include 'sections' in composite trips

JGreenlee:
This makes https://github.com/e-mission/e-mission-phone/pull/974 possible.

We retrieve a list of `cleaned_section` entries from the trip, and store this in the composite trip as `sections`.
We also filter out `speeds` and `distances` from the sections to save some bandwidth.


emission/analysis/plotting/composite_trip_creation.py
```diff
@@ -152,3 +153,15 @@ def get_locations_for_confirmed_trip(ct, max_entries=100):
         sample_rate = len(locations)//max_entries + 1
         locations = locations[::sample_rate]
     return locations
+
+def get_sections_for_confirmed_trip(ct):
+    if "cleaned_trip" not in ct["data"]:
+        return [] # untracked time has no sections
+    trip_tl = esdt.get_cleaned_timeline_for_trip(ct["user_id"], ct["data"]["cleaned_trip"])
```
shankari, line 160: I am not sure we always want to return `cleaned_section`s here. If we are only going to pick one, we should return the `inferred_section`s or ideally read it from the dynamic config.

In the old-style geojson object return, this was already a config option, but as part of the static config 
https://github.com/e-mission/e-mission-server/blob/cd0cdf9743aa7427e231fa95d887c5f1b156bf35/conf/analysis/debug.conf.json.sample#L10

JGreenlee:
@shankari I'm not having any luck retrieving inferred sections and will need some guidance.

Here's what I tried:

```python
def get_sections_for_confirmed_trip(ct):
    if "cleaned_trip" not in ct["data"]:
        return [] # untracked time has no sections
    section_key = "analysis/inferred_section" # TODO: this should be read from the dynamic config
    sections = esdt.get_sections_for_trip(key = section_key,
                                          user_id = ct["user_id"],
                                          trip_id = ct['data']['cleaned_trip'])
    # on the phone, we don't need lists of 'speeds' and 'distances'
    # for every section, and they can get big - so let's save some bandwidth
    for section in sections:
        del section["data"]["speeds"]
        del section["data"]["distances"]
        # convert enum to str - e.g. "MotionTypes.WALKING" instead of "1"
        section["data"]["sensed_mode"] = str(section["data"]["sensed_mode"])
    return sections
```
But `get_sections_for_trip()` returns no entries.

<hr>

In fact, querying for *any* `inferred_section`, I got zero hits.
```python
...
ts = esta.TimeSeries.get_time_series(user_id)
cleaned = ts.find_entries(["analysis/cleaned_section"], time_query=None)
print('For cleaned sections, ', [s['data']['sensed_mode'] for s in cleaned])
inferred = ts.find_entries(["analysis/inferred_section"], time_query=None)
print('For inferred sections, ', [s['data']['sensed_mode'] for s in inferred])
```
```
For cleaned sections,  [2, 0, 0, 2, 0, 4, 0, 0, 2, 0, 2, 0, 0, 0]
For inferred sections,  []
```

shankari:
Aha! All automated tests pass, although `emission/tests/analysisTests/userInputTests/TestUserInput.py` fails in manual testing. Let's see why that test isn't running, commit it so that it fails, and then fix it. We will need to fix the ground truth.

shankari:
To follow up on 
https://github.com/e-mission/e-mission-server/pull/917/commits/c516273e8784c04e86205956eca9b33494cefd62

> Why is that not in master? Need to double check.

Double-checked and it does seem to show up on master

```
======= Comparing trip =========
        "inferred_labels": [],
        "inferred_trip": {},
        "expectation": {},
        "confidence_threshold": 0.55,
        "expected_trip": {},

        "inferred_labels": [],
        "inferred_trip": { },
        "expectation": { },
        "confidence_threshold": 0.55,
        "expected_trip": { },
```

Not sure why we needed to add it here again?

shankari:
To follow up further on https://github.com/e-mission/e-mission-server/pull/917#issuecomment-1598013698, the changes were only on `expected_confirmed_trips`

shankari:
***Final set of changes***
- use inferred sections (with a fallback to cleaned sections), or use the config file value in composite trips
   - e.g. implement https://github.com/e-mission/e-mission-server/pull/917#discussion_r1208706830
- ensure that we have a basic random forest so that we generate some inferred values even on master
   - until we merge the two pipelines into one

shankari:
Final potential fix is to set up dummy random forest so that we return some inferred sections as well
We have a dummy random forest model at `emission/tests/data/seed_model_from_test_data.json`

The challenge is to avoid the situation where other deployers think that everything is working properly because there are no errors, and then complaining because the inferred modes are incorrect because we are using a simple/fake/mocked model.

I can think of two possible options:
- copy over the `emission/tests/data/seed_model_from_test_data.json` as part of the setup. This has the limitation above
- add an instruction to the README to copy over the seed model. This has the limitation that it won't work until people actually follow that instruction and not everybody reads the README carefully

@JGreenlee do you have a preference?

JGreenlee:
> @JGreenlee do you have a preference?

@shankari As you noted, it seems risky to include the dummy model by default because people won't realize it's fake. Also, existing partners who've been with the project probably don't want to re-read the README and it's a hassle to copy over an extra file.

I think ideally, you should be able to run the pipeline either with or without the dummy model, and you should be able to easily swap it out with your own.
So maybe instead of getting the model from a fixed filepath, run_intake_pipeline can take an optional argument that specifies the model that should be used (or none at all)

In the README, when we detail how to load data & run the pipeline, we can list both options and example usage with or without a seed model.
The 'default' workflow will work like it does now - no model and no inferences - and if people want to have dummy inferences they can pass the dummy model as an argument.

shankari:
@JGreenlee

> Also, existing partners who've been with the project probably don't want to re-read the README and it's a hassle to copy over an extra file.

This makes sense, except for the fact that no existing partners use the random forest model (the one on `master`). Everybody uses the GIS branch because it works better.
At some point, we will make the GIS branch the master (like we will do with `master_for_platform` and `master` on the phone).
We might choose to keep the random forest around as an ensemble, and should figure out how to integrate the two approaches at that time. And if we integrate, how we specify the model - I am not sure it should be a file in the first place on production.

Given all this, we just need the file right now to support inferred mode testing. I am going to just update the README.
Note that on the GIS branch, this will involve getting off the VPN if you get an SSL error.

PR #918, opened 2023-05-20
🛂 Support the new config for auto versus pre-generated tokens

shankari:
If the `opcode` field is present in the config, use it to determine whether the tokens should be auto-generated or pre-generated.

If it is not, fall back to the old study vs program check. We should remove the backwards compat code after all old-style configs have been removed

Testing done:
- Added new test with new style config
- Added new test for old style config that returns token list
- All tests pass

```
$ ./e-mission-py.bash emission/tests/netTests/TestWebServer.py
storage not configured, falling back to sample, default configuration
analysis.debug.conf.json not configured, falling back to sample, default configuration
ERROR:root:habitica not configured, game functions not supported
  File "/Users/kshankar/e-mission/gis_branch_tests/emission/net/ext_service/habitica/proxy.py", line 22, in <module>
    key_file = open('conf/net/ext_service/habitica.json')
FileNotFoundError: [Errno 2] No such file or directory: 'conf/net/ext_service/habitica.json'
Finished configuring logging for <RootLogger root (WARNING)>
Finished configuring logging for <RootLogger root (WARNING)>
.Finished configuring logging for <RootLogger root (WARNING)>
.Finished configuring logging for <RootLogger root (WARNING)>
.
Ran 3 tests in 0.735s

OK
```


shankari:
Related join webapp: https://github.com/e-mission/nrel-openpath-join-page/pull/13
Related phone changes: https://github.com/e-mission/e-mission-phone/pull/973

This should be the final piece for fixing https://github.com/e-mission/e-mission-docs/issues/850

PR #928, opened 2023-07-11
Libmamba solver Fixes e-mission/e-mission-docs#926

nataliejschultz:
Setting libmamba as the solver in the setup_conda.sh makes both local runs and runs in GH actions almost twice as fast, while fixing the issue of Docker running out of space to run conda.sh. 

While switching to mamba might be another option if different parts of conda start causing issues, this is a fix to the current issue that has the added benefit of making solving the environment much faster.[ See issue info here.](https://github.com/e-mission/e-mission-docs/issues/926)

### DIFF
--- a/setup/setup_conda.sh
+++ b/setup/setup_conda.sh
@@ -16,6 +16,8 @@ else
     bash miniconda.sh -b -p $INSTALL_PREFIX
     source $SOURCE_SCRIPT
     hash -r
+    conda install -n base conda-libmamba-solver
+    conda config --set solver libmamba
     conda config --set always_yes yes
     # Useful for debugging any issues with conda
     conda info -a


nataliejschultz:
Testing info: I have tested libmamba [both on GH actions](https://github.com/nataliejschultz/e-mission-server/actions/runs/5514145622/jobs/10053061268) and locally, by adding to the setup_conda.sh script:

```
hash -r
conda install -n base conda-libmamba-solver
conda config --set solver libmamba
conda config --set always_yes yes
```
Making libmamba the solver allowed the script to run successfully on GH actions, and was about twice as fast as previous runs. This increase in speed was also observed in my local tests. However, when running locally, I still had to increase docker resources up to 12GB.

shankari (approved):

PR #929, opened 2023-07-21
Creating TestNominatim.py

nataliejschultz:
Creating an integration test to make sure nominatim calls work properly. The tests call various classes from Nominatim.py and ensure that the format of the returned values are as expected with the current calling format.


emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,57 @@
...
+import unittest
+
+from emission.core.wrapper.trip_old import Coordinate
+
+import emission.net.ext_service.geocoder.nominatim as eco
+
+class NominatimTest(unittest.TestCase):
+
+    def test_make_url_geo(self):
+        expected_result = "http://nominatim.openstreetmap.org/search?q=Golden%2C+Colorado&format=json"
```
shankari, line 23: I don't think we should hardcode `nominatim.openstreetmap.org` for all these tests.
It is OK to hardcode them for a few tests (to test the fallback if the env variable is not present), but we should really be setting the appropriate environment variable and expect the value accordingly.

shankari (changes requested):
@nataliejschultz 

1. where are the cleanup changes to the module? Removing the google APIs, reading in the config as an environment variable....
2. this test is not actually being run right now - we should move it to `integrationTests` and run it the same way as the other integration tests, against the docker container for nominatim. Note that the exact strings returned here might also change as the OSM data changes
3. where is the code? There is no point in only checking in the tests without any code...

emission/integrationTests/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    networks:
+       - emission
+  #adding section to incorporate nominatim server functionality
+  nominatim:
+    image: mediagis/nominatim:4.2
+    container_name: norcal-nominatim
+    environment:
+      - PBF_URL=https://download.geofabrik.de/north-america/us/california/norcal-latest.osm.pbf
+      - REPLICATION_URL=https://download.geofabrik.de/north-america/us/california/norcal-updates/
+      - IMPORT_WIKIPEDIA=false
```
shankari, line 41: why are we still using norcal? I thought you wanted to switch to Rhode Island since norcal was too large

emission/integrationTests/nominatimTests/nominatim-docker-test.yml
shankari, line 1: all github actions need to be in the `.github` folder

emission/integrationTests/nominatimTests/nominatim-docker-test.yml
```diff
@@ -0,0 +1,34 @@
...
+# events but only for the master branch
+on:
+  push:
+    branches:
+        - master
+        - nominatim
+  pull_request:
+    branches:
+        - master
+
```
shankari, line 13: we don't want to run this on pull requests, only on the scheduled run

emission/integrationTests/nominatimTests/nominatim-docker-test.yml
```diff
@@ -0,0 +1,34 @@
+name: nominatim-docker-test
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches:
+        - master
+        - nominatim
```
shankari, line 9: we may be OK with running it on every push to master, but let's take it off for now to reduce our GitHub actions usage. Also, there is no `nominatim` branch in the main repo so even if we restored on push, we would not include the `nominatim` branch here

emission/integrationTests/nominatimTests/nominatim-docker-test.yml
```diff
@@ -0,0 +1,34 @@
...
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - name: Checkout
+    - uses: actions/checkout@v2
+
+    # Runs a single command using the runners shell
+    - name: Workflow test
+      run: echo Smoke test
+
+    - name: Initialize nominatim 
+      run: docker-compose -f docker-compose.yml up --exit-code-from web-server
```
shankari, line 34: this is wrong docker-compose file you need to integrationTests one

shankari (changes requested):
Apparently this does not include your most recent changes.
Can you update and I will re-review?

emission/integrationTests/nominatimTests/nominatim-docker-test.yml
shankari: to basic test the GitHub action, you should run the steps after the checkout job in the directory where you have checked out your code.

.github/workflows/nominatim-docker-test.yml
```diff
@@ -30,5 +26,5 @@ jobs:
     - name: Workflow test
       run: echo Smoke test
 
-    - name: Initialize nominatim 
-      run: docker-compose -f docker-compose.yml up --exit-code-from web-server
+    - name: Test nominatim 
+      run: docker-compose -f ../../emission/integrationTests/docker-compose.yml up --exit-code-from web-server
```
shankari, line 30: Not sure you need the `../..` at the beginning. This is running from the checked out directory

emission/integrationTests/docker-compose.yml
```diff
@@ -1,14 +1,13 @@
 version: "3"
 services:
   web-server:
-    build: ../../integrationTests
+    build: integrationTests
```
shankari, line 4: I don't think this works if you are already in `integrationTests`

emission/integrationTests/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# python 3
+FROM ubuntu:latest
+
+RUN apt-get update
+RUN apt-get install -y curl
+
+# CHANGEME: Create the files that correspond to your configuration in the conf directory
+# COPY conf/net/auth/google_auth.json /usr/src/app/conf/net/auth/google_auth.json
+COPY start_integration_tests.sh /start_integration_tests.sh
+
```
shankari, line 10: I don't see you copying the e-mission code in here!!

shankari (changes requested):
This is better, but I still think that there are errors with the relative paths.
Can you indicate how (if?) you tested this?

"and to put in outputs periodically when you have finished "one set of changes""

emission/integrationTests/docker-compose.yml
nataliejschultz: Trying out using "context" keyword to build in the current directory, rather than moving the files.

.github/workflows/nominatim-docker-test.yml
nataliejschultz: Removed :)

emission/individual_tests/TestNominatim.py
nataliejschultz: Went through and replaced with enviornment variable, which will be set to localhost URL during testing. Also currently going through and replacing all of the tests with locations in rhode island, so that they can call through the docker run server on GH.

shankari: You mean that you are changing all the real_data tests to rhode island?

That is not a good idea:
1. it is not real data any more
2. the location end points will no longer match up with the trajectories, unless
    3. you also move over the trajectories, which is an incredible amount of work

You should use fake trips instead. I have provided you with examples of tests that create fake trips.

nataliejschultz: No, I am not changing any of the real_data tests. I realize now how vague my comment was. In this file, I'm changing the unit tests to query locations/place names/latlon in rhode island, rather than those in California or Colorado. None of the Nominatim unit tests use a ground truth file.

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,54 @@
...
+standard_library.install_aliases()
+from builtins import *
+import unittest
+import os
+from emission.core.wrapper.trip_old import Coordinate
+import emission.net.ext_service.geocoder.nominatim as eco
+
+#temporarily sets NOMINATIM_QUERY_URL to the environment variable for testing.
+eco.NOMINATIM_QUERY_URL = os.environ.get("NOMINATIM_QUERY_URL")
+# print("query URL:", eco.NOMINATIM_QUERY_URL)
```
shankari, line 15: This reads the `NOMINATIM_QUERY_URL` from an environment variable in the test, but monkey patches it into the module.

emission/integrationTests/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV NOMINATIM_QUERY_URL='http://localhost:8080'
```
shankari, line 11: That will not work with this.

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -11,19 +11,12 @@
...
-    googlemaps_key_file = open("conf/net/ext_service/googlemaps.json")
-    GOOGLE_MAPS_KEY = json.load(googlemaps_key_file)["api_key"]
-    googlemaps_key_file.close()
-except:
-    print("google maps key not configured, falling back to nominatim")
 
 try:
     nominatim_file = open("conf/net/ext_service/nominatim.json")
     NOMINATIM_QUERY_URL = json.load(nominatim_file)["query_url"]
     nominatim_file.close()
```
shankari, line 18: you need to change this for it to work.

shankari (changes requested):
https://github.com/e-mission/e-mission-docs/issues/937#issuecomment-1679267601
is wrong. Please see comments below.

Again, I would encourage you to test as you go along so you know that the direction that you are taking is correct.

bin/debug/save_ground_truth.py
```diff
@@ -32,6 +32,7 @@ def save_ct_list(args):
 
     parser_diary = subparsers.add_parser('diary', help='diary-based ground truth')
     parser_diary.add_argument("date", help="date to retrieve ground truth (YYYY-MM-DD)")
+    parser_diary.add_argument("file_name", help="file name to store the result to")
```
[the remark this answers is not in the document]
nataliejschultz: Have been making much smaller, more frequent commits since this!

emission/individual_tests/TestNominatim.py
nataliejschultz: It seems to be a solution to my issue, since I can call and test functions in nominatim.py using the container data without modifying nominatim.json outside of the envirionment. Is there a different/preferred way to do this?

shankari: once you have changed the module to read from an environment variable (`os.environ.get`)
you can:
- monkey patch in here, OR
- set the environment variable and reload

So for example

```
os.environ.set("...", "http://localhost:...")
importlib.reload(eco)
```

or 

```
os.environ.set("...", "http://localhost:...")
import emission.net.... as eco
```

basically the module should read the environment variable when it is loaded and use it correctly

shankari: make sure to use fallbacks if the environment variable is not set, just like we use fallbacks if the config file is not present

nataliejschultz: This makes sense for testing. Later on, we will have a secret in secrets manager set the query url environment variable for production, and call in the server with boto3?

shankari: Conceptually. The cloud team will set the environment variable for the query URL on production using a secure method. I don' t know whether they use boto3 or not, but whatever method they use will be cyber-approved.

emission/integrationTests/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN apt-get install -y curl
 RUN echo "About to copy e-mission server code"
 COPY start_integration_tests.sh/ /start_integration_tests.sh
 
-ENV NOMINATIM_QUERY_URL='nominatim:8080'
```
shankari: is there a reason you put the `ENV` in the Dockerfile instead of the docker-compose?
Think about compile-time and run-time settings and how they map to image build and image run...

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+    else print("No key available")
+)
+
+#Creates a fake place in Rhode Island to use for testing.
+fake_id = "rhodeislander"
+key = "segmentation/raw_place"
+write_ts = 1694344333
+data = {'source': 'FakeTripGenerator','location': {'type': 'Point', 'coordinates': [-71.4128343, 41.8239891]}}
+fake_place = ecwe.Entry.create_fake_entry(fake_id, key, data, write_ts)
+
```
shankari, line 40: please check other test cases for proper test formatting. This should be in `setup`

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+class NominatimTest(unittest.TestCase):
+    maxDiff = None
+
+    #Basic query to check that nominatim and geofabrik are calling the same area.
```
shankari, line 44: the same area? you mean "return the same results"?

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+        nominatim_result = requests.get(NOMINATIM_QUERY_URL + "/reverse?lat=41.8239891&lon=-71.4128343&format=json").json()
+        geofabrik_result = requests.get(GEOFABRIK_QUERY_URL + "/reverse?lat=41.8239891&lon=-71.4128343&format=json").json()
+        key_list = ['osm_id', 'boundingbox']
+        for k in key_list:
+            self.assertEqual(nominatim_result[k], geofabrik_result[k])
+
+    #Checks the display name generated by get_filtered_place in clean_and_resample.py, which creates a cleaned place from the fake place
+    # and reverse geocodes with the coordinates.
+    def test_get_filtered_place(self):
+        raw_result = ecww.WrapperBase.__getattr__(clean.get_filtered_place(fake_place), "data")
```
shankari, line 55: I would suggest using `ecwe.Entry` instead. We don't use `WrapperBase` outside of the `ecw` package, IIRC.

emission/integrationTests/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# python 3
+FROM ubuntu:latest
+
+RUN apt-get update
+RUN apt-get install -y curl wget
```
shankari, line 5: nit: why do we need both `curl` and `wget`?

emission/integrationTests/start_integration_tests.sh
```diff
@@ -0,0 +1,37 @@
...
+
+
+echo "Setting up the test environment..."
+source setup/setup_tests.sh
+
+
+echo "Running tests..."
+source setup/activate_tests.sh
+
+echo "Adding permissions"
```
shankari, line 26:
```suggestion
echo "Adding permissions for the `runIntegrationTests.sh` script"
```

setup/docker-compose.tests.yml
```diff
@@ -32,6 +32,7 @@ services:
     networks:
        - emission
 
+
```
shankari, line 35: nit: extraneous whitespace

emission/tests/data/real_examples/shankari_2023-04-14.ground_truth
shankari, line 1: what is this file for? I don't see it used anywhere...

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+data = {'source': 'FakeTripGenerator','location': {'type': 'Point', 'coordinates': [-71.4128343, 41.8239891]}}
+fake_place = ecwe.Entry.create_fake_entry(fake_id, key, data, write_ts)
+
+class NominatimTest(unittest.TestCase):
+    maxDiff = None
+
+    #Basic query to check that nominatim and geofabrik are calling the same area.
+    def test_geofabrik_and_nominatim(self):
+        nominatim_result = requests.get(NOMINATIM_QUERY_URL + "/reverse?lat=41.8239891&lon=-71.4128343&format=json").json()
+        geofabrik_result = requests.get(GEOFABRIK_QUERY_URL + "/reverse?lat=41.8239891&lon=-71.4128343&format=json").json()
```
shankari, line 47: this is the only place where you use the `GEOFABRIK_QUERY_URL`. All other calls use nominatim OSM. This is not consistent with our prior plan: https://github.com/e-mission/e-mission-docs/issues/937#issuecomment-1716280421

Can you elaborate on the reasons for the change?

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+        expected_result_lat = Coordinate(41.8239891, -71.4128343).get_lat()
+        actual_result = eco.Geocoder.geocode("Providence, Rhode Island")
+        actual_result_lon = actual_result.get_lon()
+        actual_result_lat = actual_result.get_lat()
+        self.assertEqual(expected_result_lon, actual_result_lon)
+        self.assertEqual(expected_result_lat, actual_result_lat)
+
+    #Testing make_url_reverse, which creates a query url from a lat and lon
+    def test_make_url_reverse(self):
+        expected_result = NOMINATIM_QUERY_URL + "/reverse?lat=41.8239891&lon=-71.4128343&format=json"
```
shankari, line 88: why is this not using `make_url_geo`?

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_get_json_reverse(self):
+        expected_result = "Providence City Hall"
+        actual_result = eco.Geocoder.get_json_reverse(41.8239891, -71.4128343)["display_name"].split(",")[0]
+        self.assertEqual(expected_result, actual_result)
+
+    #Testing reverse_geocode, which reverse geocodes from a lat and lon and returns only the display name. 
+    def test_reverse_geocode(self):
+        expected_result = "Portugal Parkway, Fox Point, Providence, Providence County, 02906, United States"
+        actual_result = eco.Geocoder.reverse_geocode(41.8174476, -71.3903767)
+        self.assertEqual(expected_result, actual_result)
```
shankari, line 102: which of `get_json_reverse` and `reverse_geocode` is actually used in the current codebase? It seems like we should be able to remove the other. Doesn't hurt to have both, but simpler interfaces tend to be easier to maintain.

emission/integrationTests/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+      # - CHANGEME:/src/
+      - ../..:/src/e-mission-server
+    networks:
+       - emission
+       - nominatim
+
+  db:
+    image: mongo:4.4.0
+    ports:
+      - "1234:1234"
```
shankari, line 27: why are we exporting a port here?

emission/integrationTests/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+services:
+  web-server:
+    build: 
+      context: .
+    depends_on:
+      - db
+      - nominatim
+    environment:
+      - DB_HOST=db
+      - WEB_SERVER_HOST=0.0.0.0
```
shankari, line 11: do we need to export the webserver here?

emission/integrationTests/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+    image: nataliejschultz/rhodeisland-image:2.0
+    container_name: rhodeisland-nominatim
+    ports:
+      - "8080:8080"
+      # healthcheck currently commented out, since it isn't necessary and keeps going the entire time the container is running. 
+    healthcheck:
+      test: ["CMD", "curl", "-v", "http://localhost:8080"]
+      interval: 1m
+      timeout: 15s
+      retries: 3
```
shankari, line 51: it looks like the healthcheck is now commented in. Update comment?

emission/integrationTests/start_integration_tests.sh
```diff
@@ -0,0 +1,37 @@
...
+chmod +x runIntegrationTests.sh
+echo "Permissions added"
```
shankari, line 28: Ditto

setup/docker-compose.tests.yml
```diff
@@ -1,14 +1,14 @@
...
   web-server:
+  #builds from tests/dockerfile
     build: tests
     depends_on:
       - db
     environment:
       - DB_HOST=db
       - WEB_SERVER_HOST=0.0.0.0
-    ports:
-      - "8080:8080"
```
shankari, line 11: It's true that we don't need this, but it would be good to have a comment explaining why.
Ideally this would have been in the commit https://github.com/nataliejschultz/e-mission-server/commit/dbce159cfcb341f1ade5d0d5681f8cd3ac19d1d1 but we can at least put it in the PR now...

.github/workflows/nominatim-docker-test.yml
```diff
@@ -0,0 +1,35 @@
...
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  # pull_request:
+  #   branches:
+  #       - master
+  push:
+    branches:
+        - nominatim
```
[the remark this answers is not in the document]
nataliejschultz: Do you still want it to run on a schedule every week too? I had it this way for testing only!

.github/workflows/nominatim-docker-test.yml
```diff
@@ -0,0 +1,35 @@
...
+
+  # schedule:
+  #   # Run every Sunday at 4:05 am
+  #   - cron:  '5 4 * * 0'
```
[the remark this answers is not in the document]
nataliejschultz: Ignore above comment :)

conf/storage/db.conf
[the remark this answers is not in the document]
nataliejschultz: I will take this file out.

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+import os
+from emission.core.wrapper.trip_old import Coordinate
+import requests
+import emission.core.wrapper.entry as ecwe
+import emission.core.wrapper.wrapperbase as ecww
+import emission.net.ext_service.geocoder.nominatim as eco
+import emission.analysis.intake.cleaning.clean_and_resample as clean
+
+#temporarily sets NOMINATIM_QUERY_URL to the environment variable for testing.
+NOMINATIM_QUERY_URL_env = os.environ.get("NOMINATIM_QUERY_URL")
```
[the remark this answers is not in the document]
nataliejschultz: "Temporary" in the sense that in production it will be the geofabrik url, and in the test it will be the container name! So "temporarily" changing from the default for the purpose of the test.

emission/integrationTests/docker-compose.yml
nataliejschultz: I was running into issues with not specifying the ports. This was based on what I found on networking within GitHub actions. I'll try to run without it and see if it is necessary.

.github/workflows/nominatim-docker-test.yml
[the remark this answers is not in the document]
nataliejschultz: I had read somewhere that tests on a schedule can't be run on forked branches, so that's why it was commented out. I'll un-comment it once everything else is ready.

emission/individual_tests/TestNominatim.py
nataliejschultz: Done

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,113 @@
...
+
+NOMINATIM_QUERY_URL = (
```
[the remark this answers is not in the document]
nataliejschultz: Done

emission/individual_tests/TestNominatim.py
nataliejschultz: For the tests and geofabrik to work in production, we'll have to change the OPENSTREETMAP_QUERY_URL in nominatim.py to use the geofabrik URL with the API key. Currently, it passes in the url as an environment variable (which is the container URL during testing), but defaults to the nominiatim url if an env isn't set. We'll have to pass in the geofabrik url + API key as an environment variable with AWS secrets manager (or whatever method you had in mind) for production. Otherwise, the calls in the test are made to the container. I think I am confused about what you are suggesting!

emission/integrationTests/start_integration_tests.sh
nataliejschultz: Done

emission/integrationTests/start_integration_tests.sh
nataliejschultz: Done

emission/individual_tests/TestNominatim.py
nataliejschultz: Get_json_reverse is used in clean_and_resample's get_filtered_place function. 

reverse_geocode is used in emission/analysis/modelling/tour_model/tour_model_matrix.py and emission/incomplete_tests/TestProfile.py. 

However, reverse_geocode as it is currently in the code returns `_do_google_reverse(lat, lng)` (which I removed, along with all other google related functions), which used PyGeo.

emission/individual_tests/TestNominatim.py
nataliejschultz: I made a separate test for each method; makre_url_geo passes in an address, and make_url_reverse passes in lat/lon. Do you think this test is redundant+should be removed?

emission/individual_tests/TestNominatim.py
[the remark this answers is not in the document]
nataliejschultz: Changed :)

emission/tests/data/real_examples/shankari_2023-04-14.ground_truth
nataliejschultz: It was created when I was testing out save_ground_truth.py and making sure that the argparser line that I added worked :) removing!

emission/integrationTests/docker-compose.yml
nataliejschultz: Commented out now, not necessary with how fast the container is ready!

emission/integrationTests/Dockerfile
nataliejschultz: `wget` is used in the run command to donwload the .tar.gz file for dockerize. `curl` is used in setup_conda.sh: `curl -o miniconda.sh -L https://repo.continuum.io/miniconda/Miniconda3-py39_$EXP_CONDA_VER-1-$PLATFORM.sh;` 

I tested out replacing `curl` with `wget`, and it does not work

emission/individual_tests/TestNominatim.py
nataliejschultz: To give better context: reverse_geocode currently looks like this in production:
```
  @classmethod
    def reverse_geocode(cls, lat, lng):
        # try:
        #     jsn = cls.get_json_reverse(lat, lng)
        #     address = jsn["display_name"]
        #     return address

        # except:
        #     print "defaulting"
```
Which uses this function that I also removed:
```
def _do_google_reverse(lat, lng):
    geo = pyGeo(GOOGLE_MAPS_KEY)
    address = geo.reverse_geocode(lat, lng)
    return address[0]
```
And to replace it, I just un-commented the portion out that used nominatim, making it look like this:
```
    @classmethod
    def reverse_geocode(cls, lat, lng):
        jsn = cls.get_json_reverse(lat, lng)
        address = jsn["display_name"]
        return address
```

emission/integrationTests/docker-compose.yml
nataliejschultz: Nope! Test passed without it.

setup/docker-compose.tests.yml
nataliejschultz: This appears to be a very old version of the file, since I moved the compose to the integrationTests folder

setup/docker-compose.tests.yml
nataliejschultz: Let me know if this is true: The port is only necessary if the container is being called from somewhere else (ie when the server calls nominatim's container, it needs the port). Otherwise, the network function of GitHub actions takes care of keeping the containers separate.

emission/individual_tests/TestNominatim.py
nataliejschultz: I meant it as in the same location, since the exact results vary between geofabrik, nominatim, and the container

emission/integrationTests/docker-compose.yml
nataliejschultz: This port was not necessary for the test to succeed. However, removing the port for the container caused all of the tests that call the container to give ` urllib.error.HTTPError: HTTP Error 404: Not Found`

conf/storage/db.conf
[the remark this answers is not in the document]
nataliejschultz: Should I remove this section (below) from `start_integration_tests.sh`? Or set the variable value to the URL of the container? If so, I'll have to expose a port for the db service.

```
#set database URL using environment variable
echo ${DB_HOST}
if [ -z ${DB_HOST} ] ; then
    local_host=`hostname -i`
    sed "s_localhost_${local_host}_" conf/storage/db.conf.sample > conf/storage/db.conf
else
    sed "s_localhost_${DB_HOST}_" conf/storage/db.conf.sample > conf/storage/db.conf
fi
cat conf/storage/db.conf
```

db.conf looks like this on my end, but I haven't been pushing it as a change:

```
{
    "timeseries": {
        "url": "db",
        "result_limit": 250000
    }
}
```

.github/workflows/nominatim-docker-test.yml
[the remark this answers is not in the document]
nataliejschultz: I put in a section that will run on push to those paths that's currently commented out; will un-comment out once everything else is good to go (including the geofabrik key being in the main repo!)

emission/individual_tests/TestNominatim.py
nataliejschultz: @shankari Any comments or suggestions?

conf/storage/db.conf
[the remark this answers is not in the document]
shankari: you should not remove that section. Instead, you should set the `DB_HOST` variable in the docker-compose. That will configure `db. conf` properly **inside the container**.  similar to 
https://github.com/e-mission/e-mission-server/blob/bc78230d3196ca10249b6e95898f63fddf924f6f/setup/docker-compose.tests.yml#L8

Note that this is indeed an invalid configuration and is the reason for all the tests failing because they can't parse the storage configuration.

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -13,12 +13,12 @@
...
 from emission.core.wrapper.trip_old import Coordinate
 try:
-    NOMINATIM_QUERY_URL_env = os.environ.get("NOMINATIM_QUERY_URL", "")
-    logging.info(f"NOMINATIM_QUERY_URL_env: {NOMINATIM_QUERY_URL_env}")
-    NOMINATIM_QUERY_URL = NOMINATIM_QUERY_URL_env if NOMINATIM_QUERY_URL_env != "" else "http://nominatim.openstreetmap.org"
-
+    OPENSTREETMAP_QUERY_URL_env = os.environ.get("OPENSTREETMAP_QUERY_URL", "")
+    logging.info(f"OPENSTREETMAP_QUERY_URL_env: {OPENSTREETMAP_QUERY_URL_env}")
+    OPENSTREETMAP_QUERY_URL = OPENSTREETMAP_QUERY_URL_env if OPENSTREETMAP_QUERY_URL_env != "" else "http://nominatim.openstreetmap.org"
+    print("Open Street Map URL not configured, defaulting to nominatim:") if OPENSTREETMAP_QUERY_URL == "http://nominatim.openstreetmap.org" else print("Open Street Map URL configured!")
```
shankari, line 19: again, nominatim is software. "http://nominatim.openstreetmap.org" is the openstreetmap hosted version of nominatim. 

`Open Street Map URL not configured, defaulting to nominatim:") if OPENSTREETMAP_QUERY_URL == "http://nominatim.openstreetmap.org"` is wrong at so many levels...

shankari (changes requested):
"NOMINATIM", "OPENSTREETMAP" 
You Keep Using That Word, I Do Not Think It Means What You Think It Means

emission/individual_tests/TestNominatim.py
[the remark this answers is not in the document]
shankari: please see review:
https://github.com/e-mission/e-mission-server/pull/929#pullrequestreview-1651639356

emission/individual_tests/TestNominatim.py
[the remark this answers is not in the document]
shankari: please see review 
https://github.com/e-mission/e-mission-server/pull/929#pullrequestreview-1651639356

emission/individual_tests/TestNominatim.py
shankari: but the check is not whether they are calling the same area or not, it is whether they are returning the same results or not

emission/integrationTests/docker-compose.yml
shankari: that makes no sense. By "tests that call the container", which container do you mean? This is a 404 error, so it must be a container that is accessed via HTTP, so the nominatim container? But the nominatim container doesn't use mongodb - it can run completely independently. So I don't see how any such error could be related. Please clarify/add additional logs.

emission/individual_tests/TestNominatim.py
shankari: It seems like you can replace calls to `reverse_geocode` with `get_json_reverse.get("display_name")` and simplify the interface. But this is not critical for this PR - just a nice-to-have cleanup if you have time and motivation.

emission/individual_tests/TestNominatim.py
shankari: this has not yet been addressed

emission/individual_tests/TestNominatim.py
```diff
@@ -43,24 +40,24 @@ class NominatimTest(unittest.TestCase):
...
 
     #Checks the display name generated by get_filtered_place in clean_and_resample.py, which creates a cleaned place from the fake place
     # and reverse geocodes with the coordinates.
     def test_get_filtered_place(self):
-        raw_result = ecww.WrapperBase.__getattr__(clean.get_filtered_place(fake_place), "data")
-        print(NOMINATIM_QUERY_URL)
-        actual_result = ecww.WrapperBase.__getattr__(raw_result, "display_name")
+        raw_result = ecwe.Entry.__getattr__(clean.get_filtered_place(fake_place), "data")
+        print(OPENSTREETMAP_QUERY_URL)
+        actual_result = ecwe.Entry.__getattr__(raw_result, "display_name")
```
shankari, line 54: this is not the way to use `ecwe.Entry` please look at existing usage patterns and fix

setup/docker-compose.tests.yml
shankari: so revert the whitespace change?

emission/integrationTests/Dockerfile
shankari: Did you try replacing wget with curl? Also "did not work" doesn't really mean much - **how** did it not work?
This may be moot anyway given that you don't need dockerize, in which case you should remove `wget`

emission/individual_tests/TestNominatim.py
shankari: I don't remember what I meant by this comment any more, reverting..

emission/individual_tests/TestNominatim.py
shankari: this is not the way to use ecwe.Entry please look at existing usage patterns and fix

shankari (changes requested):

emission/integrationTests/docker-compose.yml
nataliejschultz: I removed the mongo container's port and the test worked just fine, since none of the other integration tests are currently run during my workflow. 

I also tried removing the port for the nominatim container. The rhodeisland-nominatim container is called by multiple tests within `TestNominatim.py`, each of which could not call the container without a port exposed. The point I was getting at is that the nominatim container still requires a port to be called by the tests in web-server.

General: you mentioned that we [want the mongo container present](https://github.com/e-mission/e-mission-docs/issues/937#issuecomment-1666231008), presumably for when all of the integration tests in the `integrationTests` directory are run together (assuming this is what you want). We definitely don't need the mongo port now, but once the [other tests are incorporated ](https://github.com/e-mission/e-mission-docs/issues/937#issuecomment-1686953729) the port might have to return to be called by web-server.

emission/individual_tests/TestNominatim.py
nataliejschultz: Assuming my workflow will eventually be used to run all of the integration tests, I've created a new shell script called `setup_integrationtests.sh` which is called in start_integration_tests.sh. The new shell script replaces setup_tests.sh's functionality, and runs createfakeplace.py. I could have just put createfakeplace.py into the `setup` directory and run it with import, but I think it is better to see during setup if if creating the fake place works. Also, creating the new script is more consistent with the general workflow I've seen in the server code. Finally, it opens the door to calling other scripts that might be necessary for other integration tests to work. 

These are my justifications, and either way it's better than creating the fake place in the test.

setup/docker-compose.tests.yml
nataliejschultz: Definitely thought this file was something else! Reverted the change :)

emission/integrationTests/Dockerfile
nataliejschultz: Dockerize removal committed

emission/individual_tests/TestNominatim.py
nataliejschultz: This is addressed in `bin/createfakeplace.py`

emission/individual_tests/TestNominatim.py
nataliejschultz: Okay, I think you mean something like this:
```
imports .....
def nominatim(nominatim_query_url):
    results = requests.get(nominatim_query_url)
    return results.json()

def test(self):
    lat = "41.8239891"
    lon = '-71.4128343'
    osm_query_url = f"https://nominatim.openstreetmap.org/reverse?lat={lat}&lon={lon}&format=json"
    geofabrik_query_url = f"https://geofabrik.de/...reverse?lat={lat}&lon={lon}&format=json"
    geofabrik_result = nominatim(geofabrik_query_url)
    osm_result = nominatim(osm_query_url)
    self.assertequal(geofabrik_result, osm_result)
```
Is that what you're suggesting?

shankari: yup. although I think that the URL Is not an argument - it is passed in as an environment variable.
So you will need to set the environment variable and then call `make_json_reverse` or equivalent on the module

conf/storage/db.conf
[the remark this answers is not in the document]
nataliejschultz: Removed the file. `DB_HOST=db` is set in my docker compose.

emission/individual_tests/TestNominatim.py
nataliejschultz: Changed this comment

emission/net/ext_service/geocoder/nominatim.py
nataliejschultz: This is what I have written in nominatim.py to set up the URL used in production right now:
```
 GEOFABRIK_QUERY_URL_env = os.environ.get("GEOFABRIK_QUERY_URL", "")
    logging.info(f"GEOFABRIK_QUERY_URL_env: {GEOFABRIK_QUERY_URL_env}")

    NOMINATIM_QUERY_URL = (
        GEOFABRIK_QUERY_URL_env if GEOFABRIK_QUERY_URL_env != "" else "http://nominatim.openstreetmap.org"
    )
    print("Geofabrik Query URL not configured, defaulting to OSM") if "nominatim" in NOMINATIM_QUERY_URL else print("Geofabrik Query URL configured!")
```
So the free version of nominatim from OSM is used if the geofabrik url isn't passed in as an environment variable.

shankari: This is too convoluted. Also people don't have to use GEOFABRIK - they can pick any other nominatim provider.

`nominatim..py` should only use `NOMINATIM_QUERY_URL`
- In production, the `docker-compose` should set `NOMINATIM_QUERY_URL` to the geofabrik API URL
- In testing, you should set `NOMINATIM_QUERY_URL` by setting the environment variable
https://stackoverflow.com/questions/5971312/how-to-set-environment-variables-in-python

shankari: @nataliejschultz an environment variable is a **variable**. You can get and set it just like any other variable and use it to pass in a parameter
https://github.com/e-mission/e-mission-server/pull/929#discussion_r1341910859

nataliejschultz: I am setting the variable at the top of nominatim.py: 
```
    NOMINATIM_QUERY_URL = os.environ.get("NOMINATIM_QUERY_URL")

    if NOMINATIM_QUERY_URL is None:
        raise Exception("Nominatim query url not configured")
```
I modified both of the functions (make_url_geo and make_url_reverse) in the nominatim.py module to use `NOMINATIM_QUERY_URL` as such:

```
@classmethod
def make_url_geo(cls, address):
    params = {
        "q" : address,
        "format" : "json"
    }
    query_url = NOMINATIM_QUERY_URL + "/search?"
    encoded_params = urllib.parse.urlencode(params)
    url = query_url + encoded_params
    logging.debug("For geocoding, using URL %s" % url)
    return url
```

As of now I am setting the `NOMINATIM_QUERY_URL ` environment variable to the geofabrik url in the docker compose (for testing):
```
environment:
      - DB_HOST=db
      - GFBK_KEY=$GFBK_KEY
      - NOMINATIM_QUERY_URL=https://geofabrik.de/$GFBK_KEY
```

But if we wanted to test the other urls (container/osm urls) with the same functions (ie `eco.make_url_geo` and `make_url_reverse`) we would need to monkey patch the `NOMINATIM_QUERY_URL` variable in the nominatim.py module to the other urls (container and osm). But this also would cause us to have to reset the `eco.NOMINATIM_QUERY_URL` back to the geofabrik url if we then want to compare another `eco` function with 2 different urls. 

I might be greatly overcomplicating this. Do we want the functions in the nominatim.py module to make the same calls with the other urls (container/osm)in the tests at all? Or do you want the tests that compare results between different nominatim services to formulate the call urls outside of nominatim.py, and **only** use the value of `NOMINATIM_QUERY_URL` in nominatim.py? ie this pseudocode:

```
    def test_nominatim_api(self):
        lat, lon = "41.832942092439694", "-71.41558148857203"
        osm_url = OSM_QUERY_URL + "lat=%&lon=%&format=json" % (lat, lon)
        container_url = CONTAINER_URL + "lat=%&lon=%&format=json" % (lat, lon)
        osm_result = requests.get(osm_url)
        container_result = requests.get(container_url)
        geofabrik_result = eco.Geocoder.reverse_geocode(lat, lon)
        self.assertEqual(results)
```
Where the other container URLs are set as variables at the top of TestNominatim.py

shankari: > I modified both of the functions (make_url_geo and make_url_reverse) in the nominatim.py module to use NOMINATIM_QUERY_URL as such:

You are setting the **python** variable by reading the **environment** variable. This is correct.

> As of now I am setting the NOMINATIM_QUERY_URL environment variable to the geofabrik url in the docker compose (for testing):

This is wrong. You should set two separate environment variables - for the `GEOFABRIK` and the `OSM` URLs.

> we would need to monkey patch the NOMINATIM_QUERY_URL variable in the nominatim.py module to the other urls (container and osm). 

not quite. You would set (no need to monkey patch) the `NOMINATIM_QUERY_URL` environment variable in `TestNominatim.py` and reload the module (`importlib.reload`) before making calls to `eco.Geocoder`
https://github.com/e-mission/e-mission-server/pull/929#discussion_r1346642913

> But this also would cause us to have to reset the eco.NOMINATIM_QUERY_URL back to the geofabrik url if we then want to compare another eco function with 2 different urls.

You would still need to set the `NOMINATIM_QUERY_URL` environment variable back to geofabrik, which is trivial once you are passing in the `GEOFABRIK` URL separately.

>  might be greatly overcomplicating this. 

Yes

> Do we want the functions in the nominatim.py module to make the same calls with the other urls (container/osm)in the tests at all?

No. Why would I want that?

> Or do you want the tests that compare results between different nominatim services to formulate the call urls outside of nominatim.py, and only use the value of NOMINATIM_QUERY_URL in nominatim.py? ie this pseudocode:

Largely yes. Formulate the URL in TestNominatim, set the environment variable, reload the module so that it reads the new environment variable, only use `NOMINATIM_QUERY_URL` in `nominatim.py`

nataliejschultz: > Largely yes. Formulate the URL in TestNominatim, set the environment variable, reload the module so that it reads the new environment variable, only use `NOMINATIM_QUERY_URL` in `nominatim.py`

See [this comment ](https://github.com/e-mission/e-mission-docs/issues/937#issuecomment-1754046911)on how I've addressed this!

setup/setup_integrationtests.sh
shankari, line 1: once you move `createfakeplace.py` into `emission/tests` and invoke it from the test

```
import emission.test.createfakeplace as etcfp
etcfp.create_fake_place(..)
```

then you can remove this file

emission/integrationTests/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# python 3
+FROM ubuntu:latest
+
+RUN apt-get update
+RUN apt-get install -y curl
+
+# CHANGEME: Create the files that correspond to your configuration in the conf directory
+
+RUN echo "About to copy e-mission server code"
+COPY start_integration_tests.sh/ /start_integration_tests.sh
```
shankari, line 10: and replace this with `setup/setup_tests.sh` instead

shankari (changes requested):
@nataliejschultz
Really almost there, just a few cleanup comments

emission/integrationTests/Dockerfile
nataliejschultz: Done

setup/setup_integrationtests.sh
nataliejschultz: Done

emission/net/ext_service/geocoder/nominatim.py
```diff
@@ -9,26 +9,22 @@
...
-
 try:
-    googlemaps_key_file = open("conf/net/ext_service/googlemaps.json")
-    GOOGLE_MAPS_KEY = json.load(googlemaps_key_file)["api_key"]
-    googlemaps_key_file.close()
-except:
-    print("google maps key not configured, falling back to nominatim")
+    NOMINATIM_QUERY_URL = os.environ.get("NOMINATIM_QUERY_URL")
+    # logging.info(f"NOMINATIM_QUERY_URL: {NOMINATIM_QUERY_URL}")
+    print("Nominatim Query URL Configured!")
```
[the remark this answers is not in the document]
nataliejschultz: Done

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,114 @@
...
+#Setting query URLs
+OPENSTREETMAP_QUERY_URL = os.environ.get("OPENSTREETMAP_QUERY_URL")
+GEOFABRIK_QUERY_URL = os.environ.get("GEOFABRIK_QUERY_URL")
+NOMINATIM_CONTAINER_URL = os.environ.get("NOMINATIM_CONTAINER_URL")
+
+class NominatimTest(unittest.TestCase):
+    maxDiff = None
+
+    def setUp(self) -> None:
+        return super().setUp()
```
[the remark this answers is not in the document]
nataliejschultz: done

bin/createfakeplace.py
```diff
@@ -0,0 +1,10 @@
+import emission.core.wrapper.entry as ecwe
+
+#Creates a fake, cleaned place in Rhode Island for use in TestNominatim.py:
+def create_fake_place():
```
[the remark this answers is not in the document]
nataliejschultz: Moved to setup

emission/integrationTests/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+      restart_policy:
+        condition: on-failure
+    volumes:
+      - nominatim-data:/var/lib/postgresql/14/main
+    networks:
+      - nominatim
+
+networks:
+  emission:
+  nominatim:
```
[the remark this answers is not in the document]
nataliejschultz: I tested and this worked! I had to revert some of the workflow stuff, but I'll change it back now that (I think) I've addressed everything :)

.github/workflows/nominatim-docker-test.yml
```diff
@@ -0,0 +1,36 @@
...
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  
+  push:
+    paths:
+      - 'net/ext_service/geocoder'
+      - 'analysis/intake/cleaning/clean_and_resample.py'
+
```
shankari, line 11: we want to test on pull requests for these paths as well

```suggestion
  push:
    paths:
      - 'net/ext_service/geocoder'
      - 'analysis/intake/cleaning/clean_and_resample.py'

  pull_request:
    paths:
      - 'net/ext_service/geocoder'
      - 'analysis/intake/cleaning/clean_and_resample.py'
```

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,120 @@
...
+    maxDiff = None
+
+    def setUp(self):
+        #Creates a fake, cleaned place in Rhode Island
+        fake_id = "place_in_rhodeisland"
+        key = "segmentation/raw_place"
+        write_ts = 1694344333
+        data = {'source': 'FakeTripGenerator','location': {'type': 'Point', 'coordinates': [-71.4128343, 41.8239891]}}
+        fake_place = ecwe.Entry.create_fake_entry(fake_id, key, data, write_ts)
+        return fake_place
```
shankari, line 32: The general expectation is that `setUp` will set values and not return them

```
self.fake_place = fake_place
```

emission/individual_tests/TestNominatim.py
```diff
@@ -0,0 +1,120 @@
...
+        geofabrik_result = eco.Geocoder.get_json_reverse(lat,lon)
+        key_list = ['osm_id', 'boundingbox']
+        for k in key_list:
+            self.assertEqual(osm_result[k], geofabrik_result[k])
+            self.assertEqual(container_result[k], geofabrik_result[k])
+
+    #Checks the display name generated by get_filtered_place in clean_and_resample.py, which creates a cleaned place from the fake place
+    # and reverse geocodes with the coordinates.
+    def test_get_filtered_place(self):
+        fake_place_raw = self.setUp()
```
shankari, line 63:
```suggestion
        fake_place_raw = self.fake_place
```

emission/integrationTests/docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+      - mongo-data:/data/db
+    networks:
+       - emission
+  #adding section to incorporate nominatim server functionality
+  nominatim:
+    entrypoint: /app/start.sh
+    image: nataliejschultz/rhodeisland-image:4.0
+    container_name: rhodeisland-nominatim
+    ports:
+      - "8080:8080"
```
shankari, line 45: nit: I don't think you need this either since you are only communicating with the container over the internal `emission` network

shankari (changes requested):
Really almost done.

emission/integrationTests/docker-compose.yml
nataliejschultz: I tried removing it when I had the two networks and it broke the test (commit 8a6d33946cacc4ad70c4718c8185b8ad11104f31)  . I'll try calling by just the container name now with the one network and see!

shankari:
The failures are likely because I haven't configured the geofabrik API.
Let me do that now.

shankari:
Re-running after setting the variable....

shankari:
I configured the GEOFABRIK_API secret and the tests are still failing
I don't see the printout for `NOMINATIM_QUERY_URL`

Screen Shot 2023-10-18 at 3 07 49 PM

There is also another failure due to 

```
AttributeError: 'Cleanedplace' instance has no attribute 'display_name'
```

shankari:
I'm also seeing at least one location where the nominatim URL is apparently not available.

```
web-server_1  | Nominatim URL not configured, place decoding must happen on the client
```

nataliejschultz:
> I configured the GEOFABRIK_API secret and the tests are still failing I don't see the printout for `NOMINATIM_QUERY_URL`
> 
I'll add a print for the query urls to see what's going on

nataliejschultz:
> I'm also seeing at least one location where the nominatim URL is apparently not available.
> 
> ```
> web-server_1  | Nominatim URL not configured, place decoding must happen on the client
> ```

Hmmm. Some prints are in order. Let's see what happens when I revert the last change and add a print

nataliejschultz:
It looks like it's not pulling the secret successfully. Did you put it in repository secrets, or environment secrets @shankari ? I have it in repository secrets on my fork.

shankari:
Port 8080 should not have anything to do with it.

> It looks like it's not pulling the secret successfully. Did you put it in repository secrets, or environment secrets @shankari ? I have it in repository secrets on my fork.

Repository secrets

shankari:
> Do you know if you input it as a string? I don't know offhand what the problem could be, since it was pulling just fine on my fork.

I can't see the existing secret - only override or delete it. I didn't do anything specific to mark it as a string - I just copied and pasted the API key. Was I supposed to put it in quotes or something?

Can you access one of other repo secrets, like the DOCKER_USER, for example?

shankari:
It is known to work on https://github.com/e-mission/e-mission-server/blob/master/.github/workflows/image_build_push.yml
You might want to use that as a template

nataliejschultz:
> Can you access one of other repo secrets, like the DOCKER_USER, for example?

I have tried many different ways, and essentially, not yet. I've replaced the GEOFABRIK secret with `${{ secrets.DOCKER_USER }}`, and made various modifications:

Requiring secret on workflow call:
```
on:
  workflow_call:
    secrets:
      DOCKER_USER:
        description: "Docker user known working secret test"
        required: true
```

Setting secret as an explicit environment variable:

```
env:
  GEOFABRIK_API: ${{ secrets.DOCKER_USER }}
```
and then printing that variable

```
    - name: Secret test
      run: echo "$GEOFABRIK_API"
```
which results in this: 

```
Run echo $GEOFABRIK_API
  echo $GEOFABRIK_API
  shell: /usr/bin/bash -e {0}
  env:
    GEOFABRIK_API: 
```

I'm thinking it has to do with me being the one pushing to the branch. Since I am not the repository owner, and I likely don't have permission to use repository secrets, it's blocking me from using them in my workflow. I am not sure, though.

shankari:
Bingo!
https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

Given that, and the fact that we use a fork and pull request model, we can never test this code on pull requests.
So let's remove the pull requests from the "trigger" for the workflow and let's add a CI badge for this regular run into the README

emission/integrationTests/docker-compose.yml
nataliejschultz: So after all of the secret pulling issues were figured out, I ran a[ control run](https://github.com/nataliejschultz/e-mission-server/actions/runs/6659690036) to make sure the test was still passing on my fork (it was). Then, I removed just the port 8080 + changed the URL of the container from 
`NOMINATIM_CONTAINER_URL=http://rhodeisland-nominatim:8080`
to
`NOMINATIM_CONTAINER_URL=http://rhodeisland-nominatim`
and the[ test failed](https://github.com/nataliejschultz/e-mission-server/actions/runs/6696569855) with error ` urllib.error.HTTPError: HTTP Error 404: Not Found` on test_geofabrik_and_nominatim, where the container is called.

I then [re-added the port](https://github.com/nataliejschultz/e-mission-server/actions/runs/6709760539), and the tests passed just fine.

Unless I am missing something, I think the port is necessary in this case.

shankari: I see what the issue is.
So if you specify `http://rhodeisland-nominatim`, then the default HTTP port is 80, so it is effectively trying to use 
`http://rhodeisland-nominatim:80` but there is nothing listening to port 80, so it fails.

You need to do `http://rhodeisland-nominatim:8080` regardless, but it should work even when port 8080 is not exported because of docker networking. Exporting port 8080 allows you to connect to the nominatim container from your laptop, which is not a requirement for us.

shankari:
@nataliejschultz looks like that worked?

emission/integrationTests/docker-compose.yml
nataliejschultz: You were right, setting the URL to include `:8080`, while removing the 
```
    ports:
      - "8080:8080"
```
from the nominatim service still worked. I will re-revert the schedule and _maybe_ it could be done.

nataliejschultz:
> @nataliejschultz looks like that worked? Can you confirm so that I can merge?

Updated the schedule to go back, so hopefully (if everything else is okay with you) it's done?!

shankari (approved):
Ok this finally looks good.
I am going to merge and see how well it runs.
We should also do something similar for overpass.de and the transit matching code but let's get the email messages done first.

PR #930, opened 2023-08-11
Update greedy_similarity_binning.py

humbleOldSage:
Added a new list type variable to store  labels at one place in order of trips processed.  Explained at https://github.com/e-mission/e-mission-eval-private-data/issues/35#issuecomment-1674168256

### DIFF
--- a/emission/analysis/modelling/trip_model/greedy_similarity_binning.py
+++ b/emission/analysis/modelling/trip_model/greedy_similarity_binning.py
@@ -121,6 +121,7 @@ class GreedySimilarityBinning(eamuu.TripModel):
         self.is_incremental = config['incremental_evaluation']
 
         self.bins: Dict[str, Dict] = {}
+        self.tripLabels=[]
         
 
     def fit(self, trips: List[ecwc.Confirmedtrip]):
@@ -184,9 +185,11 @@ class GreedySimilarityBinning(eamuu.TripModel):
                 logging.debug(f"adding trip to bin {bin_id} with features {trip_features}")
                 self.bins[bin_id]['feature_rows'].append(trip_features)
                 self.bins[bin_id]['labels'].append(trip_labels)
+                self.tripLabels.append(bin_id)
             else:
                 # create new bin
                 new_bin_id = str(len(self.bins))
+                self.tripLabels.append(new_bin_id)
                 new_bin_record = {
                     'feature_rows': [trip_features],
                     'labels': [trip_labels],


emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -121,6 +121,7 @@ class label to apply:
         self.is_incremental = config['incremental_evaluation']
 
         self.bins: Dict[str, Dict] = {}
+        self.tripLabels=[]
```
shankari, line 124: I don't see why we need a separate array here.
the bins already are indexed by the bin_id, and bin has the labels in it

```
"25": {
'feature_rows': ....
'labels':...
}
```

So I don't see what this is adding to the data structure

humbleOldSage: The clustering.py file builds on `data_loc` dataframe that requires

    Trip 1   -->    bin No. (say 2)
    Trip 2.  -->    bin No. (say 3) 
    Trip 3   -->   bin No.( say 2)
    Trip 4   -->   bin No.( say 1)
    Trip 5   -->   bin No.( say 1)
    .
    .
    .

However, the  way they are stored here  is  .

```
  "1": { 'feature_rows': [ [feature of Trip4], [feature of Trip5]]
            'labels': ...
  }
  
  "2": { 'feature_rows': [ [feature of Trip1], [feature of Trip3]]
            'labels': ...
  }
  
  "3": { 'feature_rows': [ [feature of Trip2]]
            'labels': ...
  }
```

We can extract the trip and its bin from the way they are already stored, but creating a separate array initially itself feels more efficient

shankari: But the problem with creating an additional data structure is that then we are changing production code to make analysis easier. We don't really need the trip -> bin mapping on production, and at some point, might want to include memoization as well. So I would prefer doing the additional work in `clustering.py` to create the dataframe from the output of the production model and not the other way around.

humbleOldSage: That makes sense. I'll remove the extra data structure and move all the  additional computations to clustering.py.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -23,6 +23,9 @@
 INVALID_QUERY = {'metadata.key': 'invalid'}
 
 class BuiltinTimeSeries(esta.TimeSeries):
+
+    entryList=[]
+
```
shankari, line 28: @humbleOldSage this is wrong. There is not only one set of entries in the database. Please read and understand the data model from chapter 5 of my thesis.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -261,6 +264,9 @@ def get_entry_at_ts(self, key, ts_key, ts):
         logging.debug("get_entry_at_ts result = %s" % retValue)
         return retValue
 
+    def getEntryList(self):
+        return self.entryList
+
```
shankari, line 269: again, this is wrong as well. you can't return **the** entry list because there is not just one.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -290,7 +296,11 @@ def to_data_df(key, entry_it, map_fn = None):
         if map_fn is None:
             map_fn = BuiltinTimeSeries._to_df_entry
         # Dataframe doesn't like to work off an iterator - it wants everything in memory
-        df = pd.DataFrame([map_fn(e) for e in entry_it])
+        
+        for e in entry_it:
+            BuiltinTimeSeries.entryList.append(map_fn(e)) 
+        df = pd.DataFrame(BuiltinTimeSeries.entryList)
+
```
shankari, line 303: Ditto.

shankari (changes requested):

emission/storage/timeseries/builtin_timeseries.py
shankari: there are existing methods to get an entry list. the trip model uses existing methods. You should not have to make changes to the `BuiltinTimeseries` for this change.

emission/storage/timeseries/builtin_timeseries.py
humbleOldSage: Figured this was not necessary

humbleOldSage:
NO changes needed here if, https://github.com/e-mission/e-mission-eval-private-data/pull/37 are approved.

shankari:
Closing this since all changes are in https://github.com/e-mission/e-mission-eval-private-data/pull/37

PR #933, opened 2023-08-20
Moving dependence from custom branch's tour_model to master's trip_model

humbleOldSage:
The following changes reduces `e-mission-server-eval-private-data`'s `TRB_label_assist` dependence on custom branch (hlu09's branch). 

The `clustering.py` file (from the commit https://github.com/e-mission/e-mission-eval-private-data/pull/37/commits/88988d3b16ade7e8841d80296120ff7d59127321 ) passes  the `config['clustering_way']` to `greedy_similarity_binning.py` as config parameter.


emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -119,6 +119,11 @@ class label to apply:
         self.sim_thresh = config['similarity_threshold_meters']
         self.apply_cutoff = config['apply_cutoff']
         self.is_incremental = config['incremental_evaluation']
+        if config.get('clustering_way') is None:
```
humbleOldSage, line 122: This is the case when other dependencies initiate `GreedySimilarityBinning`. We set the value to previous default ,i.e., 'origin-destination'.

emission/analysis/modelling/similarity/similarity_metric.py
```diff
@@ -27,15 +27,21 @@ def similarity(self, a: List[float], b: List[float]) -> List[float]:
...
         """compares the features, returning true if they are similar
         within some threshold
 
         :param a: features for a trip
         :param b: features for another trip
         :param thresh: threshold for similarity
+        :clusterinWay: clustering based on origin/destination/origin-destination-pair
         :return: true if the feature similarity is within some threshold
         """
         similarity_values = self.similarity(a, b)
```
humbleOldSage, line 40: `similarity_values` takes the form ['origin-origin' point distance, 'destination-destination' point distance] as returned from [`similarity`](https://github.com/e-mission/e-mission-server/blob/94e7478e627fa8c171323662f951c611c0993031/emission/analysis/modelling/similarity/od_similarity.py#L21) function.

emission/analysis/modelling/similarity/similarity_metric.py
```diff
@@ -27,15 +27,21 @@ def similarity(self, a: List[float], b: List[float]) -> List[float]:
...
 
-    def similar(self, a: List[float], b: List[float], thresh: float) -> bool:
+    def similar(self, a: List[float], b: List[float], thresh: float, clusteringWay :str = 'origin-destination') -> bool:
         """compares the features, returning true if they are similar
         within some threshold
 
         :param a: features for a trip
         :param b: features for another trip
         :param thresh: threshold for similarity
+        :clusterinWay: clustering based on origin/destination/origin-destination-pair
```
shankari, line 37: typo.
Not sure how this ever worked?!

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
shankari: Note that it might be easier to add explanations as code comments, like we used to in the old days.

emission/analysis/modelling/similarity/similarity_metric.py
shankari: Ditto wrt comments

emission/analysis/modelling/similarity/similarity_metric.py
```diff
@@ -27,15 +27,21 @@ def similarity(self, a: List[float], b: List[float]) -> List[float]:
...
         :return: true if the feature similarity is within some threshold
         """
         similarity_values = self.similarity(a, b)
-        is_similar = all(map(lambda sim: sim <= thresh, similarity_values))
+        if clusteringWay == 'origin':
+            is_similar = similarity_values[0] <= thresh
+        elif clusteringWay == 'destination':
+            is_similar = similarity_values[1] <= thresh
+        else:
+            is_similar = all(map(lambda sim: sim <= thresh, similarity_values))
```
shankari, line 46: so the map here is iterating over a two element array?

shankari (changes requested):
While this works, I am not sure it is the best generalization.
Note that the current implementation is extremely general - there is nothing really to indicate that this is related to trips, or that the features are related to origin/destination etc

We could theoretically pass in distances and it will still work.

With the current code structure, I think you are expected to deal with the `clusteringWay` more upstream. When you extract features from the trips to pass into the similar function, you would pass in only the origin or the destination or both (e.g. if `clustering_way == 'origin'`, `a` and `b` would be of length 1.

shankari:
<details>
<summary>The full set of dependencies in the main e-mission server for the "similar" function</summary>

```
$ grep -rl similar emission/ --exclude=*.pyc
emission//net/api/wsgiserver2.py
emission//core/wrapper/untrackedtime.py
emission//analysis/classification/inference/Classifier.ipynb
emission//analysis/modelling/tour_model_first_only_orig/prior_unused/cluster_pipeline.py
emission//analysis/modelling/tour_model_first_only_orig/data_preprocessing.py
emission//analysis/modelling/tour_model_first_only_orig/get_scores.py
emission//analysis/modelling/tour_model_first_only_orig/similarity.py
emission//analysis/modelling/tour_model_first_only_orig/load_predict.py
emission//analysis/modelling/tour_model_first_only_orig/second_round_of_clustering.py
emission//analysis/modelling/tour_model_first_only_orig/get_request_percentage.py
emission//analysis/modelling/tour_model_first_only_orig/label_processing.py
emission//analysis/modelling/tour_model_first_only_orig/trajectory_matching/Frechet.py
emission//analysis/modelling/tour_model_first_only_orig/cluster_pipeline.py
emission//analysis/modelling/tour_model_first_only_orig/evaluation_pipeline.py
emission//analysis/modelling/tour_model_first_only_orig/cluster_groundtruth.py
emission//analysis/modelling/similarity/similarity_metric.py
emission//analysis/modelling/similarity/od_similarity.py
emission//analysis/modelling/similarity/similarity_metric_type.py
emission//analysis/modelling/tour_model/prior_unused/cluster_pipeline.py
emission//analysis/modelling/tour_model/data_preprocessing.py
emission//analysis/modelling/tour_model/get_scores.py
emission//analysis/modelling/tour_model/similarity.py
emission//analysis/modelling/tour_model/load_predict.py
emission//analysis/modelling/tour_model/second_round_of_clustering.py
emission//analysis/modelling/tour_model/get_request_percentage.py
emission//analysis/modelling/tour_model/label_processing.py
emission//analysis/modelling/tour_model/trajectory_matching/Frechet.py
emission//analysis/modelling/tour_model/cluster_pipeline.py
emission//analysis/modelling/tour_model/evaluation_pipeline.py
emission//analysis/modelling/tour_model/cluster_groundtruth.py
emission//analysis/modelling/tour_model_first_only/data_preprocessing.py
emission//analysis/modelling/tour_model_first_only/load_predict.py
emission//analysis/modelling/tour_model_first_only/evaluation_pipeline.py
emission//analysis/modelling/user_model/user_utility_model.py
emission//analysis/modelling/trip_model/model_type.py
emission//analysis/modelling/trip_model/util.py
emission//analysis/modelling/trip_model/greedy_similarity_binning.py
emission//analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
emission//tests/analysisTests/resultTests/TestTimeGrouping.py
emission//tests/modellingTests/TestBackwardsCompat.py
emission//tests/modellingTests/TestRunGreedyIncrementalModel.py
emission//tests/modellingTests/TestRunGreedyModel.py
emission//tests/modellingTests/TestSimilarityMetric.py
emission//tests/modellingTests/TestSimilarityAux.py
emission//tests/modellingTests/TestGreedySimilarityBinning.py
emission//incomplete_tests/TestSimilarity.py
```

</details>

Filtering out all the tour model code, we find a lot of tests

```
$ grep -rl similar emission/ --exclude=*.pyc | grep -v tour_model
emission//net/api/wsgiserver2.py
emission//core/wrapper/untrackedtime.py
emission//analysis/classification/inference/Classifier.ipynb
emission//analysis/modelling/similarity/similarity_metric.py
emission//analysis/modelling/similarity/od_similarity.py
emission//analysis/modelling/similarity/similarity_metric_type.py
emission//analysis/modelling/user_model/user_utility_model.py
emission//analysis/modelling/trip_model/model_type.py
emission//analysis/modelling/trip_model/util.py
emission//analysis/modelling/trip_model/greedy_similarity_binning.py
emission//analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
emission//tests/analysisTests/resultTests/TestTimeGrouping.py
emission//tests/modellingTests/TestBackwardsCompat.py
emission//tests/modellingTests/TestRunGreedyIncrementalModel.py
emission//tests/modellingTests/TestRunGreedyModel.py
emission//tests/modellingTests/TestSimilarityMetric.py
emission//tests/modellingTests/TestSimilarityAux.py
emission//tests/modellingTests/TestGreedySimilarityBinning.py
emission//incomplete_tests/TestSimilarity.py
```

Filtering out all the tests, we find a handful of locations

```
$ grep -rl similar emission/ --exclude=*.pyc | grep -v tour_model | grep -v modellingTests
emission//net/api/wsgiserver2.py
emission//core/wrapper/untrackedtime.py
emission//analysis/classification/inference/Classifier.ipynb
emission//analysis/modelling/similarity/similarity_metric.py
emission//analysis/modelling/similarity/od_similarity.py
emission//analysis/modelling/similarity/similarity_metric_type.py
emission//analysis/modelling/user_model/user_utility_model.py
emission//analysis/modelling/trip_model/model_type.py
emission//analysis/modelling/trip_model/util.py
emission//analysis/modelling/trip_model/greedy_similarity_binning.py
emission//analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
emission//tests/analysisTests/resultTests/TestTimeGrouping.py
emission//incomplete_tests/TestSimilarity.py
```

humbleOldSage:
The grep statement above is matching 'similarity'  along with 'similar'.  

The results below are for similar only.

```
$ grep -rl -w similar Documents/GitHub/e-mission-server/emission/ --exclude=*.pyc | grep -v tour_model | grep -v modellingTests
Documents/GitHub/e-mission-server/emission//net/api/wsgiserver2.py
Documents/GitHub/e-mission-server/emission//core/wrapper/untrackedtime.py
Documents/GitHub/e-mission-server/emission//analysis/classification/inference/Classifier.ipynb
Documents/GitHub/e-mission-server/emission//analysis/modelling/similarity/similarity_metric.py
Documents/GitHub/e-mission-server/emission//analysis/modelling/user_model/user_utility_model.py
Documents/GitHub/e-mission-server/emission//analysis/modelling/trip_model/greedy_similarity_binning.py
Documents/GitHub/e-mission-server/emission//analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
Documents/GitHub/e-mission-server/emission//tests/analysisTests/resultTests/TestTimeGrouping.py
```

Here, except for 4 ( `similarity_metric.py`) and 6 ( `greedy_similarity_binning.py`), all others matches are in comments and so no need to change them.

4 has the definition of `similar` function and 6 has those calls for `similar` which we have been improving till now. There just one another call from 6 itself. So all in all, there are no other dependencies that we need to worry about while changing `similar`.

emission/analysis/modelling/similarity/similarity_metric.py
humbleOldSage: yes. removed it in next commit.

emission/analysis/modelling/similarity/od_similarity.py
```diff
@@ -16,6 +16,16 @@ def extract_features(self, trip: ecwc.Confirmedtrip) -> List[float]:
...
+        """
+        a : a list of point features that can take either of two forms
+                    1. [point1_latitude,point1_longitude]  
+                    2. [point1_latitude,point1_longitude,point2_latitude,point2_longitude] 
+                    
+        b : a list of point features that can take either of two forms
+                    1. [point1_latitude,point1_longitude]  
+                    2. [point1_latitude,point1_longitude,point2_latitude,point2_longitude] 
+        """
+        point_dist = [ecc.calDistance(a[i:i+2], b[i:i+2])
```
humbleOldSage, line 35: The origin/destination decision is moved upstream to `emission/analysis/modelling/trip_model/greedy_similarity_binning.py` in this commit. So we don't need separate  `calcDistance` calls.

emission/analysis/modelling/similarity/similarity_metric.py
humbleOldSage: These changes are now reverted.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -207,14 +207,22 @@ def _find_matching_bin_id(self, trip_features: List[float]) -> Optional[str]:
...
+        :param trip_features: feature row for the incoming trip. 
+                            takes the form [orig_lat, orig_lon, dest_lat, dest_lon]
         :return: the id of a bin if a match was found, otherwise None
         """
         for bin_id, bin_record in self.bins.items():
-                matches_bin = all([self.metric.similar(trip_features, bin_sample, self.sim_thresh,self.clusteringWay)
-                    for bin_sample in bin_record['feature_rows']])
-                if matches_bin:
-                    return bin_id
+            if self.clusteringWay == 'origin':
```
humbleOldSage, line 215: Moving `clusteringWay` parameter based decision making from `od_similarity`  here. This helps generalize similarity calculations. Earlier, complete features list was passed to `similar` in every scenario and  decisions taken there.

emission/analysis/modelling/similarity/similarity_metric.py
```diff
@@ -27,21 +27,17 @@ def similarity(self, a: List[float], b: List[float]) -> List[float]:
...
         :return: true if the feature similarity is within some threshold
         """
         similarity_values = self.similarity(a, b)
-        if clusteringWay == 'origin':
-            is_similar = similarity_values[0] <= thresh
-        elif clusteringWay == 'destination':
-            is_similar = similarity_values[1] <= thresh
-        else:
-            is_similar = all(map(lambda sim: sim <= thresh, similarity_values))
+        is_similar = all(sim <= thresh for sim in similarity_values)
```
humbleOldSage, line 40: Since `origin`/`destination`/`origin-destination` decision is moved upstream to `greedy_similarity_binning.py` we revert to initial version of this function. This change opens up `similar` function to more generalized usage.

emission/analysis/modelling/similarity/similarity_metric.py
humbleOldSage: These changes are now reverted.

humbleOldSage:
> Reminder: you also need to add new unit tests for the new functionality

I have already worked on a few of them. As soon as we finalize on the flow here, I'll commit them.

emission/analysis/modelling/similarity/similarity_metric.py
```diff
@@ -31,11 +31,12 @@ def similar(self, a: List[float], b: List[float], thresh: float) -> bool:
...
         within some threshold
 
-        :param a: features for a trip
+        :param a: features for a trip 
         :param b: features for another trip
         :param thresh: threshold for similarity
         :return: true if the feature similarity is within some threshold
         """
         similarity_values = self.similarity(a, b)
-        is_similar = all(map(lambda sim: sim <= thresh, similarity_values))
```
shankari, line 40: can we please revert these properly (using `git checkout`) to avoid messing up the blame?

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
shankari: As I said before:
> Note that it might be easier to add explanations as code comments, like we used to in the old days.

https://github.com/e-mission/e-mission-server/pull/933#discussion_r1300940088

> While this is better than nothing - it requires people to use git blame to find the commit, then find the related PR and scroll down to find these comments. And since we will be resolving them, they would all be hidden, so they would not be searchable anyway.

https://github.com/e-mission/e-mission-eval-private-data/pull/37#discussion_r1302993000

Please put long-term code comments into the code

emission/analysis/modelling/similarity/od_similarity.py
shankari: Yeah `calDistance` is a bit funky in that it returns two outputs. if we have more issues with this, we should consider simplifying further but I am fine with this for now.

shankari (changes requested):
Except for the minor changes, primarily around commenting, this looks fine.
Note that you also need to add unit tests for the new functionality.

humbleOldSage:
Still need to check the other tests dependent on `modellingTestAssets.py`'s `generate_mock_trips` function.

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -6,26 +6,93 @@ class TestSimilarityMetric(unittest.TestCase):
 
     def testODsAreSimilar(self):
         generate_points_thresh = 0.001  # approx. 111 meters
-        similarity_threshold = 500  # 
-        # random, but, points are sampled within a circle and should always be < sim threshold
-        trips = etmm.generate_mock_trips('bob', 2, [0, 0], [1, 1], threshold=generate_points_thresh)
+        similarity_threshold = 111  #
```
shankari, line 9: why did you have to change the threshold to 111? I can understand filtering for o and d, but why do you have to change the threshold?

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -6,26 +6,93 @@ class TestSimilarityMetric(unittest.TestCase):
...
+        metric = eamso.OriginDestinationSimilarity()
+
+        ## Sub-Test 1-2: 
+        # Two trips with neither origin nor destination coordinates within threshold
+        # must not be similar in any configuration of similarity testing.
+        trips = etmm.generate_mock_trips('bob', 2, [0, 0], [1, 1], '__', threshold=generate_points_thresh)  
+        coords0 = metric.extract_features(trips[0])
+        coords1 = metric.extract_features(trips[1])
+        similar11 = metric.similar(coords0[:2], coords1[:2], similarity_threshold)
+        similar12 = metric.similar(coords0[2:], coords1[:], similarity_threshold)
```
shankari, line 64: shouldn't this be `coords[:2]`?

humbleOldSage:
> Still need to check other tests dependent on `modellingTestAssets.py`'s `generate_mock_trips` function. Specifically, these files :

This was completed with the last commit.

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -6,26 +6,93 @@ class TestSimilarityMetric(unittest.TestCase):
...
         metric = eamso.OriginDestinationSimilarity()
-        coords0 = metric.extract_features(trips0[0])
-        coords1 = metric.extract_features(trips1[0])
-        similar = metric.similar(coords0, coords1, similarity_threshold)
-        self.assertFalse(similar)
+        coords0 = metric.extract_features(trips[0])
+        coords1 = metric.extract_features(trips[1])
+        similar21 = metric.similar(coords0[2:], coords1[2:], similarity_threshold)
+        similar22 = metric.similar(coords0, coords1, similarity_threshold)
+
```
shankari, line 76: I am generally confused here. Both of these use the same metric.
Where are you specifying that we should use destination or origin or both?

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -6,26 +6,93 @@ class TestSimilarityMetric(unittest.TestCase):
...
+        ## Sub-Test 5-6: 
+        # Two trips with destination coordinates within threshold but we check 
+        # similarity using origin coordinates or origin-and-destination 
+        # should not be similar.        
+        trips = etmm.generate_mock_trips('Caty', 2, [3, 3], [4, 4], '_d', threshold=generate_points_thresh)
+        metric = eamso.OriginDestinationSimilarity()
+        coords0 = metric.extract_features(trips[0])
+        coords1 = metric.extract_features(trips[1])
+        similar31 = metric.similar(coords0[:2], coords1[:2], similarity_threshold)
+        similar32 = metric.similar(coords0, coords1, similarity_threshold)
```
shankari, line 86: Ditto.

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -6,26 +6,93 @@ class TestSimilarityMetric(unittest.TestCase):
...
+        coords0 = metric.extract_features(trips[0])[2:]
+        coords1 = metric.extract_features(trips[1])[2:]        
+        similarD = metric.similar(coords0, coords1, similarity_threshold)
+
+        # All the similars must be true
+        self.assertTrue(similarOD1) # RESULT SUB-TEST 1
+        self.assertTrue(similarOD2) # RESULT SUB-TEST 2
+        self.assertTrue(similarOD3) # RESULT SUB-TEST 3
+        self.assertTrue(similarO)  # RESULT SUB-TEST 4
+        self.assertTrue(similarD) # RESULT SUB-TEST 5
```
shankari, line 50: I would put the assert right after the test so that it is clear what you are testing

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -24,31 +30,414 @@ def testBinning(self):
...
+        # have size 1.
+
+        at_least_one_large_bin = all(map(lambda b: len(b['feature_rows']) == 1, model1.bins.values()))
+        self.assertTrue(at_least_one_large_bin, "no bin should have more than 1 features in it")
+
+        at_least_one_large_bin = all(map(lambda b: len(b['feature_rows']) == 1, model2.bins.values()))
+        self.assertTrue(at_least_one_large_bin, "no bin should have more than 1 features in it")
+
+        at_least_one_large_bin = all(map(lambda b: len(b['feature_rows']) == 1, model3.bins.values()))
+        self.assertTrue(at_least_one_large_bin, "no bin should have more than 1 features in it")
```
shankari, line 118: nit: move the asserts just below the computation so that it is more clear what it is doing.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -24,31 +30,414 @@ def testBinning(self):
...
+        #                                             if 2 points lie within the mentioned 
+        #                                              similarity_threshold_meters
+        
+        model1_config = {
+            "metric": "od_similarity",
+            "similarity_threshold_meters": 111,  # meters,
+            "apply_cutoff": False,
+            "clustering_way": "origin",
+            "incremental_evaluation": False
+        }
```
shankari, line 188: It seems like you can create these in `setup` (that is what it is for) or at least pull them out into a separate function that you can reuse instead of copy-pasting.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -24,31 +30,414 @@ def testBinning(self):
...
+        #                   2. 'destination' -> using destination of the trip to check if 2 points
+        #                                       lie within the mentioned similarity_threshold_meters
+        #                   3. 'origin-destination' -> both origin and destination of the trip to check 
+        #                                             if 2 points lie within the mentioned 
+        #                                              similarity_threshold_meters
+        
+        model1_config = {
             "metric": "od_similarity",
-            "similarity_threshold_meters": 500,  # meters,
+            "similarity_threshold_meters": 111,  # meters,
```
shankari, line 77: why did you change this to `111`?

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -24,31 +30,414 @@ def testBinning(self):
...
+        model3.fit(trips)
+
+        # Since both the origin and the destination points of the generated trips lie 
+        # within the threshold, there should be binning in all three configs.
+
+        at_least_one_large_bin = any(map(lambda b: len(b['feature_rows']) == m, model1.bins.values()))
+        self.assertTrue(at_least_one_large_bin, "at least one bin should have at least 5 features in it")
+
+        at_least_one_large_bin = any(map(lambda b: len(b['feature_rows']) ==m, model2.bins.values()))
+        self.assertTrue(at_least_one_large_bin, "no bin should have more than 1 features in it")
```
shankari, line 435: the messages for the assert seem to be wrong given `len(b['feature_rows']) == m`

shankari (changes requested):
Almost there! You just need to clean up and polish the tests a bit.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: This was done to prevent pairing of points that were more than 111 meters away from each other, but less than 500 m. Now I realize that wont be possible, because the reference points of coordinate generation ( (0,0) or (1,1) ) are actually more than 100 kms away.

 However, after looking into this, keeping it to either (111 or 500) of the values MIGHT cause problem of wrong binning  because of the way these coordinates are generated in `modellingTestAssets.py` `generate_trip_coordinates` :
 

> ```radius = random.uniform(0, radius_threshold) if within_threshold else random.uniform(*radius_threshold, max) ```

This is why I feel this would create problem in certain cases :


 If we want  coordinates to be within `within_threshold` (set to 111m)  (points  in white like p1,p2,p3) , the radius is sampled to be less than `radius_threshold` , otherwise the radius is sampled in red marked area. However, say ,we generate 2 points (in blue) one within the `within_threshold` and the other one outside the `within_threshold` with the aim for no binning . However, our binning threshold d (500m) might result in binning them together if the outside point is sampled close enough to the circle.

 
MicrosoftTeams-image (2)

 
The way to fix this is when the points are not to be binned together, they need to be sampled around center points that are farther away than the binning threshold (500m). For eg, Below we take center points as (0,0) , (1,1) , (2,2). 
 
MicrosoftTeams-image (1)

 Fixing this in next commit.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: will do.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: yeah.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: True. Will fix.

emission/tests/modellingTests/TestSimilarityMetric.py
humbleOldSage: Same reason as https://github.com/e-mission/e-mission-server/pull/933/commits/710d1a5791212b540b883dedd5613a562071edc9#r1312355176 .

emission/tests/modellingTests/TestSimilarityMetric.py
humbleOldSage: If both the list have 4 values ( 2 for origin and 2 for destinaiton), the function will treat this as a  o-d similarity and check for similarity in both origin and destination.
if the list has 2 , its either O or D. From the POV of the function, it doesn't need to know if the 2 points are for O or from D , it just needs to tell if 2 points are close enough. The distinction between O and D is upstream, where we decide whether to send first 2 coords ( for origin) or last two (for destinaiton).

emission/tests/modellingTests/TestSimilarityMetric.py
humbleOldSage: I guess theres no variable by the name `coords`. However, it's still wrong there. It should be

```similar12 = metric.similar(coords0[2:], coords1[2:], similarity_threshold)``` . the 2 is missing in the second coords1 .

File: emission/tests/modellingTests/TestGreedySimilarityBinning.py (hunk 15 above)
Comment (humbleOldSage) [reply emission/tests/modellingTests/TestGreedySimilarityBinning.py:188]: I pull them out to a separate function. since there one value ( `clustering_way`) that keeps changing from test to test.

File: emission/tests/modellingTests/TestSimilarityMetric.py (hunk 10 above)
Comment (shankari) [reply emission/tests/modellingTests/TestSimilarityMetric.py:64]: correct. I don't claim that my reviews are ready to copy and paste 😄

File: emission/tests/modellingTests/TestSimilarityMetric.py (hunk 11 above)
Comment (shankari) [reply emission/tests/modellingTests/TestSimilarityMetric.py:76]: I understand what you are doing with the arrays.
However, you are not passing in anything to `similar` which indicates whether you should be using `o`, `d` or `od`.

I think you say that the determination of which metric to use is implicit in the array size. In general, I think that explicit is better than implicit here, although I am open to discussion around it.

Comment (humbleOldSage) [reply emission/tests/modellingTests/TestSimilarityMetric.py:76]: Yes. Its implicit. This similar  to what I am doing in Line 215 in ` greedy_similarity_binning.py` above ( around this [comment ]( https://github.com/e-mission/e-mission-server/pull/933/files#r1303879436)) where I decide the start and end point and then call similarity using if-else. 

Even if I do pass any of the three  (`origin`,`destination`,`origin-destination`) to `similar`,  the way `similar` is written in `od_similarity.py`  ( around this [comment]( https://github.com/e-mission/e-mission-server/pull/933/files#r1303869440)), It'll either :

a. not be of any  use there, OR
b. An if-else will have to be written, which is not required since we discussed previously  moved that decision making upstream to `greedy_similarity_binning.py`.

Alternately, I can add a comment here to explain this implicitness.

File: emission/tests/modellingTests/TestSimilarityMetric.py (hunk 10 above)
Comment (humbleOldSage) [reply emission/tests/modellingTests/TestSimilarityMetric.py]: Yeah. I did. I was working on its last bit during that time.

File: emission/tests/modellingTests/TestGreedySimilarityBinning.py (hunk 16 above)
Comment (shankari) [reply emission/tests/modellingTests/TestGreedySimilarityBinning.py]: As long as the thresholds are consistent, I don't see any issues with 500 (and the test has been running, and passing)  on production for a while with 500.

Essentially, we create "fake" trips. if the ends of the trips are not within the expected bin, we discard the trips. So I don't understand how we would get trips that are outside the bin radius.

Comment (shankari) [reply emission/tests/modellingTests/TestGreedySimilarityBinning.py]: For the record, the related issue was:
https://github.com/e-mission/e-mission-docs/issues/709
and the related PR for the first set of changes was:
https://github.com/e-mission/e-mission-server/pull/852

Comment (shankari) [reply emission/tests/modellingTests/TestGreedySimilarityBinning.py]: Neither had any discussion around the choice of 500.
Having said that, even if we are not discarding "invalid" trips right now, it seems like we can discard invalid trips going forward, since we don't really care about what the trips are, and they are mock trips anyway. We just want to have "n" trips that match some characteristic.

By "invalid", I mean, not matching the kinds of mock trips that we want.
So if we want 5 trips, 3 of which are binned in both o-d and 2 of which are binned only in o,
we can oversample (overcreate mock trips) and discard the ones that don't match these criteria until the criteria are met.

File: emission/tests/modellingTests/TestSimilarityMetric.py [hunk 18]
```diff
@@ -1,31 +1,61 @@
...
+        parameters= [(0,2),(2,4),[0,4]]
+        n=2
+        #this generates 2 trips one-by-one, where each trip's respective origin and destination 
+        # points are more than 500m away.
+        trips = [etmu.setTripConfig(1, (i, i), (i+1, i+1), 'od', 1)[0] for i in range(n)]    
+        trip0_coord = metric.extract_features(trips[0])
+        trip1_coord = metric.extract_features(trips[1])
+
+        for (coord_start,coord_end) in parameters:
+            with self.subTest(coordinates=(coord_start,coord_end)):
```
shankari, line 54: where is this defined?

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -95,33 +104,17 @@ def testNoPrediction(self):
...
-
-        model_config = {
-            "metric": "od_similarity",
-            "similarity_threshold_meters": 500,      # meters,
-            "apply_cutoff": False,
-            "incremental_evaluation": False
-        }
-        model = eamtg.GreedySimilarityBinning(model_config)
-
+        model = etmu.setModelConfig("od_similarity",  500,  False, "origin-destination", False)
```
shankari, line 117: I am not sure that this is actually more clean that just creating the model config. I was hoping that you would refactor the code around creating multiple configs, one for o, one for d, and one for od

shankari (changes requested):
I don't understand how these tests even run given that `self.subTest` is not defined.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -10,44 +10,65 @@ def setUp(self) -> None:
...
-        }
 
-        # generate $n trips. $m of them should have origin and destinations sampled
+        # generate $n trips.
+        n = 20     
+
+        #this generates 20 trips one-by-one, where each trip's respective origin and destination 
+        # points are more than 500m away.
+        trips = [ etmu.setTripConfig(1, (i, i), (i+1, i+1), 'od', 1)[0] for i in range(n)]    
+
```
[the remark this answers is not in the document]
humbleOldSage: This commit is from before that particular discussion. Its still to come

emission/tests/modellingTests/utilities.py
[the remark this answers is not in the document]
humbleOldSage: yeah .we can. I'll move them there.

shankari:
> I don't understand how these tests even run given that self.subTest is not defined.

It looks like the modelling tests are not actually executed as part of the CI/CD.
If I search for `modellingTest`, I don't find any matches.
But if i search for "is not" I do
If I search for "modelling", I do

```
/home/runner/work/e-mission-server/e-mission-server/emission/analysis/modelling/tour_model_first_only/evaluation_pipeline.py:81: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if len(bins) is not 0:
/home/runner/work/e-mission-server/e-mission-server/emission/analysis/modelling/tour_model_first_only/evaluation_pipeline.py:110: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if len(bins) is not 0:
```

@humbleOldSage have you tried running these tests? have you seen if they are run by `runAllTests`?

emission/tests/modellingTests/TestSimilarityMetric.py
humbleOldSage: the `subtest` context manager comes with the python unit test module. I used it here to parameterize tests.

humbleOldSage:
> I don't understand how these tests even run given that `self.subTest` is not defined.

the subtest context manager comes with the python unit test module. No need to define it I guess.

> Can you please indicate "testing done"?

Testing done, but for the  method before centroid way was discussed. Once that's done, I'll indicate the same.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: Here (in functions `testNoPrediction` and `testPrediction`), I replaced this  just to make the trip setups and model calls uniform throughout this file ,i.e., everytime a trip is setup, it uses `setTripConfig` and every time binning is called using `setModelConfig`.  

The multiple configs are not required in functions `testNoPrediction` and `testPrediction`. 
And, where multiple configs are need  in `testNoBinning` and `testBinning` , this is handled through parametrized subTests.

humbleOldSage:
> @humbleOldSage have you tried running these tests? have you seen if they are run by `runAllTests`?

yes. But I ran each of them individually. 

I figured that these tests are invisible to the  command :

    PYTHONPATH=. python -m unittest discover -v emission/tests/modellingTests/
    
because of the names of the file, which is,  `Test*`.py .
To be discoverable, they need to take the form `test_*.py` and then the above command would run.

shankari:
> because of the names of the file, which is, Test*.py .
> To be discoverable, they need to take the form test_*.py and then the above command would run.

I am not sure why you are focused on `python -m unittest discover -v emission/tests/modellingTests/` - that is not what we run in `runAllTests.sh`

Also, the files in (say) `emission/tests/analysisTests/intakeTests` all start with `Test` and are run successfully

humbleOldSage:
Got it .

So, modifying `runAllTests.sh` file   from its initial command :
      
      PYTHONPATH=. python -m unittest discover -s emission/tests -p Test*;

to 

     PYTHONPATH=. python -m unittest discover -s emission/tests/modellingTests -p Test*;

runs all 21 tests without failure on the local machine.( I changed this so that we don't have to run all other test and focus just on modelling ones ).

shankari:
> runs all 21 tests without failure on the local machine.( I changed this so that we don't have to run all other test and focus just on modelling ones ).

That's good, and you should list that in the "testing done". However, we should also make sure that the tests run as part of `runAllTests.sh` so that they are run as part of the CI. If you run `runAllTests.sh` - does it run them or not?

humbleOldSage:
When I run the entire file, just when it begins, this is what it shows me 

```
ERROR:root:habitica not configured, game functions not supported
  File "/Users/ssaini/Documents/GitHub/e-mission-server/emission/net/ext_service/habitica/proxy.py", line 22, in <module>
    key_file = open('conf/net/ext_service/habitica.json')
FileNotFoundError: [Errno 2] No such file or directory: 'conf/net/ext_service/habitica.json'
Finished configuring logging for <RootLogger root (WARNING)>
WARNING:root:No user defined overrides for key config/sensor_config and user 4f1fc8f0-4d7d-4d3b-8081-0e4b3ffc9b67, early return
WARNING:root:No user defined overrides for key config/sync_config and user 4f1fc8f0-4d7d-4d3b-8081-0e4b3ffc9b67, early return
WARNING:root:No user defined overrides for key config/consent and user 4f1fc8f0-4d7d-4d3b-8081-0e4b3ffc9b67, early return
.WARNING:root:No user defined overrides for key config/sensor_config and user 2ee2248e-9520-40a6-a54e-7af2d8aa97f3, early return
WARNING:root:No user defined overrides for key config/sync_config and user 2ee2248e-9520-40a6-a54e-7af2d8aa97f3, early return
WARNING:root:No user defined overrides for key config/consent and user 2ee2248e-9520-40a6-a54e-7af2d8aa97f3, early return
.WARNING:root:No user defined overrides for key config/sync_config and user b14e823f-1a9c-460d-815a-2aefd21dd71f, early return
WARNING:root:No user defined overrides for key config/consent and user b14e823f-1a9c-460d-815a-2aefd21dd71f, early return
.WARNING:root:No user defined overrides for key config/sync_config and user a435e378-c92a-4653-8540-b51f44b635ae, early return
WARNING:root:No user defined overrides for key config/consent and user a435e378-c92a-4653-8540-b51f44b635ae, early return
.Setting up real example for 97779ee3-004d-4ffa-b7d3-a2db3ab112d7
```

and then it gets stuck there . NO output. The maximum I waited for this was 30 mins. Will leave this overnight to see if it progresses.

shankari:
Do you still have the full dataset loaded? That is going to slow down the tests. I would suggest shutting down this DB (assuming you have the data stored on a persistent volume) and starting a new blank one for better performance.

humbleOldSage:
Ok.  I'll do this. 
 Yeah, it's persistent now.

shankari:
Ah the modellingTests are enabled! But alas, they are failing.
Did you test before sending the change for review?

humbleOldSage:
I did, on just one Testfile. I felt the the logic works. I wanted to run the logic though you first and then handle the dependencies.

shankari:
If this is the case, **please indicate testing done** and that the tests are expected to fail

humbleOldSage:
yeah .I did move the PR to draft for this reason, but I'll do this as well moving forward.

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -1,35 +1,145 @@
...
 
 import emission.core.wrapper.entry as ecwe
 import time 
 import math
 
+def setModelConfig(metric,threshold,cutoff,clustering_way,incrementalevaluation):
+    """
+    TODO: Write about each parameter to the function
+    pass in a test configuration to the binning algorithm.
+
```
shankari, line 16: I am not sure that it is worth having a wrapper (`setModelConfig`) for just two lines of code. It doesn't seem to indicate an actual improvement over just having `model_config = {...}` with essentially the same parameters multiple times.

My comment around pulling out the code https://github.com/e-mission/e-mission-server/pull/933#discussion_r1312355017 was to avoid having the same model configs created over and over again. Each of your tests created (and creates) `model1`, `model2` and `model3` with the same configurations. you should put them into `setup` or create a wrapper function that creates all the configs.

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -1,35 +1,145 @@
...
+        "incremental_evaluation": incrementalevaluation
+    }
+
+    return eamtg.GreedySimilarityBinning(model_config)
+
+def generate_random_point():
+    """Generate a completetly random point valid WGS84 latitiude and longtidude"""
+    lat=random.uniform(-90,90)
+    lon=random.uniform(-180,180)
+    return [lat,lon]
```
shankari, line 42: why do we want to generate a completely random point? We could still specify a centroid like we did before, right?

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -1,35 +1,145 @@
...
+
+def calDistanceTest(point1, point2, coordinates=False):
+    """haversine distance
+
+    :param point1: a coordinate in degrees WGS84
+    :param point2: another coordinate in degrees WGS84
+    :param coordinates: if false, expect a list of coordinates, defaults to False
+    :return: distance approximately in meters
+    """
+    earthRadius = 6371000  # meters
```
shankari, line 63: Let us please not create Yet Another Copy of this. Is there a reason we can't use either of the two existing implementations?

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -1,35 +1,145 @@
...
+        lat2 = math.radians(point2[1])
+
+
+    a = (math.sin(dLat/2) ** 2) + ((math.sin(dLon/2) ** 2) * math.cos(lat1) * math.cos(lat2))
+    c = 2 * math.atan2(math.sqrt(a), math.sqrt(1-a))
+    d = earthRadius * c
+
+    return d
+
+def setTripConfig(trips,trip_part,threshold,within_thr=None,label_data=None):
```
shankari, line 82: this is super confusing because the name of the function is `setTripConfig`, but it has an argument called trips and it calls a function called `generate_mock_trip*s*` but then returns a variable called `trip`.
Also I am not sure why we need to wrap two lines of code in a separate function

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -1,35 +1,145 @@
...
+def generate_nearby_random_points(ref_coords,threshold):
+    """
+    Generate valid WGS84 latitiude and longtidude in threshold(m) proximity to
+    ref coordinates
+    """
+
+    thresholdInWGS84 = threshold* (0.000001/0.11)
+    dx=random.uniform(-thresholdInWGS84/2,thresholdInWGS84/2)
+    dy=random.uniform(-thresholdInWGS84/2,thresholdInWGS84/2)
+    return [ref_coords[0] +dx , ref_coords[1] +dy]
```
shankari, line 53: is this guaranteed to generate points within the threshold?
Can you explain how in the code comments?

shankari (changes requested):
See only one conceptual clarification.
Others are software engineering fixes and requests for comments.

emission/tests/modellingTests/TestSimilarityMetric.py
shankari: `GreedySimilaritityBinning` now accepts the `clusteringWay` parameter
and then instead of passing it through to `similar`, it passes in the subset of features instead.

it makes sense that we don't want to have the o, d, od distinction in `similarity_metric` since the metric only cares about whether elements are similar or not.

however, it seems like in `od_similarity`, we are implicitly doing an `o`, `d`, `od` check just in a fairly convoluted fashion by expecting a certain type of input. In general, (this is more art than science), explicit is better than implicit, especially when the inputs are standard values like locations. So I think that `od_similarity` should continue to take full locations but should also take in an explicit `clusteringWay` and do the manipulation into the feature selection in there. If it were not called `od_similarity` maybe I would feel differently but it is and so we should handle the `od` distinction there.

humbleOldSage: We had a discussion regarding this here :

https://github.com/e-mission/e-mission-server/pull/933#pullrequestreview-1588325028

where we decided to move the `clustering_way` based decision making upstream .

shankari: That discussion was focused on making sure that we did not have the `if-else` in the `similar` function instead the similarity metric. It did not discuss the `od_similarity` module and the `similarity` interface. After looking through the code again, and looking at the interface exposed by the unit tests, you should have the `clusteringWay` passed in to `similarity`, do the if/then inside `similarity` and invoke `similarity` directly from the unit tests.

That will dramatically simplify your unit test interface and make everything much cleaner.

humbleOldSage:
Final Test output  from runAllTest.sh:

Screen Shot 2023-09-12 at 11 33 47 AM

emission/tests/modellingTests/modellingTestAssets.py
humbleOldSage: Commented in  commit f5944ccef1140db347211c93b851b8864c0362ee

emission/tests/modellingTests/modellingTestAssets.py
humbleOldSage: removed in next commit  f5944ccef1140db347211c93b851b8864c0362ee

emission/tests/modellingTests/modellingTestAssets.py
humbleOldSage: From the next commit f5944ccef1140db347211c93b851b8864c0362ee , this will be used  when 

1.no ref. points are passed.
2. when we have to generate points that need to be outside the threshold.

 Currently, I do pass (0,0) and (1,1) as ref points.

emission/tests/modellingTests/modellingTestAssets.py
humbleOldSage: Removed in  commit f5944ccef1140db347211c93b851b8864c0362ee

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -10,44 +11,111 @@ def setUp(self) -> None:
...
-        model.fit(trips)
 
-        # $m trip features should appear together in one bin
-        at_least_one_large_bin = any(map(lambda b: len(b['feature_rows']) == m, model.bins.values()))
-        self.assertTrue(at_least_one_large_bin, "at least one bin should have at least 5 features in it")
+        # parameters passed for testing. A list, where each element of this list takes the form 
+        # [trip part to be sampled within mentioned threshold , clustering way used to check similarity]
+        parameters= [["o_",'origin'],["_d",'destination'],["od",'origin-destination']]
+        for tp,cw in parameters:
+            with self.subTest(trip_part=tp,clustering_way=cw):
```
shankari, line 93: I think that subTest largely includes what I had in mind for avoiding the duplication of `model_config` creation code.

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -1,35 +1,72 @@
...
              circle from some coordinates up to some threshold
     """
-    angle = 2 * math.pi * random.random()
-    radius_threshold = threshold / 2
-    radius = random.uniform(0, radius_threshold) if within_threshold else random.uniform(radius_threshold, max)
-    x = radius * math.cos(angle) + ref_coords[0]
-    y = radius * math.sin(angle) + ref_coords[1]
-    return (x, y)
+    # if the point is to be generated within a threshold and it's not the first point
+    if InsideThreshold and points_list:
```
shankari, line 53: nit: by most naming conventions, variables in python start with lowercase, classes start with upper case

emission/tests/modellingTests/modellingTestAssets.py
shankari: The comment is not quite clear. Let's discuss.
Similarly, I don't know how generating a random point guarantees that it is outside the threshold.
https://github.com/e-mission/e-mission-server/pull/933/files#diff-807aaf746730e4cbae1d1275464f94df0230cf608e11320a28f56df1c22d4307R62

shankari (changes requested):

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: yes it does. That's how  I was able to reduce a lot of duplication.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -60,23 +128,24 @@ def testPrediction(self):
...
-            "incremental_evaluation": False
-        }
-        model = eamtg.GreedySimilarityBinning(model_config)
-        
+                    "metric": "od_similarity",
+                    "similarity_threshold_meters": 500,  # meters,
+                    "apply_cutoff": False,
+                    "clustering_way": 'origin_destination',  
+                    "incremental_evaluation": False
+                                }
```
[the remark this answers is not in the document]
humbleOldSage: fixed

emission/tests/modellingTests/TestGreedySimilarityBinning.py
```diff
@@ -95,33 +164,26 @@ def testNoPrediction(self):
...
-            "incremental_evaluation": False
-        }
-        model = eamtg.GreedySimilarityBinning(model_config)
-
+                    "metric": "od_similarity",
+                    "similarity_threshold_meters": 500,  # meters,
+                    "apply_cutoff": False,
+                    "clustering_way": 'origin_destination',  
+                    "incremental_evaluation": False
+                                }
```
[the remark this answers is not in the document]
humbleOldSage: fixed

emission/tests/modellingTests/modellingTestAssets.py
```diff
@@ -23,18 +23,26 @@ def generate_nearby_random_points(ref_coords,threshold):
...
+    # till radius length to get our new point, (34,0). When this function is called the next time to generate a point
+    #that has to be binned with previous one, we again generate r and theta , say 24 , 180 this time. 
+    # Now this new point is at (-24,0). Both these points are within threshold (100 in this case)limit and therefore will 
+    #be binned together.
+    radius=random.uniform(0,thresholdInWGS84/2)
+    theta=random.uniform(0,2*math.pi)
+    dx = radius * math.cos(theta)
+    dy = radius * math.sin (theta)
     #This basically gives a way to sample a point from within a square of length thresholdInWGS84 
     # around the ref. point.
```
shankari, line 39: nit: **(future fix)**  fix the comment to reflect "the circle" instead of "the square"

shankari (approved):

emission/tests/modellingTests/TestSimilarityMetric.py
shankari: given that we have now changed the mock trip creation code, this should still work with 500, correct?

emission/tests/modellingTests/modellingTestAssets.py
humbleOldSage: fixed.

emission/tests/modellingTests/TestSimilarityMetric.py
humbleOldSage: yes it does. 
And it is currently set to 500 in the latest commit.

emission/tests/modellingTests/TestGreedySimilarityBinning.py
humbleOldSage: This was fixed as well.

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -1,31 +1,59 @@
...
+                #generate 2 trips with parameter values
+                trips = etmm.generate_mock_trips('joe',2, threshold=similarity_threshold,origin=[0, 0], destination=[1, 1], within_threshold=2,trip_part=tp) 
+                # depending on the parametrs, extract the relevant coordinates
+                trip0_coords = metric.extract_features(trips[0])
+                trip1_coords = metric.extract_features(trips[1])
+                #check for similarity using relevant coordinates
+                similarOD = metric.similar(trip0_coords,trip1_coords, similarity_threshold,cw)
+                # Since both origin and destination poitns lie within threshold limits,they should be similar
+                # when we check by just origin or just destination or both origin-and-destination
+                self.assertTrue(similarOD)
```
shankari, line 30: nit: **(future fix)** if this fails, there is no message which indicates why. and given that it is an `assertTrue`, you won't even get the autogenerated message.

humbleOldSage:
Didn't test. there isn't anything changes that could cause failure.

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -1,31 +1,59 @@
...
+        trips = [ etmm.generate_mock_trips('joe',2, origin=[i, i], destination=[i+1, i+1], trip_part= 'od', within_threshold=1,threshold=500)[0] for i in range(n)]    
+        trip0_coord = metric.extract_features(trips[0])
+        trip1_coord = metric.extract_features(trips[1])
+
+        for cw in parameters:
+            with self.subTest(clustering_way=cw):      
+                IsSimilar = metric.similar(trip0_coord,trip1_coord, similarity_threshold,cw)
+                # Two trips with neither origin nor destination coordinates within the threshold
+                # must not be similar by any configuration of similarity testing.
+                self.assertFalse(IsSimilar)
```
shankari, line 56: ditto

emission/tests/modellingTests/TestSimilarityMetric.py
```diff
@@ -1,31 +1,59 @@
...
+        n=2
+        #this generates 2 trips one-by-one, where each trip's respective origin and destination 
+        # points are more than 500m away.
+        trips = [ etmm.generate_mock_trips('joe',2, origin=[i, i], destination=[i+1, i+1], trip_part= 'od', within_threshold=1,threshold=500)[0] for i in range(n)]    
+        trip0_coord = metric.extract_features(trips[0])
+        trip1_coord = metric.extract_features(trips[1])
+
+        for cw in parameters:
+            with self.subTest(clustering_way=cw):      
+                IsSimilar = metric.similar(trip0_coord,trip1_coord, similarity_threshold,cw)
```
shankari, line 53: nit: **(future fix)** also `IsSimilar` -> `isSimilar`

shankari (approved):

shankari:
I have some small cleanup fixes that you can address in the next PR. I plan to merge this once all the tests are done.

shankari:
@humbleOldSage  I expect you to address the cleanup comments in a subsequent PR

PR #937, opened 2023-09-29
Added initial interface for fetching inferred section modes

MukuFlash03:
The task is to add interface for the functionality to fetch inferred modes data given the cleaned sections and the user ids.
This is done to support [this PR](https://github.com/e-mission/op-admin-dashboard/pull/61) for the [op-admin-dashboard](https://github.com/e-mission/op-admin-dashboard) repository for a new feature.

- Reused implementation done by TTAlex and incorporated into `section_queries.py`. 

- The function is a part of `emissions/storage/decorations/section_queries.py` and can be accessed via `esds` interface.

- The function takes in the `cleaned section ids` and the `user ids` as a list of dictionaries to the function as a parameter.
This input format seems relevant as each section_id will be tied to a user_id which is generated during the intake pipeline process by storing these entries into the analysis timeseries db. 

If we pass a separate list of section ids and then another list of user ids, we can zip them into a combined list but may have to check whether each section ids has a user id or vice versa.
Additionally, the initial implementation by TTAlex, used the same input format too by extracting data from the returned dataframe object.

- Developed the test cases under `TestSectionQueries.py` by first generating the required sections-users list by fetching the running the intake pipeline for multiple users.

### DIFF
--- a/emission/tests/storageTests/TestSectionQueries.py
+++ b/emission/tests/storageTests/TestSectionQueries.py
@@ -11,6 +11,7 @@ import datetime as pydt
 import logging
 import uuid
 import json
+import bson.objectid as boi
 
 # Our imports
 import emission.storage.decorations.section_queries as esds
@@ -20,6 +21,10 @@ import emission.storage.timeseries.abstract_timeseries as esta
 
 import emission.core.wrapper.section as ecws
 import emission.core.get_database as edb
+import emission.core.wrapper.entry as ecwe
+import emission.tests.common as etc
+import emission.pipeline.intake_stage as epi
+import emission.core.wrapper.modeprediction as ecwm
 
 # Our testing imports
 import emission.tests.storageTests.analysis_ts_common as etsa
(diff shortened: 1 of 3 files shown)


shankari:
@MukuFlash03 for the record:
>  This input format seems relevant as each section_id will be tied to a user_id which is generated during the intake pipeline process by storing these entries into the analysis timeseries db.

This is not correct. It may seem that way when you focus on the server, but in the end to end system, the UUID is created when the user registers (aka signs up) using the app. All data in the system is then associated with the UUID.

emission/tests/storageTests/TestSectionQueries.py
```diff
@@ -54,6 +76,50 @@ def testQuerySections(self):
...
         self.assertEqual([entry.data for entry in ret_arr_list], ret_arr_time)
 
+    def testCleaned2InferredSectionList(self):
+
+        # Running the pipeline for the two user datasets
+        epi.run_intake_pipeline_for_user(self.testUUID1, skip_if_no_new_data = False)
+        epi.run_intake_pipeline_for_user(self.testUUID2, skip_if_no_new_data = False)
+
+        # Fetching the timeseries entries containing both raw data and analysis data after running intake pipeline
+        ts_agg = esta.TimeSeries.get_aggregate_time_series()
```
shankari, line 86: for clarification: this doesn't actually fetch the user entries - it just fetches the appropriate access layer.

emission/tests/storageTests/TestTimeSeries.py
```diff
@@ -36,10 +36,12 @@ def setUp(self):
         etc.setupRealExample(self, "emission/tests/data/real_examples/shankari_2015-aug-27")
 
     def tearDown(self):
+        edb.get_timeseries_db().delete_many({"user_id": self.testUUID1})
         edb.get_timeseries_db().delete_many({"user_id": self.testUUID})
+        edb.get_analysis_timeseries_db().delete_many({"user_id": self.testUUID1})
+        edb.get_analysis_timeseries_db().delete_many({"user_id": self.testUUID})
```
shankari, line 42: Good job cleaning this up. I am not going to insist on it this time, but in the future, you should really put these unrelated changes into a separate commit to keep the commit history clean. You can have multiple commits in a PR, and having separate commits will keep the "git blame" more relevant.

shankari (approved):
This is a clean PR, so it doesn't matter this time, but next time, please submit from a separate branch in case we need to delete/recreate/squash-merge etc.

PR #938, opened 2023-10-02
Survey Assist  Using RF

humbleOldSage:
[PARTIALLY TESTED] RF initialization and fit function. Build test written and tested.  fit of Random forest uses df . DBSCAN & SVM based approach added in case we want to switch from coordinates based to cluster based. Currently, in config we are passing coordinates.


emission/analysis/modelling/trip_model/forest_classifier.py
shankari, line 1: In general, once we have copied over `model.py` , it would be easier to have a separate model for RF which would **call** the methods from `models.py`. 

That would be a lot better than copy-pasting a code dump, including code for `loc_feature == cluster` which is currently unused. Think of the DRY principle aka copy-pasting is bad.

shankari (changes requested):
It is fine to include DBSCAN and clustering, but it should be in a different module instead of having a large bolus of code that is copy-pasted from one repo to another. DRY!

conf/analysis/trip_model.conf.json.sample
```diff
@@ -8,6 +8,24 @@
             "similarity_threshold_meters": 500,
             "apply_cutoff": false,
             "incremental_evaluation": false
+        },
+        "forest": {
+            "loc_feature" : "coordinates",
+            "radius": 100,
+            "size_thresh":1,
+            "purity_thresh":1.0,
```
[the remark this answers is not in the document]
humbleOldSage: Yes I agree these are clustering parameters, which are currently unused in RF model as we are using 'coordinates' config.They might be used if we decide to switch to 'cluster' config.  

However, I thought we should pass all the parameter that might be used in any configuration (clustering or coordinates) by the ForestModel together as a single parameter.

emission/analysis/modelling/trip_model/greedy_similarity_binning.py
```diff
@@ -133,6 +133,7 @@ def fit(self, trips: List[ecwc.Confirmedtrip]):
         corresponds to a label at the matching index of the label input
 
         :param trips: 2D array of features to train from
+        :param tripsdf: trips data in dataframe format
```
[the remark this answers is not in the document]
humbleOldSage: done

emission/analysis/modelling/trip_model/util.py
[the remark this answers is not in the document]
humbleOldSage: followed this now. At the top of the models.py file , I have included the path its taken from.

emission/analysis/modelling/trip_model/util.py
```diff
@@ -1,7 +1,13 @@
...
 from past.utils import old_div
-import numpy
+import numpy as np
+import pandas as pd
 from numpy.linalg import norm
+import copy
+
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.pipeline import make_pipeline
+from sklearn.impute import SimpleImputer
```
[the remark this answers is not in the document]
humbleOldSage: this is now moved to models.py.

emission/tests/modellingTests/TestForestModel.py
```diff
@@ -0,0 +1,172 @@
...
+import emission.tests.modellingTests.modellingTestAssets as etmm
+import emission.storage.decorations.analysis_timeseries_queries as esda
+import emission.core.get_database as edb
+import emission.storage.pipeline_queries as epq
+import emission.core.wrapper.pipelinestate as ecwp
+import numpy as np
+
+class TestRunForestModel(unittest.TestCase):
+    """these tests were copied forward during a refactor of the tour model
+    [https://github.com/e-mission/e-mission-server/blob/10772f892385d44e11e51e796b0780d8f6609a2c/emission/analysis/modelling/tour_model_first_only/load_predict.py#L114]
```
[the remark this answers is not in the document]
humbleOldSage: Yes. This file was pushed while shifting code from one laptop  to another during my MAC failures.

emission/tests/modellingTests/TestRunForestModel.py
```diff
@@ -0,0 +1,192 @@
...
+        prediction, n = eamur.predict_labels_with_n(
+            trip = test,
+            model_type=eamumt.ModelType.RANDOM_FOREST_CLASSIFIER,
+            model_storage=eamums.ModelStorage.DOCUMENT_DATABASE,
+            model_config=forest_model_config
+        )
+
+        [logging.debug(p) for p in sorted(prediction, key=lambda r: r['p'], reverse=True)]
+
+        self.assertNotEqual(len(prediction), 0, "should have a prediction")
```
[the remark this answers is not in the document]
humbleOldSage: Yes. This file was pushed while shifting code from one laptop  to another during my MAC failures.

emission/analysis/modelling/trip_model/forest_classifier.py
humbleOldSage: Yeah. This is better. Done.

emission/analysis/modelling/trip_model/util.py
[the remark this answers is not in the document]
shankari: Did you try "importing it directly with full history and committing it here"?
This seems to be copied over and re-committed.
Also, when you plan to finally finish https://github.com/e-mission/e-mission-eval-private-data/pull/37 so that we can 

> change the code in the notebook to call the code from the server, so that again, we will be testing the code that we publish.

emission/analysis/modelling/trip_model/forest_classifier.py
```diff
@@ -0,0 +1,177 @@
...
+        #
+        # purpose_pred | purpose_proba | mode_pred | mode_proba | replaced_pred | replaced proba 
+        # dog-park     |   1.0         |  e-bike   | 0.99       | walk          | 1.1 
+        #
+        #
+        # However, to keep the trip model general, the forest model is expected to return  
+        #
+        #PREDICTIONS [  {'labels': {'mode_confirm': 'e-bike', 'replaced_mode': 'walk', 'purpose_confirm': 'dog-park'},
+        #                 'p':  ( Currently average of the 3 probabilities)}]
+        labels= {
```
shankari, line 108: I don't see why we have to change this to the average of the three probabilities instead of having separate ones. I don't see why this is more general than the previous approach.

emission/analysis/modelling/trip_model/forest_classifier.py
```diff
@@ -0,0 +1,177 @@
...
+        data={}
+        attr=[ 'purpose_predictor','mode_predictor','replaced_predictor','purpose_enc','mode_enc','train_df']
+        if self.model.loc_feature == 'cluster':
+            ## confirm this includes all the extra encoders/models
+            attr.extend([ 'cluster_enc','end_cluster_model','start_cluster_model','trip_grouper'])
+        for attribute_name in attr:
+            buffer=BytesIO()
+            joblib.dump(getattr(self.model,attribute_name),buffer)
+            buffer.seek(0)
+            data[attribute_name]=buffer.getvalue()
```
shankari, line 134: Is this the recommended approach to serialize a random forest model? I have serialized random forest models before and this does not look familiar. Happy to keep it if it is in fact the currently recommended method, but need a citation

emission/analysis/modelling/trip_model/util.py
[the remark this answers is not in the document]
shankari: I still see this form
https://github.com/e-mission/e-mission-server/pull/938/files#diff-9fa11fd632d4ab408567f01777eef3ae0a89b8f66af3cda8b749c58bd0d3d4b7R4

emission/analysis/modelling/trip_model/model_type.py
```diff
@@ -24,16 +26,16 @@ def build(self, config=None) -> eamuu.TripModel:
...
         if model is None:
-            model_names = list(lambda e: e.name, MODELS.keys())
-            models = ",".join(model_names)
-            raise KeyError(f"ModelType {self.name} not found in factory, please add to build method")
-                
-        return model
+            available_models = ', '.join([ e.name for e in ModelType])
+            raise KeyError(f"ModelType {self.name} not found in factory, Available models are {available_models}."\
+                           "Otherwise please add new model to build method")
+        return model(config)
```
shankari, line 38: why did this code need to change?

emission/analysis/modelling/trip_model/models.py
```diff
@@ -0,0 +1,1194 @@
+########################################################################
+## Copied from /e-mission-eval-private-data/TRB_label_assist/models.py##
+########################################################################
+
```
shankari, line 4: Is this merged over with commit history? Does not appear to be

emission/core/wrapper/entry.py
```diff
@@ -182,6 +182,9 @@ def create_fake_entry(user_id, key, data, write_ts, create_id=False):
       result_entry.user_id = user_id
       result_entry.metadata = ecwm.Metadata.create_metadata_for_fake_result(key, write_ts)
       result_entry.data = data
+      #necessary values required by forest model
+      result_entry['data']['start_local_dt']=result_entry['metadata']['write_local_dt']
+      result_entry['data']['end_local_dt']=result_entry['metadata']['write_local_dt']
```
shankari, line 187: wait what?! the trip's start and end timestamps are almost always guaranteed to not be the same as the time it was created. And by putting this into `core/wrapper/entry.py`, you are adding this to every single entry of any type, created in any way.

This is just wrong.

emission/tests/modellingTests/TestForestModel.py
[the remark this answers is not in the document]
shankari: I don't see what the MAC failures, that happened several weeks ago, have to do with an extra file. You can always remove the file if it is no longer relevant. But it is not removed, so is it relevant or not?

I was complaining about:
(1) the class name being different from the file name
(2) the comment being incorrect (the tests were _hopefully_ not copied forward during the tour model refactor)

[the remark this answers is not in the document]
shankari: Ignoring this for now while waiting for clarification.

emission/tests/modellingTests/TestRunForestModel.py
shankari, line 1: This test is a fine start, but:
- I don't think it tests the code end to end.
It invokes the code using `import emission.analysis.modelling.trip_model.run_model as eamur`, which is not the method that is run directly from the pipeline. I don't see where you are testing whether or not the pipeline invokes this code, and how well it works
- I also don't see any testing around load/store. You have a completely new load and store method, how do you know that they work, and how do you know that the data stored is sufficient to predict properly when loaded.

shankari (changes requested):
I don't see any testing done filled in.
If you are adding a new test, it would be helpful to add logs showing that it works (in addition to the CI passing)
And the testing is very incomplete - it is not at all clear that this code will actually work in production
And there is one very egregious error
And it would be good to import model.py with full history

Other than that, looks good.

conf/analysis/trip_model.conf.json.sample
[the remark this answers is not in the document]
shankari: I am not sure what the config file would look like in that case. Is there a sample file? It looks like you removed it in 
https://github.com/e-mission/e-mission-server/pull/938/commits/b3d0db2b15342c5481727406052f2e46620fc4b6

emission/core/wrapper/entry.py
humbleOldSage: This is done in `create_fake_entry` function, which is used only for testing purpose

```
$ grep -rl create_fake_entry | grep -v __pycache__
./emission/core/wrapper/entry.py
./emission/tests/modellingTests/modellingTestAssets.py
```

Since `start_local_dt` and `end_local_dt' components were not required for greedy binning, they were not being added to the trips added via `generate_mock_trips`. For Random Forest model pipeline, they need to be part of the `Entry` data that we generate.

Since `TestRunRandomForest` (which is just the first test file written) simply checks the running of the pipeline, any value for now would suffice.

These values are bound to evolve as we keep adding more and more tests as per the requirements.

emission/analysis/modelling/trip_model/models.py
humbleOldSage: Currently, its Not. I was unaware of this way of moving files. 
I'll do it now.

emission/analysis/modelling/trip_model/model_type.py
humbleOldSage: Line 35-38 changed since `model` variable in Line 33  now has a different type of value. 

Previously,  in line 29-32, The MODELS dict was trying to instantiate an object directly during directory definition. This wouldn't work when we have multiple models in the dictionary, since the config parameter has values a.t. a single model.
Currently, MODELS store the class references.

emission/analysis/modelling/trip_model/models.py
humbleOldSage: I did this and was able to move models.py file with its commit history. However, the `models.py` has its dependency on `clustering.py` and `data_wrangling.py` file in `TRB_label_assist`. 

I can similarly copy `data_wrangling.py`( with its commit history) file since it doesn't seem to have any further dependencies.

However, `clustering.py` has further dependencies on other files. But the functions ( from the `clustering.py`) required by `models.py` are `get_distance_matrix` and `single_cluster_purity` . None of these two functions have further dependencies . So I can  copy the clustering.py ( along with the commit history) to e-mission-server and remove the unnecessary bit.

@shankari Let me know if this strategy work

shankari: ok, so `models.py` was *not* a direct copy from `TRB_label_assist`, you edited it further?
I don't think you should copy `clustering.py` over - you should refactor it to pull out the two functions that are dependencies and copy only the file with the pulled out code.
However, note that you have pending changes to `clustering.py` 
https://github.com/e-mission/e-mission-eval-private-data/pull/37

You should probably resolve those first, and get that PR merged before refactoring.

shankari: Note that I am also fine with importing the bulk of the python files implementing the algorithm in here. And just having the notebooks and the evaluation code in there

```
performance_eval.py
regenerate_classification_performance_results.py
```

This is particularly important if we want to retrain or reevaluate the model periodically.

emission/analysis/modelling/trip_model/util.py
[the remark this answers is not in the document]
humbleOldSage: This is further updated to include include models.py commit history.

emission/analysis/modelling/trip_model/models.py
humbleOldSage: >ok, so models.py was not a direct copy from TRB_label_assist, you edited it further?

Yes, In the previous version (when I simply dumped the code)  I removed all the classes not required by the RF model. I merged 'clustering.py' with the relevant class there and moved  a few functions  in 'utils.py' here in `e-mission-server`.

  In the current form, `model.py`( with its commit history) is the same as in `eval-private-data`  with no changes.
  
> However, note that you have pending changes to `clustering.py` [e-mission/e-mission-eval-private-data#37](https://github.com/e-mission/e-mission-eval-private-data/pull/37)
>> 
>> You should probably resolve those first, and get that PR merged before refactoring.


The two functions 'get_distance_matrix' and 'single_cluster_purity' required here are not affected  in that PR and will remain unaffected in future.

humbleOldSage: However, as mentioned we'll have to wait for https://github.com/e-mission/e-mission-eval-private-data/pull/37  to be merged before we begin this movement.

conf/analysis/trip_model.conf.json.sample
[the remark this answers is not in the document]
humbleOldSage: https://github.com/e-mission/e-mission-server/commit/b3d0db2b15342c5481727406052f2e46620fc4b6  The file deleted here is 'config_copy.py'. This is just the copy of original file which exists at same location.

emission/analysis/modelling/trip_model/forest_classifier.py
humbleOldSage: The pipeline expects the output from `predict_labels_with_n` function in `run_model.py` to be of the form ( figured this by running greedySimialrityBinning):

```
 [     {
'labels': {'mode_confirm': 'value1', 'replaced_mode': value2', 'purpose_confirm': 'value3'},

  'p':  ( a single value)}]
```

  
 Since, all three values are predicted separately they have different probs.  We can use different ways to get a single value from these three, but Avg seemed to be a good start.

emission/analysis/modelling/trip_model/forest_classifier.py
humbleOldSage: The method implemented above utilizes joblib, which was a recommended practice, especially for scikit-learn models in their documentation. joblib and pickle were the only methods I knew, I went with jolib. However there is another recommendation `skops` in the documentation. I didnt use it since I was unfamiliar with it. 

Can move to `skops` or any other way you recommend if required.

emission/analysis/modelling/trip_model/forest_classifier.py
```diff
@@ -0,0 +1,177 @@
...
+    def extract_features(self, trip: ecwc.Confirmedtrip) -> List[float]:
+        """
+        extract the relevant features for learning from a trip for this model instance
+
+        :param trip: the trip to extract features from
+        :type trip: Confirmedtrip
+        :return: a vector containing features to predict from
+        :rtype: List[float]
+        """
+        pass
```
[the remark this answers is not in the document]
humbleOldSage: will do. Its because the pre implemented models.py file handles extraction in this case.

emission/tests/modellingTests/TestRunForestModel.py
humbleOldSage: Currently, I am focused  on validating the core functionality of the model. This includes unit tests that are similar to the ones we have for other greedy binning modules, which will help us confirm that the model performs as expected in isolation.

>I also don't see any testing around load/store. You have a completely new load and store method, how do you know that they work, and how do you know that the data stored is sufficient to predict properly when loaded.

I didn't think I'll need this, but  I'll begin working on it .

Once I am confident that the model's internal operations are sound, I will proceed to develop tests for the integration with the pipeline.

humbleOldSage: To test if the model loads and stores correctly, these are the tests I am wiritng :

1.Round-Trip Test: Serialize an object with to_dict and then immediately deserialize it with from_dict. After deserialization, the object should have the same state as the original.

 2.Consistency Test: To Check if the serialization and deserialization process is consistent across multiple executions.

 3.Integrity Test: To  Verify that the serialized dictionary contains all necessary keys and that the deserialized object has all required attributes initialized correctly.
 
 4.Error Handling Test: I'll attempt to serialize and deserialize with incorrect or corrupted data to ensure that the methods handle errors gracefully.

 5.Type Preservation Test: I'll check if the serialization and deserialization process maintains the data types of all model attributes.

 6.Completeness Test: To ensure  that no parts of the model are lost during serialization or deserialization.

emission/tests/modellingTests/TestForestModel.py
```diff
@@ -88,86 +70,64 @@ def tearDown(self):
...
+                with open(file_path, 'r') as f:
+                    prev_predictions_list = json.load(f)
+                    logging.debug()
+                    self.assertEqual(prev_predictions_list,curr_predictions_list," previous predictions should match current predictions")
+            else:
+                with open(file_path,'w') as file:
+                    json.dump(curr_predictions_list,file,indent=4)
+                    logging.debug("Previous predicitons stored for future matching" )
+        except json.JSONDecodeError:
+            logging.debug("jsonDecodeErrorError")
```
humbleOldSage, line 132: This part is questionable. I wanted to know if dumping ( writing predictions of first run ever to be used as ground truth) in a JSON here is a good practice or not?

 Other option is to store the past predictions  in db. But I think the tear-down will clean it up while exiting. So I opted for JSON. Is there a way to protect the writing. Let me know if I am missing something here.

emission/tests/modellingTests/TestForestModel.py
```diff
@@ -1,84 +1,66 @@
...
+            "use_start_clusters":False,
+            "use_trip_clusters":True
+        }
+
+        existing_entries_for_user = list(self.ts.find_entries([esdatq.CONFIRMED_TRIP_KEY]))  
+        if len(existing_entries_for_user) != 0:
+            raise Exception(f"test invariant failed, there should be no entries for user {self.user_id}")
+        
+        # load in trips from a test file source
+        input_file = 'emission/tests/data/real_examples/shankari_2016-06-20.expected_confirmed_trips'
```
humbleOldSage, line 58: I opted for this file because it was used in one other tests for greedySimilarityBinning. However, for the user mentioned (above), there are just 6 trips. I wanted to confirm with you if I am free to use other files from this location.

humbleOldSage:
The  tests from `TestForestModel.py` file will fail.

humbleOldSage:
tested. supposed to fail.working on it.

humbleOldSage:
Tested. Passed Tests

humbleOldSage:
Will Fail due to `TestForestModelIntegration.py`.

humbleOldSage:
Although the current `TestForestModelIntegration.py` test works, I believe a  better way would be to load data ( possibly real world) and then build a model during the test and then run the prediction pipeline entirely and see if we got predictions at the `infer_label` stage

emission/tests/modellingTests/TestForestModelIntegration.py
```diff
@@ -0,0 +1,58 @@
...
+    # Tests that algorithm being tested runs and saves to the database correctly
+    def testIndividualAlgorithms(self):
+        logging.debug('TEST1')
+        for trip in self.inferred_trips:
+            entries = esdt.get_sections_for_trip("inference/labels", self.testUUID, trip.get_id())
+            logging.debug(f"ENTRIES: {entries}")
+            self.assertEqual(len(entries), len(self.test_algorithms))
+            # for entry in entries:
+            #     self.assertGreater(len(entry["data"]["prediction"]), 0)
+
```
humbleOldSage, line 51: This is commented for now since currently, there are no predictions since there is no model being built in this test. However, once we bring in the model building, there should be predictions here.

shankari:
Waiting for updates on the read/write and the high level discussion around why we are testing.

humbleOldSage:
Tested. All tests should pass.

humbleOldSage:
>  ....  updates on the read/write 

I believe this is in regards with the Forest model's read and write. `TestForestModelLoadandSave.py` handles this where I tried to cover testing as per my comment above. https://github.com/e-mission/e-mission-server/pull/938#discussion_r1388643461 .

>and the high level discussion around why we are testing.

1. Foremost, to make sure that the introduced feature ( forest model in this case) works as intended, even for unusual and edge cases. 
2. Integration testing to ensure it works with existing pipeline and doesn't introduce any other issues.
3. I guess using real world data is a good practice to see how feature  will perform for end user.
4. Additionally,  the real world example uncovers certain complexities that might not be visible in simplified test cases.
5. In our case, we can get idea about the accuracy and reliability of the results that user gets.

emission/analysis/modelling/trip_model/forest_classifier.py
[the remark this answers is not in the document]
humbleOldSage: This is included  in the upcoming commit.

emission/tests/modellingTests/TestForestModelLoadandSave.py
```diff
@@ -0,0 +1,313 @@
...
+            user_id=self.user_id,
+            model_type=eamumt.ModelType.RANDOM_FOREST_CLASSIFIER,
+            model_storage=eamums.ModelStorage.DOCUMENT_DATABASE,
+            model_config=self.forest_model_config        
+            )
+                
+        model_data=model.to_dict()
+        loaded_model_type=eamumt.ModelType.RANDOM_FOREST_CLASSIFIER
+        loaded_model = loaded_model_type.build(self.forest_model_config)
+        loaded_model.from_dict(model_data)
```
shankari, line 308: I am confused here. The model from `load_stored_trip_model` is called `model` and the model generated by building is `loaded_model`?

emission/tests/modellingTests/TestForestModelLoadandSave.py
```diff
@@ -0,0 +1,313 @@
...
+
+        
+        for attr in ['purpose_predictor','mode_predictor','replaced_predictor','purpose_enc','mode_enc','train_df']:
+            assert isinstance(getattr(loaded_model.model,attr),type(getattr(model.model,attr)))
+
```
shankari, line 313: why are you only checking types instead of actual values?

emission/tests/modellingTests/TestRunForestModel.py
```diff
@@ -0,0 +1,198 @@
...
+        edb.get_pipeline_state_db().delete_many({'user_id': self.user_id})
+
+    def testBuildForestModelFromConfig(self):
+        """
+        forest model takes config arguments via the constructor for testing
+        purposes but will load from a file in /conf/analysis/ which is tested here
+        """
+
+        eamumt.ModelType.RANDOM_FOREST_CLASSIFIER.build()
+        # success if it didn't throw
```
shankari, line 96: why not get the model and check some properties instead of just assuming success if it didn't throw?
e.g. something like the below (pseudo-code, do not blindly merge)

```suggestion
        built_model = eamumt.ModelType.RANDOM_FOREST_CLASSIFIER.build()
        self.assertEqual(build_model.get_type(), eamumt.ModelType.RANDOM_FOREST_CLASSIFIER.text)
        # success if it didn't throw
```

shankari (changes requested):
@humbleOldSage there are several comments that have not been addressed from the first review rounds. I have some additional comments on the tests as well. Please make sure that **alll** pending comments are addressed before resubmitting for review.

emission/analysis/modelling/trip_model/forest_classifier.py
```diff
@@ -0,0 +1,198 @@
...
+        random_forest_expected_keys = [
+            'loc_feature',
+            'n_estimators',
+            'criterion',
+            'min_samples_split',
+            'min_samples_leaf',
+            'max_features',
+            'bootstrap',
+        ]            
+        cluster_expected_keys= [
```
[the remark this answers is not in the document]
humbleOldSage: This is  is used when we cluster the coordinates (loc_cluster parameter = True) before passing to Random Forest .  However, Since we are  working with configuration where loc_cluster is `False` with RF , I'll mention the same there and  comment this and other related parts.

emission/analysis/modelling/trip_model/forest_classifier.py
```diff
@@ -0,0 +1,198 @@
...
+                                     criterion=config['criterion'],
+                                     max_depth=maxdepth, 
+                                     min_samples_split=config['min_samples_split'],
+                                     min_samples_leaf=config['min_samples_leaf'],
+                                     max_features=config['max_features'],
+                                     bootstrap=config['bootstrap'],
+                                     random_state=config['random_state'],
+                                     # drop_unclustered=False,
+                                     use_start_clusters=config['use_start_clusters'],
+                                     use_trip_clusters=config['use_trip_clusters'])
```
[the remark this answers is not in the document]
humbleOldSage: yeah, I have used warts now.

emission/analysis/modelling/trip_model/run_model.py
```diff
@@ -57,7 +57,6 @@ def update_trip_model(
         logging.debug(f'time query for training data collection: {time_query}')
 
         trips = _get_training_data(user_id, time_query)
-
```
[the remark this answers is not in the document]
humbleOldSage: I haven't done any changes here. I don't see any whitespace either.

emission/analysis/modelling/trip_model/util.py
```diff
@@ -1,9 +1,9 @@
 from typing import List, Tuple
 from past.utils import old_div
-import numpy
+import numpy as np
+import pandas as pd
 from numpy.linalg import norm
 
-
```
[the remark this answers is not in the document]
humbleOldSage: I figure its a blank line to separate imports from rest of the code.

emission/tests/modellingTests/TestForestModelIntegration.py
```diff
@@ -0,0 +1,132 @@
...
+            self.assertEqual(len(entries), len(self.test_algorithms))
+            for entry in entries:
+                self.assertGreater(len(entry["data"]["prediction"]), 0)
+                for singleprediction in entry["data"]["prediction"]:
+                    self.assertIsInstance(singleprediction, dict, " should be an instance of the dictionary class")
+                    self.assertIsInstance(singleprediction['labels'], dict, " should be an instance of the dictionary class")
+                    self.assertIn('mode_confirm',singleprediction['labels'].keys())
+                    self.assertIn('replaced_mode',singleprediction['labels'].keys())
+                    self.assertIn('purpose_confirm',singleprediction['labels'].keys())        
+
```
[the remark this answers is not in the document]
humbleOldSage: > This just checks that the properties exist, not that they are correct. 

The test here ensures that the model has the required properties to ensure it  can make predictions without errors. Correctness checking of the predictions would require hard coding these predictions in tests which might change as the model evolves ( even for better). In such a case, even if the model improved, the test would fail ( since the predictions changed). 

>When will the "more and more tests" from [#938 (comment)](https://github.com/e-mission/e-mission-server/pull/938#discussion_r1381828612) be implemented?

Right now we have test to check if the model loads and saves correctly, if it builds correctly, runs correctly and if it integrates in the pipeline as expected.

 I am unable to think of anymore tests  from my limited knowledge. Can we think brainstorm and think what more can be added?

emission/tests/modellingTests/TestForestModelLoadandSave.py
```diff
@@ -0,0 +1,313 @@
...
+        self.has_label_percent = 0.9 # let's make a few that don't have a label, but invariant
+                                # $clustered_trips * $has_label_percent > self.min_trips
+                                # must be correct or else this test could fail under some random test cases.
+
+        # for a negative test, below
+        self.unused_user_id = 'asdjfkl;asdfjkl;asd08234ur13fi4jhf2103mkl'
+
+        # test data can be saved between test invocations, check if data exists before generating
+        ts = esta.TimeSeries.get_time_series(user_id)
+        test_data = list(ts.find_entries(["analysis/confirmed_trip"]))
```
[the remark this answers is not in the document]
humbleOldSage: True. Here we are checking if there were any trips left after tear down of last run. The comment was misleading. I have changed the comment to imply the same.

emission/tests/modellingTests/TestForestModelLoadandSave.py
```diff
@@ -0,0 +1,313 @@
...
+            )
+
+            ts.bulk_insert(train)
+
+            # confirm data write did not fail
+            test_data = esda.get_entries(key="analysis/confirmed_trip", user_id=user_id, time_query=None)
+            if len(test_data) != self.total_trips:
+                logging.debug(f'test invariant failed after generating test data')
+                self.fail()
+            else:
```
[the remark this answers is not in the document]
humbleOldSage: Fixed an error on line 59.

> moving this out of the setup would also ensure that we don't have to call `self.fall()`

Since this is a part of setup, wouldn't it be better if we ensure that the setup is successful and then proceed towards our tests?

emission/tests/modellingTests/TestForestModelLoadandSave.py
humbleOldSage: Improved  with with better identifiers

emission/tests/modellingTests/TestForestModelLoadandSave.py
humbleOldSage: Included a pre and post serialization-deserialisation prediction assertion on a random user now.

emission/tests/modellingTests/TestRunForestModel.py
humbleOldSage: I am now checking the type of built model, along with the types of other 5 attributes.

emission/analysis/modelling/trip_model/forest_classifier.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def to_dict(self):
+        """
+        Convert the model to a dictionary suitable for storage.
+        """
+        data={}
+        attr=[ 'purpose_predictor','mode_predictor','replaced_predictor','purpose_enc','mode_enc','train_df']
+        if self.model.loc_feature == 'cluster':
+            ## confirm this includes all the extra encoders/models
+            attr.extend([ 'cluster_enc','end_cluster_model','start_cluster_model','trip_grouper'])
```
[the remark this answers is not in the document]
humbleOldSage: Even for JSON serialization, we'll have to work attribute by attribute, since each instance here contain instances of other classes (`ForestClassifierModel` has `ForestClassifier` which has 3 `RandomForestClassifiers` and 2 `OneHotWrappers` ( this further has oneHotEncoders). We'll have to add for every class its own serialization method  ( since JSON only serializes basic data structures by default) so that `json.dumps (obj)` can work.

emission/analysis/modelling/trip_model/models.py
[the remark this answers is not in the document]
humbleOldSage: I haven't done this yet since this file has other models as well ( this was brought in as it is from the `e-mission-eval-private-data` )

shankari:
Tagging @JGreenlee this is an MLOps change, we are changing the survey assist algorithm to use random forest instead of the old binning algorithm. As you can see, the code is set up to have a common structure for survey assist algorithms (find trips, train, save the model) and (load model, find prediction, store prediction). LMK if you need help understanding the structure.

We already had an implementation for the binning algorithm, this is an implementation of a second algorithm. LMK if you have any high level questions related to this.

Note that one UI implication for this is that, after the change, we will always have inferred labels for trips. With the previous binning algorithm, if it was a novel trip, there would not be any matches. But the random forest will generate predictions for every input, hopefully flagging that it is crappy. I believe we already ignore predictions with too low probability, but we should verify that it works properly end to end.

Again, LMK if you have high-level questions on the underlying models

emission/analysis/modelling/trip_model/forest_classifier.py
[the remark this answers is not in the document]
humbleOldSage: This is done.

PR #944, opened 2023-11-09
Scalability fixes - Load model once per user

MukuFlash03:
Implementing code changes for improving scalability as per the requirements in [this issue](https://github.com/e-mission/e-mission-docs/issues/950).

Initial approach involves utilizing Singleton design pattern concept on instance variables to check whether model has already been loaded before attempting to load the model.

Thus this should prevent model from being loaded if it is already present in the instance variable of the current class instance of BuiltinModelStorage.

More details present [here](https://github.com/e-mission/e-mission-docs/issues/950#issuecomment-1802918055).

### DIFF
--- a/emission/core/get_database.py
+++ b/emission/core/get_database.py
@@ -37,8 +37,17 @@ print("Connecting to database URL "+url)
 _current_db = MongoClient(url, uuidRepresentation='pythonLegacy')[db_name]
 #config_file.close()
 
+# Store the latest model globally for implementing controlled access and allow model loading only once
+_model_db = None
+
 def _get_current_db():
     return _current_db
+
+def _get_model_db():
+    return _model_db
+
+def _set_model_db(model_db):
+    _model_db = model_db
 
 def get_token_db():
     Tokens= _get_current_db().Stage_Tokens
@@ -99,7 +108,6 @@ def update_routeDistanceMatrix_db(user_id, method, updatedMatrix):
     f = open('routeDistanceMatrices/' + user_id + '_' + method + '_routeDistanceMatrix.json', 'w+')
     f.write(json.dumps(updatedMatrix))
     f.close()   
-
 
 def get_client_db():
     # current_db=MongoClient().Stage_database
@@ -231,10 +239,16 @@ def get_model_db():
     " will eventually delete them. This means that the elements are essentially
     " getting updated, only over time and as a log-structured filesystem.
     """
-    ModelDB = _get_current_db().Stage_updateable_models
-    ModelDB.create_index([("user_id", pymongo.ASCENDING)])
-    ModelDB.create_index([("metadata.key", pymongo.ASCENDING)])
-    ModelDB.create_index([("metadata.write_ts", pymongo.DESCENDING)])
+    ModelDB = _get_model_db()
+    if ModelDB == None:
+        logging.debug("Started model load in edb.get_model_db()...")
+        ModelDB = _get_current_db().Stage_updateable_models
+        ModelDB.create_index([("user_id", pymongo.ASCENDING)])
+        ModelDB.create_index([("metadata.key", pymongo.ASCENDING)])
+        ModelDB.create_index([("metadata.write_ts", pymongo.DESCENDING)])
+        _set_model_db(ModelDB)
+        logging.debug("Finished model load in edb.get_model_db()...")
+    logging.debug("Fetched model in edb.get_model_db()")       
     return ModelDB
 
 def _create_analysis_result_indices(tscoll):
--- a/emission/storage/modifiable/builtin_model_storage.py
+++ b/emission/storage/modifiable/builtin_model_storage.py
@@ -15,6 +15,13 @@ class BuiltinModelStorage(esma.ModelStorage):
         super(BuiltinModelStorage, self).__init__(user_id)
         self.key_query = lambda key: {"metadata.key": key}
         self.user_query = {"user_id": self.user_id} # UUID is mandatory for this version
+        self.current_model = None
+
+    def _get_model():
+        return self.current_model
+
+    def _set_model(model):
+        self.current_model = model
 
     def upsert_model(self, key:str, model: ecwb.WrapperBase):
         """
@@ -34,12 +41,18 @@ class BuiltinModelStorage(esma.ModelStorage):
         :return: the most recent database entry for this key
         """
         find_query = {"user_id": self.user_id, "metadata.key": key}
-        result_it = edb.get_model_db().find(find_query).sort("metadata.write_ts", -1).limit(1)
-        # this differs from the timeseries `get_first_entry` only in the find query
-        # and the fact that the sort key and sort order are hardcoded
-        # everything below this point is identical
-        # but it is also fairly trivial, so I am not sure it is worth pulling
-        # out into common code at this point
+        result_it = _get_model()
+        if result_it == None:
+            logging.debug("Started model load in builtin_model_storage.get_current_model()...")
+            result_it = edb.get_model_db().find(find_query).sort("metadata.write_ts", -1).limit(1)
+            # this differs from the timeseries `get_first_entry` only in the find query
+            # and the fact that the sort key and sort order are hardcoded
+            # everything below this point is identical
+            # but it is also fairly trivial, so I am not sure it is worth pulling
+            # out into common code at this point
+            _set_model(result_it)
+            logging.debug("Finished model load in builtin_model_storage.get_current_model()...")
+        logging.debug("Fetched model in builtin_model_storage.get_current_model()...")
         result_list = list(result_it)
         if len(result_list) == 0:
             return None


emission/core/get_database.py
```diff
@@ -37,9 +38,18 @@
 _current_db = MongoClient(url, uuidRepresentation='pythonLegacy')[db_name]
 #config_file.close()
 
+# Store the latest model globally for implementing controlled access and allow model loading only once
+_model_db = None
+
```
shankari, line 43: you do not need this since `current_db` is already a singleton.

emission/core/get_database.py
```diff
@@ -37,9 +38,18 @@
...
 def _get_current_db():
     return _current_db
 
+def _get_model_db():
+    return _model_db
+
+def _set_model_db(model_db):
+    _model_db = model_db
+
```
shankari, line 52: Getting a reference to the DB is not the same as loading data.

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -15,6 +15,13 @@ def __init__(self, user_id):
         super(BuiltinModelStorage, self).__init__(user_id)
         self.key_query = lambda key: {"metadata.key": key}
         self.user_query = {"user_id": self.user_id} # UUID is mandatory for this version
+        self.current_model = None
```
shankari, line 18: there is not only one model per user
https://github.com/e-mission/e-mission-docs/issues/950#issuecomment-1803287535

> Note that there is not really "the model" in the interface. Instead, all the interface methods take a key. Right now, we have only one model (the trip model), but we may have a user model or a tour model or other models in the future.

shankari (changes requested):
Needs redesign

MukuFlash03:
[SOLVED]
Two pipeline tests were failing - TestLabeIInferencePipeline, TestExpectationPipeline.
```
FAIL: testPipelineIntegrity (pipelineTests.TestExpectationPipeline.TestExpectationPipeline)
web-server_1  | ----------------------------------------------------------------------
web-server_1  | Traceback (most recent call last):
web-server_1  |   File "/src/e-mission-server/emission/tests/pipelineTests/TestExpectationPipeline.py", line 76, in testPipelineIntegrity
web-server_1  |     self.assertGreaterEqual(len(self.expected_trips), 1)  # Make sure we've actually loaded trips
web-server_1  | AssertionError: 0 not greater than or equal to 1
```
Took a look at the logs and saw that some more functions were called on running pipeline in these tests.
Failed since they were calling placeholder_predict_n functions in eacili inferrers.py
```
TypeError: placeholder_predictor_3() takes 1 positional argument but 2 were given

```

So, added a quick fix to update the functions so that they now receive trip_list instead of single trip and pass in a list of trips.

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -55,48 +57,67 @@ def run_prediction_pipeline(self, user_id, time_range):
...
             if self._last_trip_done is None or self._last_trip_done["data"]["end_ts"] < cleaned_trip["data"]["end_ts"]:
                 self._last_trip_done = cleaned_trip
+        print(f"{arrow.now()} Inside run_prediction_pipeline: Saving inferred_trip total time = {time.process_time() - start_insert_inferred_trip_time}")
     
     # This is where the labels for a given trip are actually predicted.
     # Though the only information passed in is the trip object, the trip object can provide the
     # user_id and other potentially useful information.
-    def compute_and_save_algorithms(self, trip):
-        predictions = []
+    def compute_and_save_algorithms(self, user_id, trip_list):
```
shankari, line 91: why do we need to pass in the `user_id` here?

emission/analysis/modelling/trip_model/run_model.py
```diff
@@ -97,7 +98,8 @@ def update_trip_model(
 
 
 def predict_labels_with_n(
-    trip: ecwc.Confirmedtrip,
+    user_id: UUID,
```
shankari, line 101: you should still be able to infer the `user_id` from the trip_list. you just have to assert that all the user_ids are the same and then use that unique one.

emission/core/get_database.py
```diff
@@ -320,4 +320,4 @@ def save(db, entry):
         # logging.debug("entry has id, calling with match, result = %s" % result.raw_result)
     else:
         result = db.insert_one(entry)
-        # logging.debug("entry has id, calling without match, result = %s" % result.inserted_id)
+        # logging.debug("entry has id, calling without match, result = %s" % result.inserted_id)
```
shankari, line 323: extraneous whitespace?

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -47,4 +47,3 @@ def get_current_model(self, key:str) -> Optional[Dict]:
             first_entry = result_list[0]
             del first_entry["_id"]
             return first_entry 
-
```
shankari, line 50: ditto

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -461,8 +461,6 @@ def find_entries_count(self, key_list = None, time_query = None, geo_query = Non
 
         For key_list = None or empty, total count of all documents are returned considering the matching entries from entire dataset.
         """
-        print("builtin_timeseries.find_entries_count() called")
-
```
shankari, line 465: why did you remove this?

emission/tests/storageTests/TestSectionQueries.py
```diff
@@ -100,7 +100,6 @@ def testCleaned2InferredSectionList(self):
         # Total = 25 = 10 (UUID1) + 15 (UUID2)
         curr_predicted_entries = esds.cleaned2inferred_section_list(section_user_list)
         self.assertEqual(len(curr_predicted_entries), len(sections_entries))
-        print(curr_predicted_entries)
```
shankari, line 103: why did you remove this?

emission/tests/storageTests/TestTimeSeries.py
```diff
@@ -184,9 +184,6 @@ def testFindEntriesCount(self):
         count_ts9 = ts_new_user.find_entries_count(key_list=key_list1)
         self.assertEqual(count_ts9, 0)
 
-        print("Assert Test for Count Data successful!")
-        
-
```
shankari, line 189: Ditto

emission/analysis/classification/inference/labels/inferrers.py
shankari, line 1: Please see comments are remove `user_id`, to match the new function signature

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -55,48 +57,67 @@ def run_prediction_pipeline(self, user_id, time_range):
...
-            lp.prediction = prediction
-            lp.start_ts = trip["data"]["start_ts"]
-            lp.end_ts = trip["data"]["end_ts"]
-            self.ts.insert_data(self.user_id, "inference/labels", lp)
-            predictions.append(lp)
-        return predictions
+            prediction_list = algorithm_fn(user_id, trip_list)
+            start_insert_inference_labels_time = time.process_time()
+            for trip, prediction in zip(trip_list, prediction_list):
+                lp = ecwl.Labelprediction()
```
shankari, line 97: **future fix??**: my one comment here is that if we standardize on an interface for the `algorithm_fn`, I believe we can simplify this further.

This can become (pseudocode):

```
for algorithm_id, algorithm_fn in primary_algorithms.items():
   algo_model = algorithm.load_model()
   for trip in trip_list:
      prediction = model.predict(trip)
      # and then the original code
```

Back when this code was written, `trip_model`, with its clean interfaces to load, predict and save, had not yet been implemented. So `algorithm_fn` was supposed to be an abstraction layer of the multiple algorithms.

But now that we have the abstractions in trip_model, I am not sure that we need two levels of abstraction. I am open to deferring the cleanup until the next round, but I am also open to just removing one level of abstraction right now, since that will result in far fewer changes.

emission/analysis/modelling/trip_model/run_model.py
```diff
@@ -110,13 +112,21 @@ def predict_labels_with_n(
...
     """
-    user_id = trip['user_id']
+    start_model_load_time = time.process_time()
     model = _load_stored_trip_model(user_id, model_type, model_storage, model_config)
-    if model is None:
-        return [], -1
-    else:
-        predictions, n = model.predict(trip)
-        return predictions, n
+    print(f"{arrow.now()} Inside predict_labels_n: Model load time = {time.process_time() - start_model_load_time}")
```
shankari, line 117: This clearly does capture the load time for the single load. Can you clarify how you captured the load time in the old implementation?

`2023-11-21T10:21:17.391582-07:00 Inside run_prediction_pipeline: Model load time = 14.285613000000001`

Since there were multiple loads, you would have needed to use an accumulator for it to be correct.

shankari (changes requested):

emission/storage/timeseries/builtin_timeseries.py
MukuFlash03: I had these print statements while working on their respective issues. I had originally meant to use them just while testing but forgot to remove them. Hence thought of cleaning the output statements generated.

emission/tests/storageTests/TestSectionQueries.py
MukuFlash03: Same as above; removing this as I had added them earlier.

emission/tests/storageTests/TestTimeSeries.py
MukuFlash03: Same as above; removing this as I had added them earlier.

emission/analysis/classification/inference/labels/pipeline.py
shankari: Right now, there is a map that has algorithm name to algorithm fn and then the algorithm fn (in our case) calls `predict_with_n`. instead, the map can have algorithm name to dict of model parameters, and then inside the for loop, we can just load the model and continue to predict trip by trip and not load the model within the predict function (which works, but is not pretty).

Alternatively, we can keep the algorithm function, `eacili.predict_cluster_confidence_discounting` but load the model in there. But I don't see the need for the second level of abstraction if we assume that all our algorithms will follow the L -> P -> S workflow.

shankari: If you are not comfortable with removing the `algorithm_fn` abstraction, you can simply do the following 2 lines of code to be changed.

```
def predict_cluster_confidence_discounting(trip, max_confidence=None, first_confidence=None, confidence_multiplier=None):
    # load application config
    model_type = eamtc.get_model_type()
    model_storage = eamtc.get_model_storage()
    ### load model  <------ NEW CODE
    labels, n = eamur.predict_labels_with_n(trip, model) <----- CHANGED CODE
    if n <= 0:  # No model data or trip didn't match a cluster
        logging.debug(f"In predict_cluster_confidence_discounting: n={n}; returning as-is")
        return labels

    confidence_coeff = n_to_confidence_coeff(n, max_confidence, first_confidence, confidence_multiplier)
    logging.debug(f"In predict_cluster_confidence_discounting: n={n}; discounting with coefficient {confidence_coeff}")

    labels = copy.deepcopy(labels)
    for l in labels: l["p"] *= confidence_coeff
    return labels
```

emission/analysis/modelling/trip_model/run_model.py
MukuFlash03: Yes, I did accumulate the times for each trip in the original implementation.
I used global and module level variables in eamt.run_model and eacil.pipeline.
In eacil.pipeline I imported the variables that summed up the load times in eamt.run_model.

In eamt.run_model
```
model_load_time = 0
predict_time = 0
def predict_labels_with_n():
# Just mentioning code used for time-keeping
    start_model_load_time = time.process_time()
    model = _load_stored_trip_model(user_id, model_type, model_storage, model_config)
    global model_load_time
    model_load_time += time.process_time() - start_model_load_time
    # print(f"Inside predict_labels_n: Model load time = {time.process_time() - start_model_load_time}")
    # print(f"{arrow.now()} Inside predict_labels_n: Model load time = {model_load_time}")
```

In eacil.pipeline
```
import emission.analysis.modelling.trip_model.run_model as eamtr

def run_prediction_pipeline(): 
# Just mentioning code used for time-keeping
        print(f"{arrow.now()} Inside run_prediction_pipeline: Model load time = {eamtr.model_load_time}")
        print(f"{arrow.now()} Inside run_prediction_pipeline: Predictions complete for trip_list in time = {eamtr.predict_time}")
```

emission/analysis/classification/inference/labels/pipeline.py
MukuFlash03: Since I was passing trip_list which was originally initialized as empty, I thought in case of an empty list there would be no way to retrieve user_id.
But, yes it would be ideal to extract user_id from trip_list and as you suggested, can assert for identical user_id.
I've added the code for this and also ensured that the prediction doesn't occur for an empty trip_list by checking this in eacil.pipeline.py before calling the predictio algorithm functions.

emission/core/get_database.py
MukuFlash03: Yes, removed an extra empty line at the end.

emission/analysis/classification/inference/labels/pipeline.py
MukuFlash03: I have a few queries which are still confusing to me.

> But now that we have the abstractions in trip_model, I am not sure that we need two levels of abstraction.

1.  By "two levels", I believe you are referring to: 
```
eacili.predict_cluster_confidence_discounting -> eamur.predict_labels_with_n -> model.predict()
```
I see that trip_model is an abstract class and one child class / concrete implementation is GreedySimilarityBinning class. 
**Do we intend on removing the calls to the first two functions in the above chain and directly call only model.predict() which is in effect a call to GreedySimilarityBinning.predict() depending on the model parameters?**

------------

> Right now, there is a map that has algorithm name to algorithm fn and then the algorithm fn (in our case) calls `predict_with_n`. instead, the map can have algorithm name to dict of model parameters...

> Alternatively, we can keep the algorithm function, `eacili.predict_cluster_confidence_discounting` but load the model in there. But I don't see the need for the second level of abstraction if we assume that all our algorithms will follow the L -> P -> S workflow.

2. I believe I've understood that the same trip model will be loaded for each user irrespective of algorithm if we move up the model loading, model parameters up into `eacilp.pipeline`.

If this is correct, then yes, using a map of algorithm_name to dict_model_parameters is useful:
```
primary_algorithms = {
    ecwl.AlgorithmTypes.CONFIDENCE_DISCOUNTED_CLUSTER: dict_model_params
}

# In the for loop
   # Load model
   for trip in trip_list:
      prediction = model.predict(trip)
```

But in doing this (in continuation with my 1st question), this will call model.predict(), and I fail to see how our algorithm_name will have a unique implementation if we are removing the call to its respective function?
For instance, on removing the call to `eacili.predict_cluster_confidence_discounting`, we are no longer doing the `confidence_coeff`calculations. 
**So what would be the significance of the algorithm name in the map in this case?**

We also have placeholder_predictor() functions in inferrers.py which are being called in TestLabelInferencePipeline.py through the compute_and_save_algorithms() as well.
I believe for these functions, loading model and making predictions won't be useful as these don't load the model and just return predictions.

emission/analysis/classification/inference/labels/pipeline.py
```diff
@@ -55,48 +57,68 @@ def run_prediction_pipeline(self, user_id, time_range):
...
+        ensemble_list = []
+
+        # Create list of inferred trips
+        for cleaned_trip in cleaned_trip_list:
             cleaned_trip_dict = copy.copy(cleaned_trip)["data"]
             inferred_trip = ecwe.Entry.create_entry(user_id, "analysis/inferred_trip", cleaned_trip_dict)
-            
+            inferred_trip_list.append(inferred_trip)
+    
+        if inferred_trip_list:
```
shankari, line 72: isn't this always going to be true, given that you define `inferred_trip_list` in line 62?
Are you actually testing against the "no trip" case?
I suspect this doesn't matter, since we just iterate through the trips downstream, but then we can just skip this check altogether given that it will always be true.

emission/analysis/classification/inference/labels/inferrers.py
```diff
@@ -141,18 +162,34 @@ def n_to_confidence_coeff(n, max_confidence=None, first_confidence=None, confide
...
-    logging.debug(f"In predict_cluster_confidence_discounting: n={n}; discounting with coefficient {confidence_coeff}")
 
-    labels = copy.deepcopy(labels)
-    for l in labels: l["p"] *= confidence_coeff
-    return labels
+    # assert and fetch unique user id for trip_list
+    user_id_list = []
+    for trip in trip_list:
+        user_id_list.append(trip['user_id'])
+    assert user_id_list.count(user_id_list[0]) == len(user_id_list), "Multiple user_ids found for trip_list, expected unique user_id for all trips"
```
shankari, line 174: nit, future fix: you should add some additional details to the log - what were the unique user IDs, how many is "multiiple"? Otherwise, if I see this log, I am not sure how to debug it.

shankari (approved):
I am going to merge this for now, but please note the "future fix" comments and address them in a cleanup pass. You should also look through all the logs that you have added and make sure that you are logging as much detail as possible, so it can actually be useful in understanding the problem.

emission/analysis/classification/inference/labels/pipeline.py
MukuFlash03: I did test this out and Python evaluates an empty list to be False when used in an if condition.
So, it does not enter the if block unless trips are loaded into `inferred_trip_list`  in the for loop above the if block and the list becomes non-empty which then evaluates to True.

PR #945, opened 2023-12-01
Receive models from emission eval private data

humbleOldSage:
Moving all the models related files from `e-mission-eval-private-data`  to `e-mission-server`.  Four files, as below, are moved from `TRB_label_assist` to  `emission/analysis/modelling/trip_model` :

1. `models.py`
2. `clustering.py`
3. `mapping.py`
4. `data_wrangling.py`

I'll link the PR that handles changes on `e-mission-eval-private-data` side below once I have it ready. This way it'll be easier to track changes on both sides.


shankari:
@humbleOldSage this has extraneous commits as well, including commits that are completely unrelated to this (e.g. 'check in percom analysis before we forget"). We should copy these files over with the commit history __related to them__, not the entire commit history of the repository.

PR #946, opened 2023-12-01
User Customize Labels

jiji14:
### Related Issue
- System Design: [GitHub Issue 1012](https://github.com/e-mission/e-mission-docs/issues/1012)
- Frontend: [Customize Label Step1](https://github.com/e-mission/e-mission-phone/pull/1110), [Customize Labe Step2](https://github.com/e-mission/e-mission-phone/pull/1120)

### What I added
1. **getUserCustomLabels:** Retrieve user's custom labels with keys(mode, purpose, replaced_mode) from the database and return a string array sorted by frequency.
2. **insertUserCustomLabel:** Insert the user's custom label with a key in the database and then return the updated label.
3. **updateUserCustomLabel:** Update the user's custom label with a key, including frequency, in the database and then return the updated label.
4. **deleteUserCustomLabel**: Change the user's custom label with a key as !isActive in the database and then return the updated modes.

### Unit Tests
1. **getUserCustomLabels**
2. **InsertUserCustomLabel**
3. **UpdateUserCustomLabel**
4. **DeleteUserCustomLabel**

<br />

### Result

[ Database ]

Screenshot 2023-12-13 at 2 01 24 PM

[ UI ]

ui1
ui2
ui3

<br />


emission/core/wrapper/user.py
```diff
@@ -127,10 +127,14 @@ def getSettings(self):
 
   @staticmethod
   def createProfile(uuid, ts):
-    initProfileObj = {'user_id': uuid,
+    initProfileObj = {
+                      'user_id': uuid,
                       'source':'Shankari',
```
shankari, line 132: nit: can eventually remove this `source` since it is hardcoded and never used. I don't remember the full history (will need to check the blame) but IIRC, it was an early attempt at multi-tenancy (e.g. users from my team versus from other teams at Berkeley)

shankari (approved):

PR #948, opened 2023-12-11
Store limited number of models and clear older ones

MukuFlash03:
Clearing or trimming number of models for a particular user from the Stage_updateable_models collection.
Constant K_MODELS_COUNT defines the number of models to keep.
Comparing with write_ts of Kth model entry to decide how many models to remove which are older than this write_ts.

Currently tested with model data available in stage dataset snapshot.
DONE - Testing done by generating mock models and cleaning them up.


emission/analysis/classification/inference/labels/inferrers.py
```diff
@@ -6,6 +6,7 @@
 import copy
 import time
 import arrow
+from uuid import UUID
```
shankari, line 9: why do we need this import?

emission/analysis/classification/inference/labels/inferrers.py
shankari, line 1: these are additional log statements, unrelated to storing a limited number of models.
Please move into separate PR

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -6,11 +6,15 @@
 
 import emission.core.get_database as edb
 import emission.storage.modifiable.abstract_model_storage as esma
+import emission.storage.decorations.analysis_timeseries_queries as esda
```
shankari, line 9: what do we need this import for?

shankari (changes requested):
I am stopping at this point because there are multiple comments related to , please clean up this PR and re-submit for review.

.docker/setup_config.sh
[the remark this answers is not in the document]
MukuFlash03: Removed these changes.

This PR was opened prior to the latest merged changes for the security fixes.
And then I continued working on them after the merged security fixes.
Hence, I thought of keeping the code updated and pulled the latest changes from master.
But yes, you would have been able to see those changes anyways when you would have merged the code.

emission/analysis/classification/inference/labels/inferrers.py
MukuFlash03: Might have added to test locally, removed from latest fixes.

emission/storage/modifiable/builtin_model_storage.py
MukuFlash03: Might have added to test locally, but is currently an unused import.
Removed from latest fixes.

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -23,9 +23,13 @@ def upsert_model(self, key:str, model: ecwb.WrapperBase):
         """
         logging.debug("upsert_doc called with key %s" % key)
         entry = ecwe.Entry.create_entry(self.user_id, key, model)
+        ## TODO: Cleanup old/obsolete models
```
shankari, line 26: why is this still a TODO? You have just done this, right?

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -23,9 +23,13 @@ def upsert_model(self, key:str, model: ecwb.WrapperBase):
...
+        # Cleaning up older models, before inserting new model
+        self.trim_model_entries(key)
         logging.debug("Inserting entry %s into model DB" % entry)
         ins_result = edb.get_model_db().insert_one(entry)
-        ## TODO: Cleanup old/obsolete models
+        new_model_count = edb.get_model_db().count_documents({"user_id": self.user_id})
+        logging.debug("New model count for user %s = %s" % (self.user_id, new_model_count))
```
shankari, line 32: I think it would be more useful to log in the change the number of modules - e.g. something like

```suggestion
        logging.debug("model count for user %s has changed %s -> %s" % (self.user_id, old_model_count, new_model_count))
```

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -48,3 +52,34 @@ def get_current_model(self, key:str) -> Optional[Dict]:
...
+                "metadata.key" : key,
+                "metadata.write_ts" : { "$lte" : write_ts_limit }
+            }
+            models_to_delete = edb.get_model_db().delete_many(filter_clause)
+            if models_to_delete.deleted_count > 0:
+                logging.debug(f"{models_to_delete.deleted_count} documents deleted successfully\n")
+            else:
+                logging.debug("No documents found or none deleted\n")
+        new_model_count = edb.get_model_db().count_documents({"user_id": self.user_id})
+        logging.debug("After trimming, model count for user %s = %s" % (self.user_id, new_model_count))
```
shankari, line 85: Or maybe this is where we need the `old -> new` log. In that case, I am not sure what the previous log adds.

emission/tests/storageTests/TestModelStorage.py
```diff
@@ -0,0 +1,138 @@
...
+        self.total_trips = 100
+        self.clustered_trips = 33    # bins must have at least self.min_trips similar trips by default
+        self.has_label_percent = 0.9 # let's make a few that don't have a label, but invariant
+                                # $clustered_trips * $has_label_percent > self.min_trips
+                                # must be correct or else this test could fail under some random test cases.
+
+        # for a negative test, below
+        self.unused_user_id = 'asdjfkl;asdfjkl;asd08234ur13fi4jhf2103mkl'
+
+        # test data can be saved between test invocations, check if data exists before generating
```
shankari, line 53: this should not be true if we tear down properly. what was the set of steps that led to the data being saved between test invocations?

emission/tests/storageTests/TestModelStorage.py
```diff
@@ -0,0 +1,138 @@
...
+                has_label_p=self.has_label_percent
+            )
+
+            ts.bulk_insert(train)
+
+            # confirm data write did not fail
+            test_data = esda.get_entries(key="analysis/confirmed_trip", user_id=user_id, time_query=None)
+            if len(test_data) != self.total_trips:
+                logging.debug(f'test invariant failed after generating test data')
+                self.fail()
```
shankari, line 87: when will you run into this situation? If this can happen frequently and we fail instead of retrying when it happens, we may have a non-deterministic test, which is not a great idea for an automated test pipeline.

emission/tests/storageTests/TestModelStorage.py
```diff
@@ -0,0 +1,138 @@
...
+        greedy_model_config = {
+            "metric": "od_similarity",
+            "similarity_threshold_meters": 500,
+            "apply_cutoff": False,
+            "clustering_way": 'origin-destination',
+            "incremental_evaluation": False
+        }
+        maximum_stored_model_count = eamtc.get_maximum_stored_model_count()
+        logging.debug(f'(TRAIN) creating a model based on trips in database')
+        for i in range(20):
```
shankari, line 120: why is this hardcoded to 20?

shankari (changes requested):
Looks good at a high level; but I would like some polishing changes...

emission/storage/modifiable/builtin_model_storage.py
MukuFlash03: Yes, it makes more sense to add the change in model count here.
I'll explain the need for the logging statement present at the beginning in `upsert_model()`.
To clarify the three different states where the logging statements are present, say we have 3 models currently.

Two sets of changes to model count are there: a) trimming models b) inserting new model

1. Current count before **trimming** models = 3
```
# In trim_model_entries()
current_model_count = edb.get_model_db().count_documents({"user_id": self.user_id})
logging.debug("Before trimming, model count for user %s = %s" % (self.user_id, current_model_count))
```

2. New count after **trimming** models = 2
```
# In trim_model_entries()
new_model_count = edb.get_model_db().count_documents({"user_id": self.user_id})
logging.debug("After trimming, model count for user %s = %s" % (self.user_id, new_model_count))
```

3. New count after **inserting** models = 3
```
# In upsert_model()
new_model_count = edb.get_model_db().count_documents({"user_id": self.user_id})
logging.debug("New model count for user %s = %s" % (self.user_id, new_model_count))
```

emission/tests/storageTests/TestModelStorage.py
MukuFlash03: So, I've taken this code from an existing test file - [TestRunGreedyModel.py](https://github.com/e-mission/e-mission-server/blob/master/emission/tests/modellingTests/TestRunGreedyModel.py), as I wasn't really sure how to generate models, trips and didn't want to re-write the code and miss any important parameters, variables needed for training, inserting models.

But I'll make changes as per my understanding.

emission/tests/storageTests/TestModelStorage.py
MukuFlash03: Answered [here](https://github.com/e-mission/e-mission-server/pull/948#discussion_r1478739529).
But I'll make changes as per my understanding.

emission/tests/storageTests/TestModelStorage.py
shankari: Right, but I would have expected to see something like `maximum_stored_model_count + 2` to be more flexible. I will accept this for now since this is part of the unit tests and the sample value is not overwritten here.

emission/tests/storageTests/TestModelStorage.py
shankari: It is fine to keep it. We likely have not run into this error before, but it is not bad to have that check, as long as it is well documented.

emission/analysis/modelling/trip_model/config.py
```diff
@@ -75,5 +75,9 @@ def get_minimum_trips():
         raise TypeError(msg)
     return minimum_trips
 
-
-
+def get_maximum_stored_model_count():
```
shankari, line 78: I'm not going to hold up the merge for this release, but would be good to avoid these extraneous whitespace changes in future PRs

emission/storage/modifiable/builtin_model_storage.py
```diff
@@ -6,7 +6,7 @@
 
 import emission.core.get_database as edb
 import emission.storage.modifiable.abstract_model_storage as esma
-
+import emission.analysis.modelling.trip_model.config as eamtc
```
shankari, line 9: like this

shankari (changes requested):
Where in here are you checking that the models that are deleted are the oldest ones? I see that the code does that with the sort on the `metadata.write_ts` but I don't see a test for that. If I changed 

```
edb.get_model_db().find(find_query).sort("metadata.write_ts", -1)
```

to 

```
edb.get_model_db().find(find_query).sort("metadata.write_ts", 1)
```

I believe the test would still pass

emission/tests/storageTests/TestModelStorage.py
MukuFlash03: Oh, understood, hardcoding was probably not the best idea. Changed this.

emission/analysis/modelling/trip_model/config.py
MukuFlash03: Oh, right. It's just that sometimes there are extra lines at the end of the file and when I'm adding new code I tend to keep the spacing between modules similar to how it is in other parts of the file. But will try to minimize these changes.

MukuFlash03:
> Where in here are you checking that the models that are deleted are the oldest ones? I see that the code does that with the sort on the `metadata.write_ts` but I don't see a test for that. If I changed
> 
> ```
> edb.get_model_db().find(find_query).sort("metadata.write_ts", -1)
> ```
> 
> to
> 
> ```
> edb.get_model_db().find(find_query).sort("metadata.write_ts", 1)
> ```
> 
> I believe the test would still pass

Thank you for pointing that out; that is indeed an important test I missed.

For testing purposes, I did change the sort(-1) to sort(1) which leads to ascending order of queried models based on write_ts.
I had the following observations :
- Initial count assertion itself fails (which is good!) as all available models satisfy `write_ts` condition of `$ lte < write_ts_limit` .
```
            filter_clause = {
                "user_id" : self.user_id,
                "metadata.key" : key,
                "metadata.write_ts" : { "$lte" : write_ts_limit }
            }
```

- Hence every time all models are being deleted and only one model remains (current_model_count = 1) which is the latest one being inserted.
```
if i <= (maximum_stored_model_count - 1):
    self.assertEqual(current_model_count, i+1)
else:
    self.assertEqual(current_model_count, maximum_stored_model_count)
```

- Just to get beyond this point, I tried testing without this assertion, to observe `write_ts_lists`:
    - For every `(maximum_stored_model_count + 1)th` model added, the first `maximum_stored_model_count ` models are deleted and the latest model is inserted and this repeats.

--------- 

So, to reinforce the testing, I’m adding another assertion to confirm that only the latest top K or max_model_count models are stored by storing the write_ts times in two different lists and checking for equality of the top K entries:
```
- model_creation_write_ts_list : stores write_ts times each time a model is created in the for loop.
- stored_model_write_ts_list : stores write_ts times of all the already stored models in the DB, which should just have the latest models.
- The last 'maximum_stored_model_count' in model_creation_write_ts_list should match those in stored_model_write_ts_list.
```

shankari:
I do not 100% understand how the first assertion tripped, but I am satisfied with the changes and am happy to merge!

shankari (approved):

PR #949, opened 2023-12-16
Fixing latest Docker image vulnerabilities

MukuFlash03:
AWS Inspector found the following vulnerable packages:

CRITICAL
perl

HIGH
nghttp2, libnghttp2-14
cryptography, libssl3
cryptography
libc6, libc-bin

Upgraded perl, libssl3, nghttp2 packages by upgrading base Ubuntu image to latest of the same LTS version - jammy (22.04).

Cryptography package was fixed by mentioning required version to be installed using conda.

Libc6, Libc-bin can be fixed by using apt-get upgrade but this upgrades all packages which is not recommended as a blanket upgrade fix. Possible that a newer Ubuntu version will have fixes for this, as the same issue was encountered last time around.

### DIFF
--- a/.docker/setup_config.sh
+++ b/.docker/setup_config.sh
@@ -13,7 +13,7 @@ source setup/setup.sh
 ## 10/02 - Mukul
 ## - Above comments talk about manually updating cryptography to version 40
 ## - I have upgraded to 41.0.4 as per latest vulnerability fixes.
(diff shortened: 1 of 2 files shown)


shankari:
@MukuFlash03 this is now merged; you should be able to bump up the server dependencies now

PR #824, opened 2021-06-25
export_data pipeline stage

jruzekowicz:
Export data pipeline stage implementation. Designed to export data in the desired format.
[export_a6718c68-989e-4d22-96f7-ef16d4bcfd6c.gz](https://github.com/e-mission/e-mission-server/files/6716360/export_a6718c68-989e-4d22-96f7-ef16d4bcfd6c.gz)

Testing: I loaded the July 22, 2015 testing data and ran the intake
pipeline. I expected to have the .gz file be outputted. After I recieved
the file, I was able to confirm that the file outputted the correct data
by cross-comparing it to the raw data, as well as running the commpand
within the notebook and printing out the expected results.
Usage: The export data stage of the pipeline is an elementary version of
a stage that can be expanded for the desired output format of any
entity. ie, any person, organization, or otherwise can export data in
their desired format by changing the keys we look for.

Attached is the output file from the export_data stage of the pipeline.

PR #952, opened 2024-01-08
Purge + Restore user timeseries data with long-term storage

MukuFlash03:
Building on top of PR #904 with the objective of integrating this into the current codebase.
Will be analyzing the file storage options (CSV, JSON) as well as including the option for mongodump/restore.
Additionally long-term storage to AWS will be looked at as well.


MukuFlash03:
Thinking of options to be made available for type of export file - CSV, JSON, Mongodump
- CSV -> Complicates restoring due to: NaN field columns, loss of data structure 
- JSON -> Works good with both export and import
- Mongodump -> Should be same as JSON

I’m proposing default option to be JSON which will occur irrespective of other type of export (CSV for now).
Since currently JSON export/import has been implemented.
But can change it to making JSON default only when CSV is chosen.
Since once Mongodump is implemented and even that data can be imported, then even Mongodump becomes an option for default.
But for systems without mongoDB and avoid complexities of mongodump, can keep JSON default as it can be viewed in a text editor compared to Mongodump.
This will mean JSON file will always be generated so will get either just JSON or a combination of JSON and other file types (JSON + CSV or JSON + CSV + Mongodump).

Currently integrated CSV export as an optional export.
I have grouped all export data file types in a boolean dictionary which can the call specific export function is that flag is set.

Next working on adding Mongodump as an option.

MukuFlash03:
Latest fixes work with JSON files used as the export file type and this same file works for importing data correctly back into the database as well.

However, with CSV files, it's a bit complicated because of two reasons: 
1) nested data structure is lost 2) data is stored as strings which converts types like arrays into strings too.

Also, with Mongodump, I tried including the python subprocess module to use it to run the mongodump command as a terminal command but there were issues with dependencies missing, then had to figure out how to run and/or save the file locally in the Docker container and then extract it out of there if needed for analysis on one's local system.

With these in mind, JSON currently seems a good option for basic export/import.
I will check out the already existing code for JSON export too.

shankari:
Yes, that is the export PR. We can also potentially use the associated pipeline state for incremental exports since this is planned to be part of a running system. I would suggest the use of command line flags to indicate whether we want an incremental dump or a full dump.

shankari:
As discussed earlier, this needs to use the existing export method.

shankari:
The scalability considerations here, if you follow the PR chain, are that we store the raw data in the timeseries, we generate the analysis timeseries from the raw timeseries, and then we use the analysis timeseries (aka with the analysis results) in the rest of our work. So the argument is that you don't need to store the raw timeseries on an ongoing basis. It is only required when we need to reset the pipeline and/or change analysis algorithms. So it can be moved to long-term storage, but with the ability to restore before we reset the pipeline.

MukuFlash03:
I started off by first understanding the differences in terms of code changes from the [newer Purge PR](https://github.com/e-mission/e-mission-server/pull/904) and the [older Export PR](https://github.com/e-mission/e-mission-server/pull/824). 

--------

A) Purge vs Export 
1) Purge PR does purge or delete entries once export is completed.
2) Export PR only exports timeseries entries to JSON dump without any deletion from the database. I did see from [this comment](https://github.com/e-mission/e-mission-server/pull/899#issuecomment-1430780732) that `bin/debug` has a script for purging as well. But this script purges all the entries for a specific user from all databases, which is not what we want.

-----

B) Do we want to allow the user to specify the time range for export + purge ? Or should it be fetched programmatically using pipeline stages like it's being done in the Purge PR and in the `emission.pipeline.export_stage`? 

1) Purge PR script does not need time range for extraction to be specified in command line inputs; fetched programmatically using last_run_ts. The correct time variable to be used is last_processed_ts however (refer to [this comment)](https://github.com/e-mission/e-mission-server/pull/899#:~:text=The%20last_processed_ts%20is%20the%20one%20used%20for%20actual%20pipeline%20operations%2C%20including%20resetting).

2) Export PR uses the [`extract_timeline_for_day_range_and_user.py`](https://github.com/e-mission/e-mission-server/blob/master/bin/debug/extract_timeline_for_day_range_and_user.py) script to use the export functionality via command-line but it needs start and end dates to be passed as arguments.
I do also see another way to execute the export functionality would be to run the export pipeline similar to how we run the intake pipeline. 
The export pipeline handles fetching the time range based on the pipeline state using:
```
start_ts = curr_state.last_processed_ts
end_ts = time.time() - END_FUZZ_AVOID_LTE
```

---------

C) Output files containing exported data differ in the format of the ObjectID and UUID

1) Purge PR script output format matches the entries in the database - ObjectID and UUID
``` 
        ...
	'_id': ObjectId('657b98c36078c82e567e462a'), 
	'user_id': UUID('2a4acad8-ee98-45ff-b4b5-322345700646'),
        ...
```

2) Export PR script output format is changed to $oid and $uuid. 
``` 
        ...
        "_id": {
            "$oid": "657b98c36078c82e567e462a"
        },
        "user_id": {
            "$uuid": "2a4acad8ee9845ffb4b5322345700646"
        }, ...
```

This is managed via `emission.storage.json_wrappers` module that uses `bson.json_util module`.
Handling this will be required when restoring the data but the conversion should be fairly simple or might even be handled automatically by MongoDB when inserting into the database.

--------

MukuFlash03:
Looking at the files involved in the [export PR](https://github.com/e-mission/e-mission-server/pull/824):

- `bin/debug/extract_timeline_for_day_range_and_user.py`: Command-line script to execute export()
- `emission/export/export.py`: Main nested export file that exports to json dump.
-` emission/core/wrapper/pipelinestate.py`: Declares a new Pipeline Stage.
- `emission/exportdata/export_data.py`: Fetches time range for export and calls `emission.export.export()`
- `emission/pipeline/export_stage.py`: Defines a new export Pipeline that calls `emission.exportdata.export_data()`
- `emission/tests/exportTests/TestExportModule.py`: Unit tests for export()

Export Data Flow: 
`emission/pipeline/export_stage.py` -> `emission/exportdata/export_data.py` -> `emission/export/export.py`

-----------

1. TestExportModule.py
I started with trying to execute this test to get a sense of the flow of data from setting up a sample user to exporting the data. 

But I got some errors:

A. FileNotFoundError

```
FileNotFoundError: [Errno 2] No such file or directory: 'emission/archived/archive_28712d00-ca16-4fae-9fc2-032a2b4bd5ea_1437607737.997_1720561247.677196.gz'

Erroneous Code:
        file_name = os.environ.get('DATA_DIR', 'emission/archived') + "/archive_%s_%s_%s" % (self.testUUID, time_query.startTs, time_query.endTs)
```

To solve this issue, I had to manually create the archive directory under emission, and then execute the test file.

------

2. Assertion Error
Fails this assertion in the code:

```
        #Testing the matching total number of confirmed trips, testing no confirmed trips in raw.
        #Testing also for the first three trips that the object ids and user ids are consistent.
        self.assertEqual(len(confirmed_trips_exported), len(confirmed_trips_db))
```

Error stack trace
```
FAIL: testExportModule (__main__.TestExportModule)
  File "/Users/mmahadik/Documents/Work/OpenPATH/Code/GitHub/e-mission-server/emission/tests/exportTests/TestExportModule.py", line 51, in testExportModule
    self.assertEqual(len(confirmed_trips_exported), len(confirmed_trips_db))
AssertionError: 0 != 6
```

I'm yet to ascertain the cause of this.

------

MukuFlash03:
### Main query for now

One of the main things that I am concerned with is that, the existing export() method takes in data from three sources: `timeseries_db, analysis_timeseries_db, usercache`.

But the requirement in this PR is to export the raw data entries stored in timeseries_db only and then go on and purge the exported data entries and restore from the exported data when required. 

----

This snippet is from `emission/export/export.py`:

```
def get_from_all_three_sources_with_retry(user_id, in_query):
    import emission.storage.timeseries.builtin_timeseries as estb

    ts = estb.BuiltinTimeSeries(user_id)
    uc = enua.UserCache.getUserCache(user_id)

    sort_key = ts._get_sort_key(in_query)
    base_ts_call = lambda tq: ts._get_entries_for_timeseries(ts.timeseries_db, None, tq,
        geo_query=None, extra_query_list=None, sort_key = sort_key)
    analysis_ts_call = lambda tq: ts._get_entries_for_timeseries(ts.analysis_timeseries_db, None, tq,
        geo_query=None, extra_query_list=None, sort_key = sort_key)
    uc_ts_call = lambda tq: (uc.getMessageCount(None, tq), uc.getMessage(None, tq))
```

------


Reasoning behind these changes to read from the three database sources is mentioned in [this issue](https://github.com/e-mission/e-mission-docs/issues/845#:~:text=The%20only%20principled%20solution%20is%20to%20actually%20read%20the%20three%20types%20of%20entries%20directly%20from%20the%20database%20and%20to%20retry%20each%20of%20them%20separately.)

[This PR](https://github.com/e-mission/e-mission-server/pull/896) added these changes to fix [this issue](https://github.com/e-mission/e-mission-docs/issues/845).

[This commit](https://github.com/e-mission/e-mission-server/pull/896/commits/568717baec7df50eae9fa99cb115ac45f4bf2c8b) has the specific changes that were added.

[This commit](https://github.com/e-mission/e-mission-server/pull/896/commits/4c6d1980519787f8f748eaabc7dc2f97323be03e) talks about restoring dumped data.

[This comment](https://github.com/e-mission/e-mission-docs/issues/845#issuecomment-1386020225) has logs about restoring user data.

------

I am going to scan these related issues again to see if there's a workaround or we'll need a separate export script altogether. 

------

MukuFlash03:
I'm somewhat unclear about the usage of last_run_ts, last_processed_ts, curr_run_ts, start_ts, end_ts in the pipeline code. 
Have found these references in the export PR, my understanding is better but need to grasp the flow entirely:

Detailed discussion here in the [export PR](https://github.com/e-mission/e-mission-server/pull/824):
- Full vs Incremental export [in this comment](https://github.com/e-mission/e-mission-server/pull/824#:~:text=The%20default%20export_stage,streamed%20on%2Ddemand)
- last_processed_ts mentioned [here](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-870708489)
- start_ts, end_ts mentioned [here](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-871760577)

Will need to read up on the pipeline code, pipeline states again.

shankari:
> B) Do we want to allow the user to specify the time range for export + purge ? Or should it be fetched programmatically using pipeline stages like it's being done in the Purge PR and in the emission.pipeline.export_stage?

Programatically. I don't see any arguments for not. In general, the "purge" PR is what we should base our work on since it is by far the most polished option.

> I am going to scan these related issues again to see if there's a workaround or we'll need a separate export script altogether.

I don't see why you need a separate export script. It should be trivial to add a database argument to the existing script. We can also consider exporting using `mongodump` instead of to csv or json!

MukuFlash03:
So, I've made some progress and have been able to implement the overall data flow for Export -> Purge -> Restore in the [TestExportPurgeRestoreTimeseries.py file](https://github.com/e-mission/e-mission-server/blob/6a990cbb9317110635a5ee5ce856b95644ed565a/emission/tests/binTests/TestExportPurgeRestoreUserTimeseries.py)

As suggested, I am now passing a database argument to the export script.
For now, I haven't added any assertion tests and I just wanted to test if it was working correctly. 

Also, there's still some issues I've found out relating to the exported data which is causing the data restore to fail as well.

---------

Two main issues (I'll discuss them in detail in below comments):
1. Extra analysis entries being exported in the [export.py](https://github.com/e-mission/e-mission-server/blob/6a990cbb9317110635a5ee5ce856b95644ed565a/emission/export/export.py) script.
2. Design considerations related to these files: export_stage pipeline, purge_user_timeseries

MukuFlash03:
### 1. Extra analysis entries being exported in the [export.py](https://github.com/e-mission/e-mission-server/blob/6a990cbb9317110635a5ee5ce856b95644ed565a/emission/export/export.py) script.

For the testing data mentioned below, I've used this sample input data file from the test data:
`emission/tests/data/real_examples/shankari_2015-07-22`

Initially, on passing the database as an argument ([code snippet](https://github.com/e-mission/e-mission-server/blob/6a990cbb9317110635a5ee5ce856b95644ed565a/emission/export/export.py#L42C1-L58C1)) to` export()`, I expected only the timeseries data to be exported.

However, I saw that the exported data contained three analysis entries as well with these three keys: 
`segmentation/raw_place, analysis/cleaned_place, analysis/confirmed_place`

----

Taking a look at the export code, I saw that data is fetched using a combination of:
- Four different queries (`loc_time_query, trip_time_query, place_time_query, first_place_extra_query`) 
- From three data sources (`timeseries, analysis, usercache`)

I tested the export function by checking for each database source individually, and observed that these three analysis entries were being fetched for each of these, irrespective of the specified database.

------

That made me realized that's somewhere in the export code itself that it's fetching analysis entries.

I found that the fourth query - [first_place_extra_query](https://github.com/e-mission/e-mission-server/blob/master/emission/export/export.py#L71-L75), is uses `find_entries()` , which is the reason for those analysis entries.

```
    # Handle the case of the first place, which has no enter_ts and won't be
    # matched by the default query
    first_place_extra_query = {'$and': [{'data.enter_ts': {'$exists': False}},
                                        {'data.exit_ts': {'$exists': True}}]}
    first_place_entry_list = list(ts.find_entries(key_list=None, time_query=None, extra_query_list=[first_place_extra_query]))
    logging.info("First place entry list = %s" % first_place_entry_list)
```

[find_entries()](https://github.com/e-mission/e-mission-server/blob/master/emission/storage/timeseries/builtin_timeseries.py#L223) combines the results from the two databases:
```
        return itertools.chain(orig_ts_db_result, analysis_ts_db_result)
```

Finally, [export.py](https://github.com/e-mission/e-mission-server/blob/master/emission/export/export.py#L78-L92) returns this as the combined list of data to be exported:

```
combined_list = loc_entry_list + trip_entry_list + place_entry_list + first_place_entry_list
....
....
        combined_filename = "%s.gz" % (file_name)
        with gzip.open(combined_filename, "wt") as gcfd:
            json.dump(combined_list,
                gcfd, default=esj.wrapped_default, allow_nan=False, indent=4)
```

MukuFlash03:
### So, what are the concerns with this?

This leads to some problems with the both the **data being purged** and the **data being restored**.

For purging, we aren't then only merging the raw data in the timeseries but also some analysis data.
So, let's say we only delete the timeseries entries and keep these analysis entries as it is.

But then, for restoring, since the export data file contains these extra analysis entries, and we have not deleted this analysis entries, the insert operation fails due to these entries already existing in the analysis database.

Note: For restoring data, I'm making use of the existing script: `load_multi_timeline_for_range.py`
I had make some modifications to allow the script to be invoked as a function from other files and not just via the command line.


Note: Perhaps this is related
I found this function being used in the `load_multi_timeline_for_range` script: `analyse_timeline()` present in [bin/debug/common.py](https://github.com/e-mission/e-mission-server/blob/master/bin/debug/common.py#L27-L32) checks if raw data has analysis keys or not an then sets / resets a boolean variable `needs_rerun`.

------

Distribution of entries in the exported data fetched from the timeseries database (and three entries from analysis db)

This calculation is from [export.py()](https://github.com/e-mission/e-mission-server/blob/master/emission/export/export.py#L78-L92)
```
combined_list]() = loc_entry_list + trip_entry_list + place_entry_list + first_place_entry_list
```

- combined_list = 650
- loc_entry_list = 647
- trip_entry_list = 0
- place_entry_list = 0
- first_place_entry_list = 3

loc_entry_list keys (total = 647)
- background/location = 5
- background/filtered_location = 7
- background/motion_activity = 625
- statemachine/transition = 10

MukuFlash03:
### 2. What exactly is raw data? 

**Does it include only `background` keys or others as well ?**

As seen in the exported data distribution in this comment, there are `background/*` keys as well as `statemachine/transition` keys.

Additionally, I also tested the export function with some staging data that I had.
- I got entries with `data.field = USERCACHE` as well. 
- These were fetched from the timeseries.db using the loc_time_query in [export.py()](https://github.com/e-mission/e-mission-server/blob/master/emission/export/export.py#L63-L76) and all had the same key - `stats/pipeline_time`.
- None of these had the background keys.

-------

### 3. Design considerations related to these files: export_stage pipeline, purge_user_timeseries

Trying to outline what is the **better way to utilize the export function**.

First approach is to **directly use** the export() script in `export.py`, which is the way I'm doing it currently in the latest set of code changes.

Second approach is to run the export pipeline, similar to the way we run the import pipeline.
- A point against using this, is that we'll need to pass the some parameters that are mainly required by the export() but will have to cascade them down from `export_stage.py -> export_data.py -> export.py `
- These parameters are specific in the purge script such as: `export_file_name` (specified by user), `database source` (specified as `timeseries_db`)

MukuFlash03:
### 4. Time range for export and purge

**Do I need to figure out how to extract the time range?**

In both the below approaches, the time is being fetched programmatically. 
However, we need a time-range and just a single time, in which case, the implementation in the export PR is more suitable.

--------

Right now, I've utilized the same approach to get the time range for export as being done in [export_data.py](url).

```
ts = esta.TimeSeries.get_time_series(user_id)
time_query = espq.get_time_range_for_export_data(user_id)
```

In the original [Purge PR](https://github.com/e-mission/e-mission-server/pull/904/files), this is how they're using the `last_ts_run` only. This doesn't use a time-range.

```
cstate = esp.get_current_state(user_id, ecwp.PipelineStages.CREATE_CONFIRMED_OBJECTS)
last_ts_run = cstate['last_ts_run']
```

shankari:
@MukuFlash03 the expectation here is not that the export script will be a black box solution to your needs. The expectation is that you will adapt it to meet the needs here. The original export script was written as a way for an external system that wanted to work with the data to get a frequently updated copy. This would include both analysed and raw data. To understand the difference between analysed and raw data, and the e-mission data model in general, I would suggest that you read chapter 6 of my thesis.

The export script is our code. The new purge script will also be our code. We don't need to treat it as a black box, we only need to make the changes to make it work for both use cases.

The approach with the pipeline state tracking the current state of the export/purge is what we need to retain across use cases.

MukuFlash03:
I am in the process of basing the purge script on the existing export script.
To also factor in maintaining the current pipeline state, I also thought it would be a good idea to start with replicating not just the export.py script but both export_data.py and export_stage.py and thereby introducing the pipeline code with pipeline states as well.

Now, assuming the pipeline code is a good approach for our purge usecase, there's also the consideration that whether inside of these we should have two different states for purge - 1) export data 2) delete data.
If yes, then similar replication would be needed for the restore functionality too - 1) import data 2) insert data.

Points for having separate states:
- Helps keep the states isolated, in case we end up extending the functionality and having more states later on such as some sort of verification after export and deleting entries.

Points against having separate states:
- The export and delete operations aren't that complicated or big that we'd need to separate states.
Likewise for restore process involving import / insert.

-----

Another point just came to my mind.
I've got the json export / import to work but as suggested I'd be using mongodump / mongorestore.

While for purging we'd still have two steps (export and delete), with mongorestore, restore process would need a single step and not import and insert both.


So for now, I do feel just using different functions instead of different pipeline states entirely should work.

MukuFlash03:
I was also thinking of ensuring that any errors in exporting or deleting data (likewise for restoring) are handled well. Especially, we should be able to know how much data has been exported / deleted and the purge process can continue from there.

I did find some discussion on this in the export PR([comment 1](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-871760577), [comment 2](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-870708489)) specifically related to the `last_processed_ts` but I am unsure if this was implemented ?
The reason I doubt this is because I re-ran the export script more than once but it still kept copying all the entries for the user. I did use an email ID while running the TestExportModule.py to ensure it uses the same UUID across multiple executions of the script.

I expected that if the last_processed_ts part had been implemented for the export stage, then this shouldn't have happened. It should have exported all the entries once and then not exported any more entries since all data has been exported already.

I'm not sure if this how the export process was designed to work or I'm understanding it wrong.
But if it was designed like this, then why is it exporting all entries?

If my understanding is incorrect, consider the case say when the exported files are lost and users want to re-export their data. I feel there should be a way to let the user export data even after all data has been exported already; but the export process won't do this. 

But, I feel the pipeline reset is helpful in this case? This would reset the states and data can be exported again.

MukuFlash03:
### Incorrect Pipeline State used in Export ?

Also, while inspecting the logs on running the export script, I found a possible issue.

This is how it looks 

```
INFO:root:Found 647 loc-like entries, 0 trip-like entries, 0 place-like entries = 647 total entries
INFO:root:timeline has unique keys = {'background/location', 'background/filtered_location', 'background/motion_activity', 'statemachine/transition'}
INFO:root:Combined list:
INFO:root:647
DEBUG:root:After run, last_trip_done == None, must be early return
INFO:root:For stage PipelineStages.EXPECTATION_POPULATION, last_ts_processed is unchanged
DEBUG:root:About to save object PipelineState({'_id': ObjectId('66a180bbf6762d6ceacd1421'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 15, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860279.315834})
DEBUG:root:After saving state PipelineState({'_id': ObjectId('66a180bbf6762d6ceacd1421'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 15, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860279.315834}), list is [{'_id': ObjectId('66a180b4f6762d6ceacd1200'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 12, 'curr_run_ts': None, 'last_processed_ts': None, 'last_ts_run': 1721860271.571258}, {'_id': ObjectId('66a180b4f6762d6ceacd1201'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 6, 'curr_run_ts': None, 'last_processed_ts': None, 'last_ts_run': 1721860271.578481}, {'_id': ObjectId('66a180b4f6762d6ceacd1202'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 1, 'curr_run_ts': None, 'last_processed_ts': 1437612893.937, 'last_ts_run': 1721860271.5822809}, {'_id': ObjectId('66a180b8f6762d6ceacd1210'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 2, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860275.5275788}, {'_id': ObjectId('66a180b9f6762d6ceacd1233'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 3, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860276.162166}, {'_id': ObjectId('66a180b9f6762d6ceacd1248'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 11, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860276.6067011}, {'_id': ObjectId('66a180bbf6762d6ceacd140c'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 4, 'curr_run_ts': None, 'last_processed_ts': None, 'last_ts_run': None}, {'_id': ObjectId('66a180bbf6762d6ceacd140e'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 14, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860278.144404}, {'_id': ObjectId('66a180bbf6762d6ceacd1421'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 15, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860279.315834}, {'_id': ObjectId('66a180bbf6762d6ceacd1428'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 13, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860278.193198}, {'_id': ObjectId('66a180bbf6762d6ceacd1436'), 'user_id': UUID('58852b1f-364c-4b7e-a1ec-78dfbd4dbbe3'), 'pipeline_stage': 17, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721860278.7693272}]

```


This is where I added logs :
```
DEBUG:root:Inside export_data.export_data: if last_trip_done None
DEBUG:root:After run, last_trip_done == None, must be early return
DEBUG:root:Inside export_data.export_data: Before mark_export_data_done
DEBUG:root:Inside pipeline_queries.mark_export_data_done: Before mark_stage_done
DEBUG:root:Inside pipeline_queries.mark_stage_done: Before if last_processed_ts - PipelineStages.EXPECTATION_POPULATION
DEBUG:root:Inside pipeline_queries.mark_stage_done: inside else before - PipelineStages.EXPECTATION_POPULATION
INFO:root:For stage PipelineStages.EXPECTATION_POPULATION, last_ts_processed is unchanged
DEBUG:root:Inside pipeline_queries.mark_stage_done: inside else after - PipelineStages.EXPECTATION_POPULATION
DEBUG:root:Inside pipeline_queries.mark_stage_done: before save object - PipelineStages.EXPECTATION_POPULATION
DEBUG:root:About to save object PipelineState({'_id': ObjectId('66a18c8229c89ba695f3538d'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 15, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863294.755502})
DEBUG:root:After saving state PipelineState({'_id': ObjectId('66a18c8229c89ba695f3538d'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 15, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863294.755502}), list is [{'_id': ObjectId('66a18c7b29c89ba695f3516c'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 12, 'curr_run_ts': None, 'last_processed_ts': None, 'last_ts_run': 1721863286.5213718}, {'_id': ObjectId('66a18c7b29c89ba695f3516d'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 6, 'curr_run_ts': None, 'last_processed_ts': None, 'last_ts_run': 1721863286.528563}, {'_id': ObjectId('66a18c7b29c89ba695f3516e'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 1, 'curr_run_ts': None, 'last_processed_ts': 1437612893.937, 'last_ts_run': 1721863286.532139}, {'_id': ObjectId('66a18c7f29c89ba695f3517c'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 2, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863290.6022348}, {'_id': ObjectId('66a18c8029c89ba695f3519f'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 3, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863291.613833}, {'_id': ObjectId('66a18c8129c89ba695f351b4'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 11, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863292.032438}, {'_id': ObjectId('66a18c8229c89ba695f35378'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 4, 'curr_run_ts': None, 'last_processed_ts': None, 'last_ts_run': None}, {'_id': ObjectId('66a18c8229c89ba695f3537a'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 14, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863293.4024498}, {'_id': ObjectId('66a18c8229c89ba695f3538d'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 15, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863294.755502}, {'_id': ObjectId('66a18c8229c89ba695f35394'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 13, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863293.443583}, {'_id': ObjectId('66a18c8329c89ba695f353a2'), 'user_id': UUID('f8853d31-6f2e-48c7-873c-6683266a7607'), 'pipeline_stage': 17, 'curr_run_ts': None, 'last_processed_ts': 1437607737.997, 'last_ts_run': 1721863294.2245262}]
DEBUG:root:Inside pipeline_queries.mark_stage_done: after save object - PipelineStages.EXPECTATION_POPULATION
DEBUG:root:Inside pipeline_queries.mark_export_data_done: After mark_stage_done
DEBUG:root:Inside export_data.export_data: After mark_export_data_done
```

At first I thought the incorrect pipeline state was being set in [pipeline_queries.py](https://github.com/e-mission/e-mission-server/blob/master/emission/storage/pipeline_queries.py#L225-L233) in `mark_export_data_done`:
```
if last_trip_done is None:
        mark_stage_done(user_id, ps.PipelineStages.EXPORT_DATA, None)
    else:
        mark_stage_done(user_id, ps.PipelineStages.EXPORT_DATA,
                        last_trip_done.data.end_ts + END_FUZZ_AVOID_LTE)
```

But this is correct.

-----

I then checked the `PipelineStages` class in [emission/core/wrapper/pipelinestate.py](https://github.com/e-mission/e-mission-server/blob/master/emission/core/wrapper/pipelinestate.py#L12-L31) and found two identical states:
```
class PipelineStages(enum.Enum):
    EXPECTATION_POPULATION = 15
    EXPORT_DATA = 15
```

So, is this messing the export pipeline process and why it always exports all entries?

MukuFlash03:
I went ahead and assigned EXPORT_DATA a new unique pipeline state.

I now see that the count of entries exported in my purge script focussing only on timeseries_db entries now matches the count of exported entries in the export script.
I'm using this sample data: emission/tests/data/real_examples/shankari_2015-07-22 ([link](https://github.com/e-mission/e-mission-server/blob/master/emission/tests/data/real_examples/shankari_2015-07-22))

For the export script:
Without making the change, I was getting 647 timeseries entries.
With the change, I now have 1906 timeseries entries.

This also matches the number of unique entries in the raw data in the sample data file.

MukuFlash03:
I did test running the export multiple times but it still exports all the entries.

Is this because last_processed_ts is not used in the [export.py](https://github.com/e-mission/e-mission-server/blob/master/emission/exportdata/export_data.py#L14-L21) file anywhere?

```
def export_data(user_id):
    try:
        edp = ExportDataPipeline()
        edp.user_id = user_id
        edp.run_export_data_pipeline(user_id)
        if edp.last_trip_done is None:
            logging.debug("After run, last_trip_done == None, must be early return")
        espq.mark_export_data_done(user_id, edp.last_trip_done)
```

I do see that `last_trip_done` is used instead. 
I also saw that the pipeline queries and how some stages use the `last_processed_ts` to mark the stage done while some use `last_trip_done` or `last_section_done`.

MukuFlash03:
Implemented logic to handle last_processed_ts correctly for the purge function for now. Working on restore function next.

Initially I had based my code on the export script, using it as a template however it does not have the logic for updating the last_trip_done (alternatively used in some stages instead of last_processed_ts).  Hence, I observed that on running the purge / restore script multiple times, data was being exported / imported multiple times despite there not being any new data.
This is highlighted in [this comment](https://github.com/e-mission/e-mission-server/pull/952#issuecomment-2249130047).

-----

To fix this, I went through the pipeline code and took a note of the different stages and how they were specifically using a variable (`last_processed_ts` or `last_trip_done` or `last_section_done`) to update the latest trip being processed. 
This was either:
- done iteratively for each trip entry or,
- just being calculated at the end of the operation by directly fetching the last indexed entry.

MukuFlash03:
Here’s an overview of the pipeline stages along with the approaches used for updating the last_processed_ts.

|S. No. | Pipeline Stage | Function | Filename | Variable used | Approach
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 1. | USERCACHE | uh.moveToLongTerm() | emission/net/usercache/builtin_usercache_handler.py | last_processed_ts | metadata.write_ts used for each trip iteratively |
| 2. | USER_INPUT_MATCH_INCOMING | eaum.match_incoming_user_inputs() | emission/analysis/userinput/matcher.py | last_processed_ts | metadata.write_ts used for each trip iteratively |
| 3. | ACCURACY_FILTERING | eaicf.filter_accuracy() | emission/analysis/intake/cleaning/filter_accuracy.py | last_processed_ts | metadata.write_ts used by indexing last element |
| 4. | TRIP_SEGMENTATION | eaist.segment_current_trips() | emission/analysis/intake/segmentation/trip_segmentation.py | last_processed_ts | filter_methods used |
| 5. | SECTION_SEGMENTATION | eaiss.segment_current_sections() | emission/analysis/intake/segmentation/section_segmentation.py | last_trip_done | last trip fetched directly by indexing and data.end_ts used |
| 6. | JUMP_SMOOTHING | eaicl.filter_current_sections() | emission/analysis/intake/cleaning/location_smoothing.py | last_section_done | last section fetched directly by indexing and data.end_ts used |
| 7. | CLEAN_RESAMPLING | eaicr.clean_and_resample() | emission/analysis/intake/cleaning/clean_and_resample.py | last_section_done | last_place trip fetched using functions present in e-mission-server/emission/storage/decorations/timeline.py and data.enter_ts used | 
| 8. | MODE_INFERENCE | eacimr.predict_mode() | emission/analysis/classification/inference/mode/rule_engine.py | last_section_done | last predicted section fetched directly by indexing and data.end_ts used |
| 9. | LABEL_INFERENCE | eacilp.infer_labels() | emission/analysis/classification/inference/labels/pipeline.py | last_trip_done | iteratively last_trip_done's data.end_ts comparing with cleaned_trip's data.end_ts and updated if newer trip found |
| 10. | EXPECTATION_POPULATION | eaue.populate_expectations() | emission/analysis/userinput/expectations.py | last_trip_done | iteratively last_trip_done's data.end_ts comparing with expected_trip's data.end_ts and updated if newer trip found |
| 11. | CREATE_CONFIRMED_OBJECTS | eaum.create_confirmed_objects() | emission/analysis/userinput/matcher.py | last_processed_ts | last_place trip fetched using functions present in e-mission-server/emission/storage/decorations/timeline.py and data.enter_ts used |
| 12. | CREATE_COMPOSITE_OBJECTS | eapcc.create_composite_objects() | emission/analysis/plotting/composite_trip_creation.py | last_processed_ts | iteratively creating composite trips and using data.end_ts value |

MukuFlash03:
Having seen all these, I opted to go with the approach that first fetches all trip entries first before performing the operation. 
```
trips_to_export = self.get_exported_timeseries_entries(user_id, ts, time_query.startTs, time_query.endTs, export_queries)
...
...
# Inside get_exported_timeseries_entries()
(ts_db_count, ts_db_result) = ts._get_entries_for_timeseries(ts.timeseries_db, None, tq, geo_query=None, extra_query_list=None, sort_key = sort_key)
```

Then once the specific stage's primary operations are completed, then the last trip is directly indexed. I'm using the value present in `data.ts` to keep track of the last processed entry for the Purge stage.
```
            if len(trips_to_export) == 0:
                # Didn't process anything new so start at the same point next time
                self._last_processed_ts = None
            else:  
                self._last_processed_ts = trips_to_export[-1]['data']['ts']
```

MukuFlash03:
I thought of using `data.enter_ts` or `data.end_ts` like other stages were doing it by making use of the wrapper `Entry()`.
Also, these stages were using `ts.find_entries`() which combines entries from both timeseries_db and analysis_timeseries_db

So, I decided to use `_get_entries_for_timeseries()` to query just the timeseries db entries and retrieve raw trip data.
But this was giving empty entries when trying to convert result cursor obtained above. 

Hence, decided to go ahead with `data.ts`.
I did observed that this value is the same as `metadata.write_ts`, so not sure if this is the right value. 

I will take a second look and possibly using a similar `Entry()` based approach and see if `end_ts` can be used. But it seems these values are more catered for the analysis timeseries value and not raw data?

MukuFlash03:
**PENDING / TODO**:
1. Add additional test cases to add for the logic to handle last_processed_ts.
- Checking whether retrieved last_processed_ts is the max last_processed_ts in the raw exported data / restored data
- After running export / restore, re-run it to check whether incremental export occurs and entire data isn’t exported or restored every time.

2. Add last_processed_ts logic for Restore function as well.

MukuFlash03:
Summary of changes made since last time around (details in below comments)

1. Added export_type parameter for “full” or “incremental” export.
- Implemented both full and incremental export by keeping track of time ranges in 1 hour slots until current time is reached (present when pipeline / code is run).

-----

2. Added separate export_timeseries and import_timeseries scripts. Main difference between existing scripts:
- export_timeseries: Removed extra queries (place, trip, first_place) to ensure we deal only with timeseries database entries.
- import_timeseries: Removed loading pipeline states

-----

3. Improved tests for both export types (will improvise further): 
    1. Checking count of data before and after purging and restoring.
    2. Checking number of export files generated.
    3. Comparing object IDs of a few entries from DB and from export files
    4. Checking multiple restore doesn’t add duplicate entries.

-----

Some highlights from Export PR
- Full vs Incremental export discussion ([link](https://github.com/e-mission/e-mission-server/pull/824#:~:text=The%20default%20export_stage,streamed%20on%2Ddemand))
- last_processed_ts ([link](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-870708489))
- start_ts, end_ts ([link](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-871760577))
- File storage, FIFO stream ([link](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-872319092))

MukuFlash03:
Main questions I had which also give some reasoning around the changes I’ve made.
If my assumptions are not correct, will need to change approach for purging mainly with regards to what "ts" values to base logic around.

[Commit](https://github.com/e-mission/e-mission-server/pull/952/commits/ec162ad8141ee6a9606070a169ace620ff71d1b3) with questions in more detail.

---

1. Right approach for Export + Purge? (See below for details)
2. Do we need all 3 types of entries: locations, trips, places? (See below comment)
3. Is pipeline_states export necessary? (See below comment)

-------

Approach A
1. Export data in chunks to File
3. Delete exported data from DB.
4. Repeat until all data purged.

Flow looks like: Export -> Delete -> Export -> Delete -> Export -> Delete


Approach B
1. Export data in chunks to file.
2. Repeat until all data exported.
3. Delete all exported data from DB.

Flow looks like: Export -> Export -> Export ... -> Delete

MukuFlash03:
2. Do we need all 3 types of entries: locations, trips, places?

From [export.py](https://github.com/e-mission/e-mission-server/blob/master/emission/export/export.py)
```
    loc_time_query = estt.TimeQuery("data.ts", start_ts, end_ts)
    loc_entry_list = get_from_all_three_sources_with_retry(user_id, loc_time_query, databases)
    trip_time_query = estt.TimeQuery("data.start_ts", start_ts, end_ts)
    trip_entry_list = get_from_all_three_sources_with_retry(user_id, trip_time_query, databases)
    place_time_query = estt.TimeQuery("data.enter_ts", start_ts, end_ts)
    place_entry_list = get_from_all_three_sources_with_retry(user_id, place_time_query, databases)

```

For now, commented out code from [export_timeseries.py](https://github.com/e-mission/e-mission-server/pull/952/files#diff-ecec0e55923ab3c54d6eb414c136b83d2936790fff0907c1d05fc981abecdeb3) that gets all three types of entries.
If we only need location entries, can simplify code further to just work for these entries.

I went through Shankari’s thesis and understood that entries are broken down as per the travel diary structure: `place -> trip -> location`.

In my current logic, I’m using `data.ts` as the main time value.
If only trip entries present or only place entries, then directly choosing latest `data.ts` is incorrect since trips use `data.enter_ts` while places use `data.start_ts`

Searching codebase for references and read through Shankari’s thesis to `start_ts` and `enter_ts`.
I've gotten some hints that `start_ts` and `enter_ts` are `analysis_timeseries` entries?

In that case, can we ignore these since the purge + restore functionality is concerned primarily with raw timeseries data?

-----

**Hint 1: [Timeseries_Sample.ipynb](https://github.com/e-mission/e-mission-server/blob/master/Timeseries_Sample.ipynb?short_path=50f73aa#L446-L485)**
- ct_df fetches analysis/cleaned_trip entries -> analysis DB

```
# ——

trip_start_end_fuzz = 10 # seconds
ct_df = ts.get_data_df("analysis/cleaned_trip", time_query=None)
tl = esdl.get_cleaned_timeline(test_user_id, ct_df.iloc[0].start_ts - trip_start_end_fuzz, ct_df.iloc[-1].end_ts + trip_start_end_fuzz)

# ——

for e in tl:
    if 'enter_ts' in e.data:
        # Must be place-like
        print(e.metadata.key, e.data.enter_fmt_time, "->", e.data.exit_fmt_time)
    else:
        print(e.metadata.key, e.data.start_fmt_time, "->", e.data.end_fmt_time)

# ——
```

-----

**Hint 2:  bin/historical/migrations/populate_local_dt.py**

- Looks like old code, some changes were last made 8 years ago.
- The collection parameter refers to some non-time series databases as seen from the function calls.
- The entry[start_ts] or entry[‘enter_ts’] values are then used in the find query by setting data.ts to this value.

```
# ——

def fix_trips_or_sections(collection):
    tsdb = edb.get_timeseries_db()
    for entry in collection.find():
        start_loc_entry = tsdb.find_one({'user_id': entry['user_id'],
            'metadata.key': 'background/location', 'data.ts': entry['start_ts']})
        end_loc_entry = tsdb.find_one({'user_id': entry['user_id'],
            'metadata.key': 'background/location', 'data.ts': entry['end_ts']})


# ——

def fix_stops_or_places(collection):
    tsdb = edb.get_timeseries_db()
    for entry in collection.find():
        if 'enter_ts' in entry:
            enter_loc_entry = tsdb.find_one({'user_id': entry['user_id'],
                'metadata.key': 'background/location', 'data.ts': entry['enter_ts']})

# ——

    elif args.key == "trips":
        fix_trips_or_sections(edb.get_trip_new_db())
    elif args.key == "sections":
        fix_trips_or_sections(edb.get_section_new_db())
    elif args.key == "places":
        fix_stops_or_places(edb.get_place_db())

```

MukuFlash03:
Continued discussion of changes made:

1. Added export_type parameter for “full” or “incremental” export.

Discussion from Export PR

[Comment](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-870549213): Detailed discussion by kafitz on full vs incremental export + REST API
[Comment](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-870708489) : Shankari explained difference between full export and incremental export with usage of last_processed_ts for 1 hour’s worth of data
[Comment](https://github.com/e-mission/e-mission-server/pull/824#issuecomment-871760577): Queries by jruzekowicz around start_ts, end_ts, last_processed_ts, file storage, queue
[Comment](https://github.com/e-mission/e-mission-server/pull/824#pullrequestreview-698521194) : Next steps for intake pipeline to export one hour’s worth of data (don’t see this implemented)

------

**1. A. Last_processed_ts explained by Shankari** 

```
Our pipeline consists of multiple stages and for every stage we store the timestamp of how far along in the timeline the data has been processed. 
So if you think of, you know, a timeline as a long hello linear pathway into the past and the future and the last processed time stamp is a marker along that which indicates how much of the timeline you have processed. 

There is also last_run_ts which is when the pipeline was last run but last run is not the same as last processed because it could be that I went on a trip and then didn't have connectivity at the end of the trip. 
So the data still on my phone. But on the server, the pipeline runs every hour and so the pipeline has run, but it hasn't even seen my data yet.

It should be if the export code was written correctly, that if you export ones and then you export again and it well only it should only export the new data that came in after the previous export. 
That is the whole point of using a pipeline as opposed to a script. I mean, you know, using the pipeline instead of a naive script.  

For a one time export, you can run export script manually and then it's fine to write a script that takes a snapshot of the data and sends it over. And we do that with the Mongo dumps all the time.

But if you want to run this as part of an ongoing system and you wanna archive on an ongoing basis the raw data that comes in, you don't want to run a script that pulls all of the data every time, because that's inefficient. 
You just pulled it yesterday. Why do you need to pull all of the data again today, right?
```

MukuFlash03:
**1. B. What does 1 hour’s worth of data mean?**

“1 hour’s worth of data” -> different connotations:
- Export PR had analysis db entries as well as timeseries db entries.
- Purge PR, dealing with timeseries db entries.

For the export PR
- 1 hour is important since the analysis pipeline runs every hour and the export data sources include the both timeseries and analysis_timeseries DBs.
- Also, 1 hour would help reduce volume of data being processed.

For the purge PR,
- 1 hour with regards to analysis pipeline may not be relevant since we are dealing only with raw timeseries data.
- But at the same time, we can still have the purge pipeline run hourly to move new data to storage once the analysis pipeline has finished using the new data.
- There’s also the 2nd use case to help with large data volumes.

------

Time range of exported data in Purge pipeline.
- In any case, purge pipeline runs upto the current time or until no more raw timeseries entries present in db for the user.
- If Purge pipeline running for the first time for a user:
    - Then it will export and purge all the timeseries data for a user from its first entry with the earliest data.ts value.
    - This data can be really old data and first purge run might take a lot of time)
- If Purge pipeline has already been run before for a user:
    - Then it will set start_ts to last_processed_ts and export data from that point onwards.
    - Once our new purge pipeline system is in place, it would run hourly, then it would eventually just have a small subset of entries. This would be latest data that was just pulled into the timeseries db and is ready to be purged once it has been used in the analysis pipeline.

MukuFlash03:
**1.C. Does restore script also need to handle 1 hour data?**

I've based the [import_timeseries.py](https://github.com/e-mission/e-mission-server/pull/952/files#diff-0c29acd1f3ad909f4c097d8a31c6b83321f97325900a7013f6946b62a949b18b) script on the existing [load_multi_timeline_for_range.py](https://github.com/e-mission/e-mission-server/blob/master/bin/debug/load_multi_timeline_for_range.py) script.

We shouldn’t need to handle 1 hour logic in [restore_data.py](https://github.com/e-mission/e-mission-server/pull/952/files#diff-6b1def7c9c8637a39ed0f2946220c602477c3c6a0da441606f8041cb3a031809) since:
- Internally, `load_multi_timeline_for_range` already reads in batches with default `batch_size` = 10000 entries.
- We’ll be handling 1 hour logic in [purge_data.py](https://github.com/e-mission/e-mission-server/pull/952/files#diff-ce909e5d8c74b4dfa3bce2b1bf7c021efdaf73b72134d870c88b33bccd3659a8) while exporting entries so no need to handle it again in `restore_data` as we can be assured that our exported files will have 1 hour’s data only.
    - There is the case however where a lot of data volume is there even in 1 hour’s worth of data.
    - Again, we don’t need to worry as `import_timeseries` script processes data in batches.

MukuFlash03:
**1. D. Exit conditions for purge pipeline**

The break condition to exit the while loop would be:
1. Start_ts of any time range exceeds init_end_ts which is present time of running pipeline (default value set by pipeline)
```
            if current_start_ts >= initEndTs:
                break
```
2. No more entries within current time range AND no more entries in timeseries db.
```
            elif export_queries is None and count_entries_1 == 0:
                # Didn't process anything new so start at the same point next time
                # self._last_processed_ts = None
                logging.debug("No new data to export, breaking out of while loop")
                print("No new data to export, breaking out of while loop")
                break
```

While the first condition is fairly obvious, I had to consider the 2nd condition, since I came across an edge case where not all entries exported. 
This happened when: the difference in the ts values of entries is more than 1 hour.
In my previous logic, I just checked that there are no entries retrieved in a time range and exited the loop.
But there might still be entries in the timeseries db beyond the 1 hour window.

-----

Example from sample data:

```
Entry No: 1387 - 55affbab7d65cb39ee976734 1437596198.387 background/location
Entry No: 1388 - 55b08d347d65cb39ee976a8f 1437582760.583 statemachine/transition
```

The first time range gets entries from 1906 to 1498.
The second time range gets entries from 1497 to 1388.

For the third time range, `start_ts` is set as previous iteration’s `end_ts` = 1437585293.881.
And `end_ts` will be set to `start_ts` + 1 hour (or +3600 seconds) = 1437588893.881

Now, the next entry 1387 has a `ts value` that is more than 1 hour ahead of the current `end_ts`.
So, in the time range for the 3rd iteration `start_ts` (1437585293.881) to `end_ts` (1437588893.881), no entries are found.
Thus the loop was breaking despite there being more data available to export.

-----

Hence, I added the condition to check the timeseries db entries in each iteration.
The count of these raw entries would be updated since we are also deleting entries after exporting them.

MukuFlash03:
Continued discussion of changes made:

2. Added separate export_timeseries and import_timeseries scripts. 

Main difference between existing scripts:
a. [export_timeseries](https://github.com/e-mission/e-mission-server/pull/952/files#diff-ecec0e55923ab3c54d6eb414c136b83d2936790fff0907c1d05fc981abecdeb3) based off [extract_timeline_for_day_range_and_user](https://github.com/e-mission/e-mission-server/blob/master/bin/debug/extract_timeline_for_day_range_and_user.py)
- Removed extra queries (place, trip, first_place) to ensure we deal only with timeseries database entries.

b. [import_timeseries](https://github.com/e-mission/e-mission-server/pull/952/files#diff-0c29acd1f3ad909f4c097d8a31c6b83321f97325900a7013f6946b62a949b18b) based off [load_multi_timeline_for_range](https://github.com/e-mission/e-mission-server/blob/master/bin/debug/load_multi_timeline_for_range.py)
- Removed loading pipeline states

------

**A. Switched to using `data.ts` instead of `metadata.write_ts`**

Usercache is the first stage and there are not trips loaded hence we work with `metadata.write_ts`.
For most stages, we are using trip data -> `data.end_ts`. 
But these stages are a part of the analysis intake pipeline.

Since we’re dealing with raw timeseries data for purging / restoring data, we mainly have data.ts in the raw timeseries entries.
Sometimes the `data.ts` value doesn’t match the `metadata.write_ts` which I’m guessing is because of some latency between sensing the values at the time `data.ts` but writing it onto device storage occurs at `metadata.write_ts`

So I went ahead with using `data.ts` for the purposes of this PR.

MukuFlash03:
**2. B. Removed pipeline_states_export file**

Check [commit message](https://github.com/e-mission/e-mission-server/pull/952/commits/63f7985e5b045cbbff972318f37f018390e1e66b) for more details

------

To summarize:

Code referenced from [purge_data](https://github.com/e-mission/e-mission-server/pull/952/files#diff-ce909e5d8c74b4dfa3bce2b1bf7c021efdaf73b72134d870c88b33bccd3659a8)

**Problem faced**
- The `last_processed_ts` field in the exported pipeline state files is `NULL` because it's updated only after the function exits and the stage is marked as done. 
- During `run_purge_pipeline`, export_pipeline_states is called to export pipeline states. I got the idea to do this from [extract_timeline_for_day_range_and_user](https://github.com/e-mission/e-mission-server/blob/master/bin/debug/extract_timeline_for_day_range_and_user.py).
But `last_processed_ts` is only set in the DB after the function completes. 

**What I tried to resolve it**
- Moving the `export_pipeline_states` call to the parent function after the stage is marked as done would result in only one pipeline states file.
- For incremental exports, for handling multiple time ranges, I thought of using `earliest start_ts` and `latest end_ts` (current time) to name the pipeline states file. So instead of having multiple pipeline_states files, only one file would exist for the entire time range for the incremental export.
- But `load_multi_timeline_for_range` attempts to load pipeline states based on a file prefix. In the current code, it is assumed that the file_prefix for the actual raw data will match the pipeline_states file.
But if only one file exists for a range, it causes a FileNotFoundError.

-------

Pipeline states were added in Jan 2018
Commit that added it in extract_timeline_for_day_range_and_user.py
[b38366b](https://github.com/e-mission/e-mission-server/commit/b38366beb8d716b10d4e0a14d64a321150969375)

Commit that added it in load_multi_timeline_for_range.py
[06a0a4e](https://github.com/e-mission/e-mission-server/commit/06a0a4e2b241847807f43f89331808fc7c11388b)

Why do we need it as explained in commits above
“””
The raw data and the analysis results do not constitute the entire state of a
pipeline. In particular, if we store only the raw + analysis results, and then
we try to run the pipeline again, we will end up with two copies of the
analysis results.
“””

-----

Do we need it for purge PR?
- The commit message states that pipelinestates were also exported / loaded so that we don't have duplicate analysis entries.
- In the purge PR, we are dealing with `timeseries_db` data.

Thu, can we remove it from the purge_restore related code?

shankari:
@MukuFlash03 is this currently "Ready for review" or waiting for "Questions from Shankari"? Can you put it into the correct column?

shankari:
The pipeline state is part of the analysis pipeline, so is related to analysed data.
While archiving the raw data, we should not store the pipeline state.
And we certainly should not restore the pipeline state while restoring onto a running, production system.

PR #956, opened 2024-02-04
Improving Trip Segmentation by reducing DB calls

humbleOldSage:
The changes  below  that led to these performance upgrades are investigated in   https://github.com/e-mission/e-mission-docs/issues/1041 . They are :

1.  db calls for transition and motion dataframes are moved upstream from  `is_tracking_restarted_in_range` function  and `get_ongoing_motion_in_range` in `restart_checking.py` to `trip_segmentaiton.py`.  The old setting which had multiple db calls ( for each iteration ) now happen once in the improved setting.

2. All the other changes in `trip_segmentation.py` and `dwell_segmentation_dist_filter.py` are just to support the change  in point 1 ( above).

3. in `dwell_segmentation_time_filter.py`,other than the changes to support point 1 ( above), there an additional improvement.  The calculations for `last10PointsDistances` and `last5MinsPoints` are vectorised.  For this,  `calDistance` in `common.py` now supports numpy arrays.

### DIFF
--- a/emission/analysis/intake/segmentation/restart_checking.py
+++ b/emission/analysis/intake/segmentation/restart_checking.py
@@ -9,7 +9,7 @@ import logging
 import emission.core.wrapper.transition as ecwt
 import emission.storage.timeseries.timequery as estt
 
-def is_tracking_restarted_in_range(start_ts, end_ts, transition_df):
+def is_tracking_restarted_in_range(start_ts, end_ts, timeseries):
     """
     Check to see if tracing was restarted between the times specified
     :param start_ts: the start of the time range to check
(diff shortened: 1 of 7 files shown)


shankari:
> in dwell_segmentation_time_filter.py,other than the changes to support point 1 ( above), there an additional improvement. The calculations for last10PointsDistances and last5MinsPoints are vectorised. For this, calDistance in common.py now supports numpy arrays.

I don't see any indication of the performance improvement related to this change in the issue. I would anticipate the change to be minimal, since `len(last10PointsDistances) == 10` and vectorization is unlikely to be helpful until we reach thousands of rows. I am not sure that the additional code complexity is worth it.

Also, it would be helpful if you would link to the specific comment in the issue that documented the improvement, and even duplicate the numbers here to make it easier to look up...

humbleOldSage:
> I don't see any indication of the performance improvement related to this change in the issue. 

I missed  this improvement in the issue. I have now added.
https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1925828315


>I would anticipate the change to be minimal, since `len(last10PointsDistances) == 10` and vectorization is unlikely to be helpful until we reach thousands of rows. I am not sure that the additional code complexity is worth it.

Yes, it's just for 10 rows. But, since this part falls inside the loop, it's the overall loop time that is improved from 1.96 to 1.23s as mentioned in the issue.  

> ... and even duplicate the numbers here to make it easier to look up...

Sure.

humbleOldSage:
Overall, Current AndorraWrapper runtime is ~1.6s , iOS Runtime is ~0.4s and CombinedWrapper runtime is ~2.1

shankari:
@humbleOldSage there is a comment in the issue that indicates that there were some failures, but I don't see any updates on the fix. https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1909368396

Also, what are you testing this on? Why do we have the magic number of 327? Please also separate the vectorization from the code that actually reduces DB calls so we can evaluate each of them independently

humbleOldSage:
Did , in a separate PR https://github.com/e-mission/e-mission-server/pull/958 . Closing this one since no updates expected here.

PR #957, opened 2024-02-04
Faster Trip Invalidation

humbleOldSage:
Instead of invalidating one ooid from the list at a time, use UpdateOne and bulkwrite to
 invalidate entire list . This is supported by findings here  https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1892868291


shankari:
It is not clear what the **units** of the comparison are (seconds, milliseconds?)
https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1895088283

There are typically a few out of order elements, so single versus bulk insertion should not really matter. But we should verify before we abandon this change.

PR #958, opened 2024-02-06
Improving Trip Segmentation by reducing DB calls

humbleOldSage:
The changes  below  that led to these performance upgrades are investigated in   https://github.com/e-mission/e-mission-docs/issues/1041 . They are :

1.  db calls for transition and motion dataframes are moved upstream from  `is_tracking_restarted_in_range` function  and `get_ongoing_motion_in_range` in `restart_checking.py` to `trip_segmentaiton.py`.  The old setting which had multiple db calls ( for each iteration ) now happen once in the improved setting.

2. All the other changes in `trip_segmentation.py`,`dwell_segmentation_dist_filter.py` and `dwell_segmentation_time_filter.py` are just to support the change  in point 1 ( above).

### DIFF
--- a/emission/analysis/intake/segmentation/restart_checking.py
+++ b/emission/analysis/intake/segmentation/restart_checking.py
@@ -9,7 +9,7 @@ import logging
 import emission.core.wrapper.transition as ecwt
 import emission.storage.timeseries.timequery as estt
 
-def is_tracking_restarted_in_range(start_ts, end_ts, timeseries):
+def is_tracking_restarted_in_range(start_ts, end_ts, transition_df):
     """
     Check to see if tracing was restarted between the times specified
     :param start_ts: the start of the time range to check
@@ -17,28 +17,32 @@ def is_tracking_restarted_in_range(start_ts, end_ts, timeseries):
     :param timeseries: the timeseries to use for checking
     :return:
     """
-    import emission.storage.timeseries.timequery as estt
+    transition_df_start_idx=transition_df.ts.searchsorted(start_ts,side='left')    
+    transition_df_end_idx=transition_df.ts.searchsorted(end_ts,side='right')
+    transition_df_for_current=transition_df.iloc[transition_df_start_idx:transition_df_end_idx]
 
-    tq = estt.TimeQuery(timeType="data.ts", startTs=start_ts,
-                        endTs=end_ts)
-    transition_df = timeseries.get_data_df("statemachine/transition", tq)
-    if len(transition_df) == 0:
+    if len(transition_df_for_current) == 0:
         logging.debug("In range %s -> %s found no transitions" %
-                      (tq.startTs, tq.endTs))
+                      (start_ts, end_ts))
         return False
     logging.debug("In range %s -> %s found transitions %s" %
-                  (tq.startTs, tq.endTs, transition_df[["fmt_time", "curr_state", "transition"]]))
-    return _is_tracking_restarted_android(transition_df) or \
-           _is_tracking_restarted_ios(transition_df)
+                  (start_ts, end_ts, transition_df_for_current[["fmt_time", "curr_state", "transition"]]))
+    return _is_tracking_restarted_android(transition_df_for_current) or \
+           _is_tracking_restarted_ios(transition_df_for_current)
 
-def get_ongoing_motion_in_range(start_ts, end_ts, timeseries):
-    tq = estt.TimeQuery(timeType = "data.ts", startTs = start_ts,
-                        endTs = end_ts)
-    motion_list = list(timeseries.find_entries(["background/motion_activity"], tq))
+def get_ongoing_motion_in_range(start_ts, end_ts, motion_df):
+    ## in case when we receive an empty dataframe, there's nothing to
+    ## process
+    if motion_df.shape == (0,0): 
+        return motion_df
+
+    motion_df_start_idx=motion_df.ts.searchsorted(start_ts,side='left')    
+    motion_df_end_idx=motion_df.ts.searchsorted(end_ts,side='right')
+    filtered_motion_df=motion_df.iloc[motion_df_start_idx:motion_df_end_idx]
     logging.debug("Found %s motion_activity entries in range %s -> %s" %
-                  (len(motion_list), tq.startTs, tq.endTs))
-    logging.debug("sample activities are %s" % motion_list[0:5])
-    return motion_list
+                 (len(filtered_motion_df),start_ts, end_ts))
+    #logging.debug("sample activities are %s" % filtered_motion_df.head())
+    return filtered_motion_df
 
 def _is_tracking_restarted_android(transition_df):
     """
--- a/emission/analysis/intake/segmentation/trip_segmentation.py
+++ b/emission/analysis/intake/segmentation/trip_segmentation.py
@@ -7,7 +7,6 @@ standard_library.install_aliases()
 from builtins import *
 from builtins import object
 import logging
-import pandas as pd
 
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.storage.decorations.place_queries as esdp
--- a/emission/analysis/intake/segmentation/trip_segmentation.py
+++ b/emission/analysis/intake/segmentation/trip_segmentation.py
@@ -7,6 +7,7 @@ standard_library.install_aliases()
 from builtins import *
 from builtins import object
 import logging
+import pandas as pd
 
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.storage.decorations.place_queries as esdp
@@ -65,6 +66,12 @@ def segment_current_trips(user_id):
     # We need to use the appropriate filter based on the incoming data
     # So let's read in the location points for the specified query
     loc_df = ts.get_data_df("background/filtered_location", time_query)
+    transition_df = ts.get_data_df("statemachine/transition", time_query)
+    motion_df = ts.get_data_df("background/motion_activity",time_query)
+    if len(transition_df) > 0:
+        logging.debug("transition_df = %s" % transition_df[["fmt_time", "transition"]])
+    else:
+        logging.debug("no transitions found. This can happen for continuous sensing")
     if len(loc_df) == 0:
         # no new segments, no need to keep looking at these again
         logging.debug("len(loc_df) == 0, early return")
@@ -89,12 +96,12 @@ def segment_current_trips(user_id):
     if len(filters_in_df) == 1:
         # Common case - let's make it easy
         
-        segmentation_points = filter_methods[filters_in_df[0]].segment_into_trips(ts,
+        segmentation_points = filter_methods[filters_in_df[0]].segment_into_trips(transition_df,motion_df,ts,
             time_query)
     else:
         segmentation_points = get_combined_segmentation_points(ts, loc_df, time_query,
                                                                filters_in_df,
-                                                               filter_methods)
+                                                               filter_methods,transition_df,motion_df)
     # Create and store trips and places based on the segmentation points
     if segmentation_points is None:
         epq.mark_segmentation_failed(user_id)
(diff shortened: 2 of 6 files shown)


humbleOldSage:
Replying to  the comment https://github.com/e-mission/e-mission-server/pull/956#issuecomment-1926135255 here :

> Please also separate the vectorization from the code that actually reduces DB calls so we can evaluate each of them independently

The current  PR just has reduced DB calls. In this case, the runtime for androidWrapper is at  ~2.1s ( as mentioned here https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1903310181 in the issue) , IOS Wrapper at 0.4s and CombinedWrapper at ~2.6s. 

> Also, what are you testing this on?

I am using the `TestTripSegmentation.py` that read data from `emission/tests/data/real_examples/shankari_2015-aug-27`

>  Why do we have the magic number of 327? 

 When the data  in `emission/tests/data/real_examples/shankari_2015-aug-27` is setup,  querying for 'background/filtered_location' from `segment_into_trips` in`dwell_segmentation_time_filter.py`  returns  327 points .
 
         filtered_points_pre_ts_diff_df = timeseries.get_data_df("background/filtered_location", time_query)

The for loop then iterates on these points 

        for idx, row in filtered_points_df.iterrows():

humbleOldSage:
> there is a comment in the issue that indicates that there were some failures, but I don't see any updates on the fix. https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1909368396

This is in regards to the fact that the db call for  "background/filtered_location": 

     ts.get_data_df("background/filtered_location", time_query)

Is present in these 3 files `trip_segmentation`, `dwell_segmentation_dist_filter.py` and  `dwell_segmentation_time_filter.py`. Since  `dwell_segmentation_dist_filter.py` and  `dwell_segmentation_time_filter.py` are downstream from `trip_segmentation`, the calls in these 2 should be easily replaceable by passing dataframe from `trip_segementation` and thus there should just be 1 db call. However, if we do so the test for `combinedWrapper` in `TestTripSegmentation` fails.  So, for now there are separate calls in each function.

 This causes an overhead of ~0.2s for androidWrapper and IosWrapper in the current test setup.

However, there is nothing else failing here.

shankari:
> The current PR just has reduced DB calls. In this case, the runtime for androidWrapper is at ~2.1s ( as mentioned here https://github.com/e-mission/e-mission-docs/issues/1041#issuecomment-1903310181 in the issue) , IOS Wrapper at 0.4s and CombinedWrapper at ~2.6s.

And what was it before? Just reporting the new number does not capture the difference. I don't see a comment in the issue for the time taken in `androidWrapper` before your change. I only see "It took ~6.8s to run the entire segment_current_trips from ..."

> I am using the TestTripSegmentation.py that read data from emission/tests/data/real_examples/shankari_2015-aug-27

This is not a sufficient test. `shankari_2015-aug-27` only contains android data. That is why you see that "IOS Wrapper at 0.4s", it is not actually running any significant iOS code. You have also flagged this as "Looking into the Android run time for now.". You need to test and report real numbers on iOS as well.

humbleOldSage:
> > I am using the TestTripSegmentation.py that read data from emission/tests/data/real_examples/shankari_2015-aug-27
> 
> This is not a sufficient test. `shankari_2015-aug-27` only contains android data. 


   On second look, for the iOSWrapper, the test reads data from`iphone_2015-11-06`. All in all, there are indeed two real world example files, one for android, another for iOS in the test.


        etc.setupRealExample(self, "emission/tests/data/real_examples/shankari_2015-aug-27")
        self.androidUUID = self.testUUID

        .
        .
        .
        .
        with open("emission/tests/data/real_examples/iphone_2015-11-06") as fp:



> That is why you see that "IOS Wrapper at 0.4s", it is not actually running any significant iOS code.

I believe the IosWrapper is at 0.4s due to lesser trips, which we  come to know after running the pipeline. ( 2 in case of iOS vs 8 for android ). 

I tried other real example files with platform as iOS. The other files with iOS data are :

iphone_2016-02-22 ,
iphone_3_2016-02-22,
iphone_2016-02-22,
sunil_2016-07-20

once the example is set (using all of them one by one), the trips are either 0 or 2 trips for iOS, giving the similar runtime of ~0.4s - 0.5s.

humbleOldSage:
The  runtimes are :

|  |  iOS   | Android  | Combined   |
| ---------  |-------|---------|-------------|
| Previous   | ~0.6s | ~6.8s    | ~7.9s    |
| Current.   | ~0.4s  | ~2.1s    |~2.6s    |

shankari:
@JGreenlee scalability improvement

emission/analysis/intake/segmentation/restart_checking.py
```diff
@@ -9,36 +9,40 @@
...
     Check to see if tracing was restarted between the times specified
     :param start_ts: the start of the time range to check
     :param end_ts: the end of the time range to check
     :param timeseries: the timeseries to use for checking
     :return:
     """
-    import emission.storage.timeseries.timequery as estt
+    transition_df_start_idx=transition_df.ts.searchsorted(start_ts,side='left')    
+    transition_df_end_idx=transition_df.ts.searchsorted(end_ts,side='right')
+    transition_df_for_current=transition_df.iloc[transition_df_start_idx:transition_df_end_idx]
```
shankari, line 22: Is there a reason you are using this instead of something like

```
    transition_df_for_current=transition_df[transition_df.ts >= start_ts && transition_df.ts <= end_ts]
```

or 

```
    transition_df_for_current=transition_df.query('ts >= start_ts & .ts <= end_ts')
```

emission/analysis/intake/segmentation/restart_checking.py
```diff
@@ -9,36 +9,40 @@
...
-    motion_list = list(timeseries.find_entries(["background/motion_activity"], tq))
+def get_ongoing_motion_in_range(start_ts, end_ts, motion_df):
+    ## in case when we receive an empty dataframe, there's nothing to
+    ## process
+    if motion_df.shape == (0,0): 
+        return motion_df
+
+    motion_df_start_idx=motion_df.ts.searchsorted(start_ts,side='left')    
+    motion_df_end_idx=motion_df.ts.searchsorted(end_ts,side='right')
+    filtered_motion_df=motion_df.iloc[motion_df_start_idx:motion_df_end_idx]
```
shankari, line 41: Ditto

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -7,6 +7,7 @@
 from builtins import *
 from builtins import object
 import logging
+import pandas as pd
```
shankari, line 10: why do we need to add this import?

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_dist_filter.py
```diff
@@ -144,7 +144,7 @@ def segment_into_trips(self, timeseries, time_query):
                 logging.debug("Found %d transitions after last point, not ending trip..." % len(stopped_moving_after_last))
         return segmentation_points
 
-    def has_trip_ended(self, lastPoint, currPoint, timeseries):
+    def has_trip_ended(self, lastPoint, currPoint, timeseries, motion_df):
```
shankari, line 147: if you are already passing in the `motion_df` here, why do you need to also pass in the timeseries?

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_dist_filter.py
```diff
@@ -173,14 +173,14 @@ def has_trip_ended(self, lastPoint, currPoint, timeseries):
             # for this kind of test
             speedThreshold = old_div(float(self.distance_threshold * 2), (old_div(self.time_threshold, 2)))
 
-            if eaisr.is_tracking_restarted_in_range(lastPoint.ts, currPoint.ts, timeseries):
+            if eaisr.is_tracking_restarted_in_range(lastPoint.ts, currPoint.ts, self.transition_df):
```
shankari, line 176: why is `transition_df` a module field?

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_dist_filter.py
```diff
@@ -173,14 +173,14 @@ def has_trip_ended(self, lastPoint, currPoint, timeseries):
...
                 logging.debug("tracking was restarted, ending trip")
                 return True
 
             # In general, we get multiple locations between each motion activity. If we see a bunch of motion activities
             # between two location points, and there is a large gap between the last location and the first
             # motion activity as well, let us just assume that there was a restart
-            ongoing_motion_in_range = eaisr.get_ongoing_motion_in_range(lastPoint.ts, currPoint.ts, timeseries)
+            ongoing_motion_in_range = eaisr.get_ongoing_motion_in_range(lastPoint.ts, currPoint.ts, motion_df)
```
shankari, line 183: while motion_df is not. They are both read upfront and should be treated the same way for clarity

emission/analysis/intake/segmentation/trip_segmentation_methods/trip_end_detection_corner_cases.py
```diff
@@ -23,8 +23,8 @@ def is_huge_invalid_ts_offset(filterMethod, lastPoint, currPoint, timeseries,
                         ecwm.MotionTypes.NONE.value,
                         ecwm.MotionTypes.STOPPED_WHILE_IN_VEHICLE.value]
 
-    non_still_motions = [ma for ma in motionInRange if ma["data"]["type"] not in ignore_modes_list and ma["data"]["confidence"] == 100] 
-    logging.debug("non_still_motions = %s" % [(ecwm.MotionTypes(ma["data"]["type"]), ma["data"]["confidence"], ma["data"]["fmt_time"]) for ma in non_still_motions])
+    non_still_motions=motionInRange[~motionInRange['type'].isin(ignore_modes_list) & (motionInRange['confidence'] ==100)]
+    #logging.debug("non_still_motions = %s" % [(ecwm.MotionTypes(ma["data"]["type"]), ma["data"]["confidence"], ma["data"]["fmt_time"]) for ma in non_still_motions])    logging.debug("non_still_motions = %s" % [(ecwm.MotionTypes(ma["data"]["type"]), ma["data"]["confidence"], ma["data"]["fmt_time"]) for ma in non_still_motions])
```
shankari, line 27: this change doesn't appear to be related to this fix.

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
```diff
@@ -199,7 +199,7 @@ def continue_just_ended(self, idx, currPoint, filtered_points_df):
             else:
                 return False
 
-    def has_trip_ended(self, prev_point, curr_point, timeseries, last10PointsDistances, last5MinsDistances, last5MinTimes):
+    def has_trip_ended(self, prev_point, curr_point, timeseries, last10PointsDistances, last5MinsDistances, last5MinTimes, transition_df, motion_df):
```
shankari, line 202: Ditto. if we are passing in the transition_df and motion_df, why do we also need the timeseries?

shankari (changes requested):
I don't think that < 10 trips is going to give us a real sense of the scalability improvement. Please run more complex tests:
- `TestPipelineRealData` for a combination of correctness and performance
- against a real large dataset, after resetting the pipeline, for a rough estimate of performance.

emission/analysis/intake/segmentation/restart_checking.py
humbleOldSage: First was the performance reason, O(log (n)) here vs O (n) in others. 
Second, in case of query(), it  creates a boolean series that marks rows to keep or discard, which happens internally, which increases temporary memory usage. For very large DataFrames, this can be an issue.

can use this one ,if log(n) vs O(n) is not an issue : 

   ` transition_df_for_current=transition_df[transition_df.ts >= start_ts && transition_df.ts <= end_ts]`

emission/analysis/intake/segmentation/trip_segmentation.py
humbleOldSage: Unused currently. Removed.

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_dist_filter.py
humbleOldSage: it's being used here :

       timeseries.invalidate_raw_entry(currPoint["_id"])

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_time_filter.py
humbleOldSage: Here, it's unused. Cleared.

PR #959, opened 2024-02-07
Integration tests for Geofabrik Overpass API 

nataliejschultz:
Addressing e-mission/e-mission-docs#1036

As of now, there are three tests in the TestOverpass.py module:

1. `test_overpass`, which compares a call to the free version of overpass with a call to the geofabrik version
2. `test_get_stops_near`, which passes a set of coordinates (as a list inside a dictionary) to `get_stops_near` in `match_stops.py`.
3. `test_get_predicted_transit_mode`, which passes two sets of coordinates to `get_predicted_transit_mode` in `match_stops.py`.


nataliejschultz:
Currently, tests 2 and 3 only call the free version of the API to make calls. There are a few ways that come to mind to change this: 

1. Add hard-coded query to geofabrik as seen in test 1 and compare with the free call (not preferred)
2. Add a similar block of code to `match_stops.py` as we did for `nominatim.py` using environment variables:
```
try:
    NOMINATIM_QUERY_URL = os.environ.get("NOMINATIM_QUERY_URL")
    logging.info(f"NOMINATIM_QUERY_URL: {NOMINATIM_QUERY_URL}")
    print("Nominatim Query URL Configured:", NOMINATIM_QUERY_URL)

    if NOMINATIM_QUERY_URL is None:
        raise Exception("Nominatim query url not configured")
except:
    print("Nominatim URL not configured, place decoding must happen on the client")
```

3. Add a file called `overpass_server.json` with the geofabrik query URL and work with the current method of setting the query URL in `match_stops.py`
```
try:
    config_file = open('conf/net/ext_service/overpass_server.json')
except:
    print("overpass not configured, falling back to default overleaf.de")
    config_file = open('conf/net/ext_service/overpass_server.json.sample')
config_data = json.load(config_file)
url = config_data["url"]
```

I think the answer is a combination of 2 and 3, but I'd like some feedback from @shankari on my tests before deciding what to do

shankari:
@nataliejschultz high-level question: why do we need to keep the configuration in a file? There is a single variable in the file.
Recall that for the nominatim use case, we switched the config from a file to an environment variable. I would suggest using the same pattern here.

nataliejschultz:
Coming back to this as I wait for review on other PRs. 
The overpass url is now based on setting the OVERPASS_KEY environment variable in `match_stops.py` (rather than sample file), to be consistent with current direction of codebase:

```
try:
    OVERPASS_KEY = os.environ.get("OVERPASS_KEY")
    url = 'https://overpass.geofabrik.de/' + OVERPASS_KEY + '/'
    print("overpass configured")
except:
    print("overpass not configured, falling back to public overpass api")
    url = "https://lz4.overpass-api.de/"
```
I made sure that using `unset OVERPASS_KEY` in terminal triggered the `"overpass not configured"` print, and it did:
 
Screenshot 2024-06-17 at 1 28 05 PM


This readies the code to use the overpass key for calls in both the tests and any usage of match_stops.py.

Tests passing [on GitHub Actions](https://github.com/nataliejschultz/e-mission-server/actions/runs/9553313467) and locally:
Screenshot 2024-06-17 at 1 16 11 PM

emission/net/ext_service/transit_matching/match_stops.py
```diff
@@ -3,24 +3,23 @@
 import requests
 import attrdict as ad
 import itertools
-import copy
+import os
 import time
 
 try:
-    config_file = open('conf/net/ext_service/overpass_server.json')
+    OVERPASS_KEY = os.environ.get("OVERPASS_KEY")
```
shankari, line 10: nit: I would rename this to include GEOFABRIK in the name (e.g. `GEOFABRIK_OVERPASS_KEY`), since there might be other commercial providers of OSM data.

emission/individual_tests/TestOverpass.py
```diff
@@ -0,0 +1,62 @@
...
+
+#Set up query
+OVERPASS_KEY = os.environ.get("OVERPASS_KEY")
+
+#Sample loc1 = NREL East Gate
+loc1 = {'coordinates': [-105.16844103184974, 39.740428870224605]}
+#Sample loc2 = Denver Union Station
+loc2 = {'coordinates': [-105.00083982302972, 39.753710532185025]}
+#Sample loc3 = Grand Junction Train Station, CO
+loc3 = {'coordinates': [-108.57055213129632, 39.06472424640481]}
```
shankari, line 22: It's great that you have identified these three locations, but I only see one of them being used, and it is hardcoded. Do you plan to use the other two?

emission/individual_tests/setup_overpass_tests.sh
shankari, line 1: nit: please rename to `setup_and_test` or similar since it doesn't just setup, it also runs the tests.

shankari (changes requested):

emission/individual_tests/TestOverpass.py
nataliejschultz: locs 2 and 3 are used in test_get_predicted_transit_mode, lines 54 and 55, passed in as stops 1 and 2

shankari (approved):

PR #967, opened 2024-05-13
Nominatim test regression fix

nataliejschultz:
Attempting to fix the [regression](https://github.com/e-mission/e-mission-server/actions/runs/9048804406) that was identified.


nataliejschultz:
Adding the git install command allowed the test to pass [here](https://github.com/nataliejschultz/e-mission-server/actions/runs/9066573206)

shankari:
Squash merging to avoid commit churn

PR #968, opened 2024-05-16
make localdate queries behave like canonical datetime range filters

JGreenlee:
The existing implementation that was here was addressing the localdate fields separately and treating them like a set of filters rather than one date range from start -> end. The "set of filters" was good enough for some things, like filtering to a specific month, but not for the canonical datetime range filtering where you want everything between an arbitrary start and end datetime. The complexity in the logic to form the necessary NoSQL queries is more than expected, but after the dust settles it's about the same amount of code as the old implementation was.

In short, this implementation finds what units are used by both start and end localdates, walks down them (from largest unit to smallest unit), and combines $and and $or blocks to determine the upper and lower bound

### DIFF
--- a/emission/storage/decorations/local_date_queries.py
+++ b/emission/storage/decorations/local_date_queries.py
@@ -11,47 +11,40 @@ import arrow
 
 import emission.core.wrapper.localdate as ecwl
 
-def get_range_query(field_name, start_local_dt, end_local_dt):
-    if list(start_local_dt.keys()) != list(end_local_dt.keys()):
-        raise RuntimeError("start_local_dt.keys() = %s does not match end_local_dt.keys() = %s" %
-            (list(start_local_dt.keys()), list(end_local_dt.keys())))
-    query_result = {}
-    for key in start_local_dt:
-        curr_field = "%s.%s" % (field_name, key)
-        gte_lte_query = {}
-        try:
-            start_int = int(start_local_dt[key])
-        except:
-            logging.info("start_local_dt[%s] = %s, not an integer, skipping" %
-                            (key, start_local_dt[key]))
-            continue
+def get_range_query(field_prefix, start_ld, end_ld):
+    units = [u for u in ecwl.DATETIME_UNITS if u in start_ld and u in end_ld]
+    logging.debug(f'get_range_query: units = {units}')
+    try:
+        gt_query = get_comparison_query(field_prefix, start_ld, end_ld, units, 'gt')
+        lt_query = get_comparison_query(field_prefix, end_ld, start_ld, units, 'lt')
+        logging.debug(f'get_range_query: gt_query = {gt_query}, lt_query = {lt_query}')
+        return { "$and": [gt_query, lt_query] } if gt_query and lt_query else {}
+    except AssertionError as e:
+        logging.error(f'Invalid range from {str(start_ld)} to {str(end_ld)}: {str(e)}')
+        return None
 
-        try:
-            end_int = int(end_local_dt[key])
-        except:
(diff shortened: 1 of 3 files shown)


JGreenlee:
The failing test:

```python
    def testLocalRangeRolloverQuery(self):
        """
        Search for all entries between 8:18 and 9:08 local time, both inclusive
        """
        start_local_dt = ecwl.LocalDate({'year': 2015, 'month': 8, 'hour': 8, 'minute': 18})
        end_local_dt = ecwl.LocalDate({'year': 2015, 'month': 8, 'hour': 9, 'minute': 8})
        final_query = {"user_id": self.testUUID}
        final_query.update(esdl.get_range_query("data.local_dt", start_local_dt, end_local_dt))
        entries = edb.get_timeseries_db().find(final_query).sort('data.ts', pymongo.ASCENDING)
        self.assertEqual(448, edb.get_timeseries_db().count_documents(final_query))

        entries_list = list(entries)

        # Note that since this is a set of filters, as opposed to a range, this
        # returns all entries between 18 and 8 in both hours.
        # so 8:18 is valid, but so is 9:57
        self.assertEqual(ecwe.Entry(entries_list[0]).data.local_dt.hour, 8)
        self.assertEqual(ecwe.Entry(entries_list[0]).data.local_dt.minute, 18)
        self.assertEqual(ecwe.Entry(entries_list[-1]).data.local_dt.hour, 9)
        self.assertEqual(ecwe.Entry(entries_list[-1]).data.local_dt.minute, 57)
```

Since the query returned from `get_range_query` now behaves like a proper datetime range query, we should only see the entries in the time between 8:18 and 9:08. 9:57 is not valid anymore.

JGreenlee:
The most recent run of the test found 232 entries, but expected 448. That is because it was expecting entries after 9:08.
I inspected `shankari_2015-aug-27` and verified that there are indeed 232 entries if 9:08 (inclusive) is the cutoff.

I am going to modify the test to expect 232 entries.

shankari (changes requested):
@JGreenlee I am fine with merging this for now to unblock the dashboard changes, but this is fairly complex, so could you please add in some comments, and maybe examples of the comparison query outputs so we can understand the code better.

emission/storage/decorations/local_date_queries.py
```diff
@@ -11,47 +11,40 @@
...
-
-        if is_rollover:
-            gte_lte_query = get_rollover_query(start_int, end_int)
-        else:
-            gte_lte_query = get_standard_query(start_int, end_int)
-
-        if len(gte_lte_query) > 0:
-            query_result.update({curr_field: gte_lte_query})
+def get_range_query(field_prefix, start_ld, end_ld):
+    units = [u for u in ecwl.DATETIME_UNITS if u in start_ld and u in end_ld]
```
shankari, line 15: For example, what if `u` is in `start_ld` but not in `end_ld`. If I specify a month in `start_ld` but no month in `end_ld`, I intuitively want to get all trips after the start but be open-ended wrt the end. Alternatively, you should check that there always is a matching pair as I did before.

shankari (commented):
Finally, at a high level, this converts the local date queries into ranges, but drops the old filtering capability completely. For local dates, I think we actually want both - I want to be able to see data for a range based on local time, but I also want to be able to see only weekday trips versus weekend trips (for example).

Again, the current implementation was a weird mishmash of range and filter, so I am glad we fixed it. I am merging this now, but I do think we should add back filter support at some point, maybe when we actually need it.

JGreenlee:
Will address these comments when I get a chance – trying to finish the metrics implementation first

PR #970, opened 2024-06-07
for yyyy_mm_dd metrics, query by fmt_time instead of local_dt

JGreenlee:
I recently rewrote the local dt queries used by TimeComponentQuery (https://github.com/e-mission/e-mission-server/pull/968) to make them behave as datetime range queries. But this results in some pretty complex queries which could have nested $and and $or conditions. It felt overengineered but that was the logic required if we were to query date ranges by looking at local dt objects where year, month, day, etc are all recorded separately. Unfortunately, those queries run super slowly on production. I think this is because the $and / $or conditions prevented MongoDB from optimizing efficiently via indexing

Looked for a different solution, I found something better and simpler. At first I didn't think using fmt_time fields with $lt and $gt comparisons would work becuase they're ISO strings, not numbers. But luckily MongoDB can compare strings this way. it performs a lexicographical comparison where 0-9 < A-Z < a-z (like ASCII). ISO has the standard format 0000-00-00T00:00:00
The dashes and the T will always be in the same places, so effectively, only the numbers will be compared. And the timezone info is at the end, so it doesn't get considered as long as we don't include it in the start and end inputs. The only thing that specifically needs to be handled is making the end range inclusive. If I query from "2024-06-01" to "2024-06-03", an entry like "2024-06-03T23:59:59-04:00" should match. But lexicographically, that comes after "2024-06-03". Appending a "Z" to the end range solves this.

The result of all this is that the TimeQuery class (timequery.py) is now able to handle either timestamp or ISO dates, and this is what metrics.py will use for summarize_by_yyyy_mm_dd.

### DIFF
--- a/emission/core/wrapper/localdate.py
+++ b/emission/core/wrapper/localdate.py
@@ -9,9 +9,6 @@ import logging
 import arrow
 import emission.core.wrapper.wrapperbase as ecwb
 
-# specify the order of time units, from largest to smallest
-DATETIME_UNITS = ['year', 'month', 'day', 'hour', 'minute', 'second']
-
 class LocalDate(ecwb.WrapperBase):
     """
     Supporting wrapper class that stores the expansions of the components
--- a/emission/net/api/metrics.py
+++ b/emission/net/api/metrics.py
@@ -12,7 +12,7 @@ import emission.analysis.result.metrics.time_grouping as earmt
 import emission.analysis.result.metrics.simple_metrics as earms
 import emission.storage.decorations.analysis_timeseries_queries as esda
 import emission.storage.decorations.local_date_queries as esdl
-import emission.storage.timeseries.timequery as esttq
+import emission.storage.timeseries.fmt_time_query as estf
 
 import emcommon.metrics.metrics_summaries as emcms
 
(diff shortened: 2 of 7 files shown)


JGreenlee:
@shankari
Since this method will work for querying datetime ranges and is probably more performant, should I revert https://github.com/e-mission/e-mission-server/pull/968 back to the old "set of filters" implementation? You mentioned it was useful for certain things like weekday/weekend filtering.
Maybe go back to the old implementation but rename it to "filter query" instead of "range query" ?

shankari:
>  Maybe go back to the old implementation but rename it to "filter query" instead of "range query" ?

I vote for "filter query" instead of "range query"

JGreenlee:
I realized that the TimeQuery on `e-mission-server` (which works w/ MongoDB) mirrors TimeQuery in the native code plugins (which work with SQL dbs).

For that reason, I don't want to rename anything in `TimeQuery` right now. So instead of extending the functionality of `TimeQuery` to work with either ts or fmt_time, I'm going to make a new class called `FmtTimeQuery` (or similar)

I briefly read up on SQL queries to make sure they work the same way when comparing strings. They do – so later on, we could implement `FmtTimeQuery` on the native code UserCache plugin too

shankari (approved):
- Glad to see that this worked out; the ISO folks designed their format correctly!
- I am merging this for now, but we should add tests for the new `FmtTimeQuery`, similar to the current filter query
- I assume that the changes to the local date queries are just reverting the previous changes. I spot checked and that does seem to be true, but please flag locations that were not a simple revert so that I can take a closer look later.

PR #973, opened 2024-08-02
📍 Store the entire nominatim response; not just 'address'

JGreenlee:


shankari:
The packaging changes rewrite is taking too long, so I'm merging this proactively. This should allow @JGreenlee  to fix 
https://github.com/e-mission/e-mission-phone/pull/1166 so we don't have to change it again immediately.

shankari:
@JGreenlee can you pull from master to this branch so that this incorporates https://github.com/e-mission/e-mission-server/commit/86614ede2288d407236d9653b8a9f41c3fcb5150 and the automated tests pass?

shankari (approved):

PR #981, opened 2024-09-16
Update version of e-mission-common

Abby-Wheelis:
Running into issues when trying to import newer functionality to the public dashboard because the version of e-mission-common is `0.5.3` and does not include the functionality that I need

### DIFF
--- a/setup/environment36.yml
+++ b/setup/environment36.yml
@@ -19,7 +19,7 @@ dependencies:
 - scipy=1.10.0
 - utm=0.7.0
 - pip:
-  - git+https://github.com/JGreenlee/e-mission-common@0.5.3
+  - git+https://github.com/JGreenlee/e-mission-common@0.6.0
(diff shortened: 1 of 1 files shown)


Abby-Wheelis:
@JGreenlee would updating to `0.6.1` now create any issues for you? For the newest round of dashboard changes we will need a higher version than the current `0.5.3` so we can get the base mode colors (then soon to get the footprints, though that change is still a draft for the time being)

JGreenlee:
If I've done everything correctly,then no, upgrading to 0.6.1 on `e-mission-server` will not break anything. 
Let me verify that though.

JGreenlee:
@Abby-Wheelis Yes, it is fine.
`e-mission-common` 0.6.1 preserves backwards compatibility with the version of e-mission-phone that is on production currently, so upgrading the server now is not a problem

shankari (approved):

PR #986, opened 2024-10-02
Stats Timing

TeachMeTW:
# Add Functionality and Unit Tests for Function Timing and Error Handling

## Overview
Introduces new functionality for tracking execution times and error handling in the dashboard through the `store_dashboard_time` and `store_dashboard_error` decorators. These functions store timing and error data for various operations. Additionally, unit tests are added to verify that execution times and errors are correctly logged.

### 1. New Functionality in `stats_query.py`
- **`store_dashboard_time`**:
    - This function records the execution time of a code fragment or function and stores it in the database under the key `stats/dashboard_time`.
    - It captures the timestamp and the elapsed time using a `Timer` object and stores it via the `store_stats_entry` method.

- **`store_dashboard_error`**:
    - This function captures and logs the time at which an error occurs during the execution of a code fragment or function.
    - It stores the error data in the database under the key `stats/dashboard_error`.

### 2. New in `TestStatsQuery`
- **New Methods**:
    - **`setUpClass`**: Initializes the timeseries database before any tests are executed.
    - **`tearDown`**: Ensures that database entries related to the test are cleared after each run to maintain isolation.
    - **`verify_stats_entries`**: Verifies that expected entries (execution times or errors) are stored correctly in the database.

- **Unit Tests**:
    - **`test_single_function_timing`**:
        - Times the execution of a single function and stores the timing data in the database.
        - Verifies that the correct execution time is logged.
        
    - **`test_multiple_functions_timing`**:
        - Times and verifies two separate functions and their combined execution time.
        - Checks that individual and combined timings are accurately logged.

    - **`test_faulty_function_timing`**:
        - Simulates a faulty function that raises an exception.
        - Ensures that even when a function fails, the timing and error data are correctly logged as an error in the database.

## Impact
- This PR introduces essential functionality to log and track the performance and failures of code fragments or functions.
- The added unit tests ensure that these features are working correctly and provide robust coverage for timing and error handling.


TeachMeTW:
@shankari 

Current implementation on commit d19392d builds upon existing code base of `store_stats_entry` where `store_function_time` calls `store_stats_entry(user_id, "stats/function_time", stage_string, ts, reading)`


It takes a similar approach to `run_intake_pipeline` from `intake_stage.py` but instead of user_id; it takes a function:

```
def run_function_pipeline_step(func, skip_if_no_new_data):
    """
    Execute a single step in the function pipeline.

    :param func: The function to execute
    :param skip_if_no_new_data: Flag to determine if the function should be skipped based on custom logic
    :return: None
    """
    func_name = func.__name__

    with ect.Timer() as timer:
        logging.info(f"********** Function {func_name}: Starting execution **********")
        print(f"{arrow.now()} ********** Function {func_name}: Starting execution **********")
        result = func()

    # Store the execution time
    esds.store_function_time(func_name, "EXECUTION",
                             time.time(), timer.elapsed)

    if skip_if_no_new_data and not result:
        print(f"No new data for {func_name}, and skip_if_no_new_data = {skip_if_no_new_data}, skipping the rest of the pipeline")
        return
    else:
        print(f"Function {func_name} executed with result = {result} and skip_if_no_new_data = {skip_if_no_new_data}, continuing")

    logging.info(f"********** Function {func_name}: Completed execution **********")
```




## An example use case:
```
# Define test functions
def test_function_1():
    logging.info("Executing test_function_1")
    time.sleep(1)  # Simulate processing time
    return True  # Indicate successful execution

def test_function_2():
    logging.info("Executing test_function_2")
    time.sleep(1)
    return True

def test_function_faulty():
    logging.info("Executing test_function_faulty")
    time.sleep(1)
    raise ValueError("Simulated error in test_function_faulty")

def test_function_3():
    logging.info("Executing test_function_3")
    time.sleep(1)
    return True

if __name__ == "__main__":
    # Ensure the logs directory exists
    os.makedirs("logs", exist_ok=True)
    
    # Define the list of test functions, including the faulty one
    function_list = [
        test_function_1,
        test_function_2,
        test_function_faulty,  # This will raise an exception
        test_function_3  # This should execute normally after the faulty function
    ]
    
    # Run the pipeline with process number 1 and skip_if_no_new_data set to True
    run_function_pipeline(process_number=1, function_list=function_list, skip_if_no_new_data=True)

```

## Concerns:

Now I am unsure how to change the query from user_id to function_id/make a new db collection for it within timeseries; the metadata is right but the schema is not. I cannot access it on the notebook due to a keyerror: `KeyError: 'stats/function_time'` but I can access it on the mongo via: `db.Stage_timeseries.find({ "user_id": "test_function_1" }).pretty()`

### Output of the Find:

```
{
        "_id" : ObjectId("6700f706b37230e8be74d445"),
        "user_id" : "test_function_1",
        "metadata" : {
                "key" : "stats/function_time",
                "platform" : "server",
                "write_ts" : 1728116486.4357288,
                "time_zone" : "America/Los_Angeles",
                "write_local_dt" : {
                        "year" : 2024,
                        "month" : 10,
                        "day" : 5,
                        "hour" : 1,
                        "minute" : 21,
                        "second" : 26,
                        "weekday" : 5,
                        "timezone" : "America/Los_Angeles"
                },
                "write_fmt_time" : "2024-10-05T01:21:26.435729-07:00"
        },
        "data" : {
                "name" : "EXECUTION",
                "ts" : 1728116486.4328,
                "reading" : 1.001198583
        }
}
```

## Questions:

How can I update the location/not use the same as pipeline time or is it ok? Do I need extra steps for aggregate? Is this what you had envisioned?

shankari:
@TeachMeTW let us please add this to the project board so that I can see it

shankari:
@TeachMeTW

I am not sure why we are focusing on instrumenting one function at a time using a decorator. I believe that the issue that @JGreenlee brought up was that we don't always want to instrument a whole function, but may want to parts of code that are part of a function. That's why I suggested that we use the `with...` pattern and pointed to the pipeline as an example.

It seems like you have created a new `time_functions.py` module that is similar to the pipeline module? I am not quite sure what the new function is supposed to do. Are you suggesting that we should replace the existing pipeline function with this code?

Maybe it would be easier if I could write down what I expect you to do in greater detail.
- You should create a new function called `store_dashboard_time` in `emission/storage/decorations/stats_queries.py`
- You should use it to create `stats/dashboard/time` entries in the database
- You should use that function in the dashboard code for timing using the `with` pattern

You do not need to edit the existing pipeline.
Let me see if a detailed code review makes this any easier.

emission/storage/decorations/stats_queries.py
```diff
@@ -46,3 +48,22 @@ def store_stats_entry(user_id, metadata_key, name, ts, reading):
...
+def store_function_time(user_id: str, stage_string: str, ts: float, reading: float):
+    """
+    Stores the execution time of a function.
+
+    Parameters:
+    - user_id (str): The ID of the user.
+    - stage_string (str): The name of the function being timed.
+    - ts (float): The timestamp when the function execution started.
+    - reading (float): The duration of the function execution in milliseconds.
+
```
shankari, line 60:
```suggestion
    """
    Stores statistics about execution times in dashboard code. Both of our current dashboards generate _aggregate_ metrics. I don't see that changing in the foreseeable future, since we don't really want to work at a per-user level in the python dashboards. So we don't pass in the user_id, only a string indicating the name of the step being instrumented, and the value.
    
    Parameters:
    - code_fragment_name (str): The name of the function being timed.
    - ts (float): The timestamp when the function execution started.
    - reading (float): The duration of the function execution in milliseconds.
```

shankari (changes requested):
@TeachMeTW @JGreenlee please see my comments in https://github.com/e-mission/op-admin-dashboard/pull/141#pullrequestreview-2363746948 for an example of how to use the newly created function.

To clarify, the suggestions below are intended to be guidelines to show you what I expect. I have not tested them, and I don't promise that they will work without any changes.

shankari:
@TeachMeTW not sure if you consider these changes done.
If you think that they are done, please move them back to "Ready for review"
Note also that the tests are failing so they are not actually done.
Before you move to "Ready for review" please indicate testing done for the new functionality

**To clarify, the suggestions above are intended to be guidelines to show you what I expect. I have not tested them, and I don't promise that they will work without any changes.**

TeachMeTW:
@shankari No not done yet; just added your suggestions -- I'm still missing the main part of the implementation.

TeachMeTW:
@shankari 

Dashboard time now is in Stage_analysis_timeseries; not too sure how to get it to `stats/dashboard/time`
```
> db.Stage_analysis_timeseries.find({ "metadata.key": "stats/dashboard_time" }).pretty()
{
	"_id" : ObjectId("670c31a0ff3d9aa6f3653e41"),
	"user_id" : null,
	"metadata" : {
		"key" : "stats/dashboard_time",
		"platform" : "server",
		"write_ts" : 1728852384.882551,
		"time_zone" : "America/Los_Angeles",
		"write_local_dt" : {
			"year" : 2024,
			"month" : 10,
			"day" : 13,
			"hour" : 13,
			"minute" : 46,
			"second" : 24,
			"weekday" : 6,
			"timezone" : "America/Los_Angeles"
		},
		"write_fmt_time" : "2024-10-13T13:46:24.882551-07:00"
	},
	"data" : {
		"name" : "test_function_1",
		"ts" : 1728852383.862375,
		"reading" : 1005.1576660000001
	}
}
{
	"_id" : ObjectId("670c31a3ff3d9aa6f3653e43"),
	"user_id" : null,
	"metadata" : {
		"key" : "stats/dashboard_time",
		"platform" : "server",
		"write_ts" : 1728852387.351635,
		"time_zone" : "America/Los_Angeles",
		"write_local_dt" : {
			"year" : 2024,
			"month" : 10,
			"day" : 13,
			"hour" : 13,
			"minute" : 46,
			"second" : 27,
			"weekday" : 6,
			"timezone" : "America/Los_Angeles"
		},
		"write_fmt_time" : "2024-10-13T13:46:27.351635-07:00"
	},
	"data" : {
		"name" : "test_function_2",
		"ts" : 1728852385.34658,
		"reading" : 2004.730167
	}
}
{
	"_id" : ObjectId("670c31a6ff3d9aa6f3653e44"),
	"user_id" : null,
	"metadata" : {
		"key" : "stats/dashboard_time",
		"platform" : "server",
		"write_ts" : 1728852390.402445,
		"time_zone" : "America/Los_Angeles",
		"write_local_dt" : {
			"year" : 2024,
			"month" : 10,
			"day" : 13,
			"hour" : 13,
			"minute" : 46,
			"second" : 30,
			"weekday" : 6,
			"timezone" : "America/Los_Angeles"
		},
		"write_fmt_time" : "2024-10-13T13:46:30.402445-07:00"
	},
	"data" : {
		"name" : "test_function_3",
		"ts" : 1728852387.399441,
		"reading" : 3002.8265420000002
	}
}
```

I've updated the test and some code; see latest commit

emission/tests/funcTests/TestFunctionTiming.py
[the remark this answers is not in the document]
TeachMeTW: @shankari see latest commit if this fix is what you had in mind.

emission/storage/decorations/stats_queries.py
```diff
@@ -44,5 +46,23 @@ def store_stats_entry(user_id, metadata_key, name, ts, reading):
     "reading": reading
   }
   new_entry = ecwe.Entry.create_entry(user_id, metadata_key, data)
-  return esta.TimeSeries.get_time_series(user_id).insert(new_entry)
+  return esta.TimeSeries.get_non_user_time_series().insert(new_entry)
```
shankari, line 49: @TeachMeTW you cannot change this line - it will break the existing uses of `store_stats_entry` (e.g. in `store_pipeline_time`).

You will need to check for the presence of `user_id` and call the appropriate method, or call `esta.TimeSeries.get_non_user_time_series().insert(new_entry)` directly here

https://github.com/e-mission/e-mission-server/pull/986/files#diff-0e4360245e864d2687db5f80e19494b632e5bfcc4fb27f75eca4672e75ee8091R63

Please document the pros and cons of that choice before making the change.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, user_id):
                 "stats/server_api_time": self.timeseries_db,
                 "stats/server_api_error": self.timeseries_db,
                 "stats/pipeline_time": self.timeseries_db,
+                "stats/dashboard_time": self.timeseries_db,
```
shankari, line 57: this should be a `non_user_timeseries_db` instead of either `timeseries` or `analysis_timeseries`, both of which are user keyed.

emission/storage/timeseries/non_user_timeseries.py
```diff
@@ -5,10 +5,11 @@
 
 import emission.core.get_database as edb
 import emission.storage.timeseries.builtin_timeseries as bits
+import emission.core.wrapper.entry as ecwe  # Added missing import
 
 class NonUserTimeSeries(bits.BuiltinTimeSeries):
     def __init__(self):
-        super(AggregateTimeSeries, self).__init__(None)
+        super(NonUserTimeSeries, self).__init__(None)  # Corrected superclass name
```
shankari, line 12: This is incorrect. We are defining the `NonUserTimeSeries` here, we can't also inherit from it at the same time.

emission/storage/timeseries/non_user_timeseries.py
```diff
@@ -17,26 +18,27 @@ def get_uuid_list():
         return []
 
     def get_timeseries_db(self, key):
-        return self.timeseries_db
+        return self.timeseries_db  # Ensure this is intended to ignore 'key'
```
shankari, line 21: this should be `non_user_timeseries_db` as well

emission/storage/timeseries/non_user_timeseries.py
```diff
@@ -51,16 +53,28 @@ def insert(self, entry):
         Inserts the specified entry and returns the object ID 
         """
         logging.debug("insert called")
-        if type(entry) == dict:
+        if isinstance(entry, dict):
```
shankari, line 56: why do we need to change this?

emission/storage/timeseries/non_user_timeseries.py
```diff
@@ -51,16 +53,28 @@ def insert(self, entry):
...
+
+        logging.debug(f"Inserting entry {entry} into timeseries")
+        try:
+            result = collection.insert_one(entry)
+            logging.debug(f"Inserted entry with ID: {result.inserted_id}")
+            return result.inserted_id
+        except pymongo.errors.PyMongoError as e:
+            logging.error(f"Failed to insert entry: {e}")
+            raise
+
```
shankari, line 77: I don't think this is wrong, but can you highlight why this was needed?

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,95 @@
+# emission/tests/funcTests/TestFunctionTiming.py
+
+import logging
+import time
+from typing import Callable, List
+
+# Import the store_dashboard_time and store_dashboard_error functions
+from emission.storage.decorations.stats_queries import (
```
shankari, line 8: we don't use `from X import Y` in our code; our standard is to use `import X as x` and then use `x.function` in the rest of the code.

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,95 @@
...
+    store_dashboard_time,
+    store_dashboard_error
+)
+
+# Import the existing Timer context manager
+from emission.core.timer import Timer as ECT_Timer
```
shankari, line 14: Ditto

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Store the execution time
+        store_dashboard_time(
+            code_fragment_name=function_name,
+            ts=timestamp,
+            reading=elapsed_ms
+        )
+        print(f"Function '{function_name}' executed successfully in {elapsed_ms:.2f} ms.")
+        logging.info(f"Function '{function_name}' executed successfully in {elapsed_ms:.2f} ms.")
+
```
shankari, line 67: this checks that the store works without errors, but does not check that the store was actually successful and that the data can be read.

shankari (changes requested):
> Dashboard time now is in Stage_analysis_timeseries; not too sure how to get it to stats/dashboard/time

I am not sure what that statement means; let's unpack it a little.
1. Dashboard time should not be in the analysis_timeseries after this change. I do not see any references to the analysis_timeseries here. Have you cleared out the database from previous runs?
2. I don't understand what `stats/dashboard/time` you are referring to. Can you clarify where you expect to see that?

emission/storage/timeseries/non_user_timeseries.py
TeachMeTW: @shankari When I put `AggregateTimeSeries` instead of `NonUserTimeSeries`;

I get 
```
  File "/Users/rsimpson/e-mission-server/emission/tests/funcTests/TestFunctionTiming.py", line 95, in <module>
    main()
  File "/Users/rsimpson/e-mission-server/emission/tests/funcTests/TestFunctionTiming.py", line 92, in main
    execute_and_time_function(func)
  File "/Users/rsimpson/e-mission-server/emission/tests/funcTests/TestFunctionTiming.py", line 74, in execute_and_time_function
    store_dashboard_error(
  File "/Users/rsimpson/e-mission-server/emission/storage/decorations/stats_queries.py", line 78, in store_dashboard_error
    store_stats_entry(None, "stats/dashboard_error", code_fragment_name, ts, reading)
  File "/Users/rsimpson/e-mission-server/emission/storage/decorations/stats_queries.py", line 56, in store_stats_entry
    time_series = esta.TimeSeries.get_non_user_time_series()
  File "/Users/rsimpson/e-mission-server/emission/storage/timeseries/abstract_timeseries.py", line 30, in get_non_user_time_series
    return nuts.NonUserTimeSeries()
  File "/Users/rsimpson/e-mission-server/emission/storage/timeseries/non_user_timeseries.py", line 12, in __init__
    super(AggregateTimeSeries, self).__init__(None)
NameError: name 'AggregateTimeSeries' is not defined
```

emission/storage/timeseries/non_user_timeseries.py
TeachMeTW: I didn't know where the current tests were being stored so included this to track it down; hence where I found it in aggregate

shankari:
I added the non_user_timeseries as part of https://github.com/e-mission/e-mission-server/pull/509 to support storing an aggregate random forest model across all the users in a program. We abandoned that approach, in favor of the GIS-based mode detection, so we currently don't generate and store any aggregate models (only user specific models). However, we may resurrect it when it comes to the behavior/demographic based models.

So the question is: should we store the dashboard timing data into non_user_timeseries?
- Pro: it is not associated with a user
- Con: we will be storing analysed data in the non_user_timeseries, and that is what it was designed for, so we will end up with a mixture of raw and analysed data in the non_user_timeseries.

What we really need is non_user_raw and non_user_analysed timeseries.

However, I will point out that we already store non-user (although not really aggregate) raw data into the timeseries.
Timing data from the API for calls through the webservice are stored with `user_id = None`
So at least for now, we can be consistent with this, and just store the dashboard timing stats with `user_id = None`

The disadvantage of this is that we cannot make any assertions that the timeseries data will always have a user_id
So for clarity, it may make sense to pull out the non user data into a separate collection.
But we can do that later in a cleanup phase and do it consistently for both the API calls and these timing calls.

So let's just store this in the timeseries for now; it will also require less plumbing from @TeachMeTW

emission/storage/decorations/stats_queries.py
shankari, line 49: revert this back to the `get_time_series`

emission/storage/timeseries/non_user_timeseries.py
shankari, line 12: fix this.

emission/storage/timeseries/non_user_timeseries.py
shankari, line 21: can stay as `timeseries_db`

emission/tests/funcTests/TestFunctionTiming.py
shankari: you should be able to check that it is in the timeseries_db.

shankari (changes requested):
After the revelations in https://github.com/e-mission/e-mission-server/pull/986#issuecomment-2412110159 we are not going to use `non_user_timeseries_db` after all, instead we will store in `timeseries_db` with `user_id = None`.

So you can revert the changes listed here.

TeachMeTW:
@shankari @JGreenlee Since we are no longer using `non_user_timeseries`, I revered those changes and fixed the code that was requested. I also modified the logging and added verification. It is now stored in `Stage_timeseries`. Am I in the right direction?

```
> db.Stage_timeseries.find({ "metadata.key": "stats/dashboard_time" }).pretty()
{
	"_id" : ObjectId("670d8267fb380e467d11478f"),
	"user_id" : null,
	"metadata" : {
		"key" : "stats/dashboard_time",
		"platform" : "server",
		"write_ts" : 1728938599.210308,
		"time_zone" : "America/Los_Angeles",
		"write_local_dt" : {
			"year" : 2024,
			"month" : 10,
			"day" : 14,
			"hour" : 13,
			"minute" : 43,
			"second" : 19,
			"weekday" : 0,
			"timezone" : "America/Los_Angeles"
		},
		"write_fmt_time" : "2024-10-14T13:43:19.210308-07:00"
	},
	"data" : {
		"name" : "test_function_1",
		"ts" : 1728938598.1843588,
		"reading" : 1005.0427499999998
	}
}

```

I also verified with:
```
total_documents = gdb.get_timeseries_db().count_documents({
    "metadata.key": "stats/dashboard_time"
})
print(f"Total documents in Stage_timeseries with metadata.key 'stats/dashboard_time': {total_documents}")
```

image

image

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,161 @@
...
+            code_fragment_name=function_name,
+            ts=timestamp,
+            reading=elapsed_ms
+        )
+        print(f"Function '{function_name}' failed after {elapsed_ms:.2f} ms with error: {e}")
+        logging.error(f"Function '{function_name}' failed after {elapsed_ms:.2f} ms with error: {e}")
+
+        # Verification: Adjusted Error Query to Match Document Structure
+        timeseries_db = gdb.get_timeseries_db()
+        current_time = time.time()
```
shankari, line 121: where is this used?

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,161 @@
...
+        sdq.store_dashboard_error(
+            code_fragment_name=function_name,
+            ts=timestamp,
+            reading=elapsed_ms
+        )
+        print(f"Function '{function_name}' failed after {elapsed_ms:.2f} ms with error: {e}")
+        logging.error(f"Function '{function_name}' failed after {elapsed_ms:.2f} ms with error: {e}")
+
+        # Verification: Adjusted Error Query to Match Document Structure
+        timeseries_db = gdb.get_timeseries_db()
```
shankari, line 120: please change to use the timeseries interface to read the data as well. That will make it easier to (potentially) switch to a different database under the hood later. if we choose to switch a specialized timeseries DB/event DB instead of mongodb.

shankari (changes requested):
Yup, this is almost there.
Before I do a detailed review and approve, please fix this high-level issue.

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+        print(f"Function '{function_name}' executed successfully in {elapsed_ms:.2f} ms.")
+        logging.info(f"Function '{function_name}' executed successfully in {elapsed_ms:.2f} ms.")
+
+        # Verification: Adjusted Query to Match Document Structure
+        timeseries_db = gdb.get_timeseries_db()
+        
+        # Define a time window (e.g., +/- 5 seconds) to account for any timing discrepancies
+        time_window = 5  # seconds
+        current_time = time.time()
```
JGreenlee, line 74: Why do you need to give a +/- tolerance?
You just called `store_dashboard_time` with `ts=timestamp, reading=elapsed_ms`. 
So don't we expect `data.ts` to be exactly `timestamp` and `data.reading` to be exactly `elapsed_ms` ?

TeachMeTW: I added that because originally the verification was failing -- thought it had something to do with tolerance. I can remove it and retest

emission/tests/funcTests/TestFunctionTiming.py
TeachMeTW: @shankari What do you mean by timeseries interface?

TeachMeTW: @shankari this is the built in edb method, i just named it gdb, can rename it to edb like your previous functions

TeachMeTW: is this what you meant? `timeseries_db = builtin_ts.get_timeseries_db(key="stats/dashboard_time")`

TeachMeTW:
Verified that dashboard error is stored correctly as well:
```
> db.Stage_timeseries.find({ "metadata.key": "stats/dashboard_error" }).pretty()
{
	"_id" : ObjectId("670da2a9e09a7956e93d89f3"),
	"user_id" : null,
	"metadata" : {
		"key" : "stats/dashboard_error",
		"platform" : "server",
		"write_ts" : 1728946857.6014469,
		"time_zone" : "America/Los_Angeles",
		"write_local_dt" : {
			"year" : 2024,
			"month" : 10,
			"day" : 14,
			"hour" : 16,
			"minute" : 0,
			"second" : 57,
			"weekday" : 0,
			"timezone" : "America/Los_Angeles"
		},
		"write_fmt_time" : "2024-10-14T16:00:57.601447-07:00"
	},
	"data" : {
		"name" : "test_function_faulty",
		"ts" : 1728946856.595753,
		"reading" : 1005.088458
	}
}

```

emission/tests/funcTests/TestFunctionTiming.py
shankari: No, I meant that you should use the timeseries interface (defined in `emission.storage.timeseries.abstract_timeseries`  - look at existing implementations in the server or dashboard code for examples.

Do *not* use the mongodb collections directly - the new methods will not work if we switch to a different underlying database later.

TeachMeTW: Ah I was trying to find regular time series; did not realize it was abstract

TeachMeTW: @shankari I used the `abstract_timeseries`; modified `_get_query` in the built-in to take a dict AND object as opposed to just object (for verification that we are putting the items in the db)

TeachMeTW: also added more logging/error check for debugging

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -130,28 +130,53 @@ def _get_query(self, key_list = None, time_query = None, geo_query = None,
...
-            ret_query.update(time_query.get_query())
-        if geo_query is not None:
-            ret_query.update(geo_query.get_query())
-        if extra_query_list is not None:
+        
+        # Update with user-specific queries
+        if hasattr(self, 'user_query') and isinstance(self.user_query, dict):
+            ret_query.update(self.user_query)
+        elif hasattr(self, 'user_query'):
+            logging.warning("user_query is not a dict and will not be used.")
```
shankari, line 138: I don't understand why this is needed. Please explain the rationale behind this change.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -130,28 +130,53 @@ def _get_query(self, key_list = None, time_query = None, geo_query = None,
...
-                                         list(overlap_keys))
-                else:
-                    ret_query.update(extra_query)
+                if overlap_keys:
+                    logging.info(
+                        f"eq_keys = {eq_keys}, curr_keys = {curr_keys}, overlap_keys = {overlap_keys}"
+                    )
+                    raise AttributeError(f"extra query would overwrite keys {list(overlap_keys)}")
+                ret_query.update(extra_query)
+
```
shankari, line 179: Extraneous changes; please revert to make the code review clean

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -67,20 +68,21 @@ def execute_and_time_function(func: t.Callable[[], bool]):
...
         # Verification: Adjusted Query to Match Document Structure
-        timeseries_db = builtin_ts.get_timeseries_db(key="stats/dashboard_time")
-        
+        timeseries_db = esta.TimeSeries.get_time_series(None)
 
-        query = {
-            "metadata.key": "stats/dashboard_time",
-            "data.name": function_name,
+        # Define the time range and additional filters
+        time_query = {
```
shankari, line 74: we should not be hardcoding the `time_query` in mongodb format. Please look at prior examples of how queries are implemented.

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -67,20 +68,21 @@ def execute_and_time_function(func: t.Callable[[], bool]):
...
+            "data.reading": {"$gte": elapsed_ms, "$lte": elapsed_ms}
         }
-        
-        # Retrieve the most recent document for the function
-        stored_document = timeseries_db.find_one(
-            query,
-            sort=[("data.ts", -1)]
+
+        # Retrieve the first matching entry using the get_first_entry method
+        stored_document = timeseries_db.get_first_entry(
```
shankari, line 81: I don't think we should use `get_first_entry` here - just because you wrote `find_one` in the old code doesn't mean that you should change `get_first_entry` in the new code.

This is not actually even correct because if you test multiple functions, you will have multiple entries in the database. So the first entry is not the one you should be checking every time.

shankari (changes requested):
@TeachMeTW please check with @JGreenlee tomorrow before moving this back for my review!

emission/storage/timeseries/builtin_timeseries.py
TeachMeTW: I changed the simple if `{}_query is not None` checks to `hasattr` and `isinstance` to allow query to be optionally included as a dictionary while ensuring it doesn't disrupt existing object-based queries.

emission/tests/funcTests/TestFunctionTiming.py
TeachMeTW: @shankari Would using `get_entry_at_ts` be a better alternative? Since the functions run at diff ts (unless its parallel)

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,151 @@
...
+
+        elapsed_seconds = timer.elapsed  # Accessing the float attribute directly
+        elapsed_ms = elapsed_seconds * 1000  # Convert to milliseconds
+
+        # Store the execution time
+        sdq.store_dashboard_time(
+            code_fragment_name=function_name,
+            ts=timestamp,
+            reading=elapsed_ms
+        )
```
JGreenlee, line 66: Why not have `store_dashboard_time` accept `timer` as a parameter? That way, you won't need to repeat these lines of code everywhere you want to instrument

JGreenlee: Then to verify that the reading was stored, check the most recent `stats/dashboard_time` entry and validate the `reading` lines up with `timer.elapsed`

(To check the most recent entry, should we implement `find_one` on the aggregate timeseries?)

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,151 @@
...
+        stored_document = timeseries_db.get_entry_at_ts(
+            key="stats/dashboard_time",
+            ts_key="data.ts",
+            ts=timestamp
+        )
+
+        if stored_document:
+            # Inspect the stored document
+            stored_ts = stored_document.get("data", {}).get("ts", 0)
+            stored_reading = stored_document.get("data", {}).get("reading", 0)
```
JGreenlee, line 83: I would just use `stored_document['data']['reading']` and let the `except` handle the rest for you
We expect this to be an entry so it should have `['data']['reading']`

Sometimes it is better to ask for forgiveness than permission

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,151 @@
...
+    timestamp = time.time()
+
+    logging.info(f"Starting timing for function: {function_name}")
+
+    try:
+        with ec_timer.Timer() as timer:
+            result = func()  # Execute the test function
+
+        elapsed_seconds = timer.elapsed  # Accessing the float attribute directly
+        elapsed_ms = elapsed_seconds * 1000  # Convert to milliseconds
```
JGreenlee, line 58: Why not just have both `ts` and `reading` be in seconds?
Were the pipeline stats stored in milliseconds?

JGreenlee: Ok, looking at `store_pipeline_time` I see that the reading is in `ms`. So we can stay consistent with that.

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # Store the execution time
+        sdq.store_dashboard_time(
+            code_fragment_name=function_name,
+            timer=timer
+        )
```
JGreenlee, line 64: After you make this call, the assertions you want to make:
- there is a new entry that exists
- it has `reading` equal to `timer.elapsed * 1000`
- it has `name` equal to `function_name`

emission/tests/funcTests/TestFunctionTiming.py
JGreenlee, line 1: I would structure this file as class, `TestFunctionTiming` that extends `unittest.TestCase` to be consistent with the rest of `emission.tests`

emission/tests/funcTests/TestFunctionTiming.py
JGreenlee: Should we implement find_one on the aggregate timeseries?
What's the advantage of doing that vs using `get_first_entry` ?

.gitignore
```diff
@@ -22,3 +22,5 @@ conf/**/*.json
 *.ipynb_checkpoints*
 
 *.secret
+
+*.ipynb
```
JGreenlee, line 26:
```suggestion
```

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -150,6 +177,7 @@ def _get_query(self, key_list = None, time_query = None, geo_query = None,
                                          list(overlap_keys))
                 else:
                     ret_query.update(extra_query)
+
```
JGreenlee, line 180: try to minimize whitespace-only changes

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,206 @@
...
+# Import the existing Timer context manager
+import emission.core.timer as ec_timer
+
+# Import the database module for verification
+import emission.storage.timeseries.abstract_timeseries as esta
+import emission.core.get_database as edb
+
+class TestFunctionTiming(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
```
JGreenlee, line 21:
```suggestion
    def setUpClass(self):
```

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        Set up resources before any tests are run.
+        """
+        logging.basicConfig(level=logging.INFO)
+        cls.timeseries_db = esta.TimeSeries.get_time_series(None)
```
JGreenlee, line 26:
```suggestion
    def setUpClass(self):
        """
        Set up resources before any tests are run.
        """
        logging.basicConfig(level=logging.INFO)
        self.timeseries_db = esta.TimeSeries.get_time_series(None)
```

emission/storage/timeseries/builtin_timeseries.py
TeachMeTW: Can revert this change; it was for debugging when i was using the mongo style query

TeachMeTW:
@shankari I used `find_entry` with time_query of None per discussed and previous implementations. Let me know if this is right.

shankari:
@TeachMeTW are you done with your changes? If so, please move back to "Ready for review"
I am not going to review PRs that are still being worked on

TeachMeTW:
> @TeachMeTW are you done with your changes? If so, please move back to "Ready for review" I am not going to review PRs that are still being worked on

It is now moved @shankari

emission/storage/timeseries/non_user_timeseries.py
```diff
@@ -78,6 +92,7 @@ def update(entry):
         """
         raise AttributeError("non_user_timeseries does not support updates")
 
+
```
TeachMeTW, line 95: This white space doesnt exist in the copy of code I have but shows up here

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        # Define the metadata keys to clean
+        keys_to_clean = ["stats/dashboard_time", "stats/dashboard_error"]
+        try:
+            # Delete entries where metadata.key is in keys_to_clean from timeseries_db
+            edb.get_timeseries_db().delete_many({"metadata.key": {"$in": keys_to_clean}})
+            logging.info("Cleaned up database entries after test based on metadata keys.")
+        except Exception as e:
+            logging.error(f"Error cleaning up database entries: {e}")
+            raise
```
shankari, line 40: If you are tearing down the database entries after every test, you will never end up with two entries, and will not be able to verify that the entries don't override each other.

You need to have a test that invokes multiple functions.

We are also not instrumenting only in chunks of one function at a time - we also want to instrument code blocks. Where is the test for that?

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,225 @@
...
+            # Assert that at least one document was retrieved
+            self.assertTrue(
+                len(stored_documents) > 0,
+                f"Data for '{function_name}' was not found in the database."
+            )
+
+            # Iterate over each document and inspect its contents
+            for idx, stored_document in enumerate(stored_documents):
+                try:
+                    stored_ts = stored_document['data']['ts']
```
shankari, line 111: with the way your test is currently set up, there will only be one entry at a time.

emission/tests/funcTests/TestFunctionTiming.py
shankari: > Why not have `store_dashboard_time` accept `timer` as a parameter? 

That is a great suggestion, @JGreenlee, and a clear refinement of my previous implementation. I think that what @TeachMeTW has implemented now is good enough to unblock us, but maybe we can go back to this at some time in the future and refine further (convert the pipeline time functions to this form as well, put the stat name string into the timer etc)

shankari (changes requested):
Again, please have @JGreenlee review before you send it over to me.

emission/storage/timeseries/builtin_timeseries.py
```diff
@@ -128,18 +130,43 @@ def _get_query(self, key_list = None, time_query = None, geo_query = None,
...
+        # Handle geo_query
+        if geo_query:
+            if hasattr(geo_query, 'get_query') and callable(geo_query.get_query):
+                geo_query_dict = geo_query.get_query()
+                ret_query.update(geo_query_dict)
+            elif isinstance(geo_query, dict):
+                ret_query.update(geo_query)
+            else:
+                raise TypeError("geo_query must have a get_query method or be a dict")
+
```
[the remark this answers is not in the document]
TeachMeTW: The reason for these changes was that I believe the mongo query was a dict but previous implementation only took objects; basically it boils down to flexibility. By allowing time_query and geo_query to be either objects with a get_query method or dictionaries, the code becomes more flexible and adaptable to different use cases (like the dict vs objects).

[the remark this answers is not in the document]
shankari: But it also makes it more tied to mongodb. If you wanted to change the underlying database to something else (e.g. postgres, or TimescaleDB), you would need to go through and find every invocation of `find_entries` and change it. That is the reason we have an abstract interface in the first place.

We can argue about the need for flexibility versus abstraction further, but not now. This change is not needed for this particular PR. We should not be making unnecessary core library changes as part of an unrelated PR.

Please revert.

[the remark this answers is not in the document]
TeachMeTW: Reverted

emission/storage/timeseries/non_user_timeseries.py
shankari, line 1: please revert this file since it is not related to this change. Your code is also wrong (AggregateTimeSeries is not in `emission.core` but in `emission.storage`)

shankari (changes requested):
There are some also unaddressed changes from before.

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,435 @@
...
+    time.sleep(5)
+    return True
+
+def test_function_faulty():
+    logging.info("Executing test_function_faulty")
+    time.sleep(1)
+    raise ValueError("Simulated error in test_function_faulty")
+
+if __name__ == "__main__":
+    unittest.main()
```
JGreenlee, line 435:
```suggestion
```

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,435 @@
...
+        # Log details of each stored document
+        for doc in stored_documents:
+            try:
+                name = doc['data']['name']
+                reading = doc['data']['reading']
+                ts = doc['data']['ts']
+                logging.info(f"Stored Entry - Name: {name}, Reading: {reading} ms, Timestamp: {ts}")
+            except KeyError as e:
+                logging.error(f"Missing key {e} in stored document: {doc}")
+
```
JGreenlee, line 114:
```suggestion
```

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,435 @@
...
+        logging.info("########### Finished test_function_long_delay ###########\n")
+
+    def test_function_faulty(self):
+        """
+        Test execution and timing of test_function_faulty, which is expected to raise an exception.
+        """
+        logging.info("\n########### Starting test_function_faulty ###########")
+        with self.assertRaises(ValueError) as context:
+            self.execute_and_time_function(test_function_faulty)
+        self.assertIn("Simulated error in test_function_faulty", str(context.exception))
```
JGreenlee, line 73: Is there a cleaner way to expect an error?

JGreenlee: Maybe something like 

```python
try:
   self.execute_and_time_function(test_function_faulty);
   self.assert(false)
except:
  self.assert(true)
```

emission/tests/funcTests/TestFunctionTiming.py
JGreenlee, line 1: I think this now does what we want it to, but it seems really bulky (over 400 lines) for a fairly simple task.
Is it overcomplicated? Is it overkill wrt logs?

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,435 @@
...
+
+            # Assert that the stored_name matches function_name exactly
+            self.assertEqual(
+                stored_name,
+                function_name,
+                msg=(
+                    f"Stored name '{stored_name}' does not match function "
+                    f"name '{function_name}'."
+                )
+            )
```
JGreenlee, line 309: This is all repeated in `test_code_block_instrumentation`. Let's extract it to a separate function called in both places

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,435 @@
...
+                stored_name_error,
+                function_name,
+                msg=(
+                    f"Stored error name '{stored_name_error}' does not match function "
+                    f"name '{function_name}'."
+                )
+            )
+
+            # Re-raise the exception to let the test fail
+            raise
```
JGreenlee, line 378: this except block as well

emission/tests/funcTests/TestFunctionTiming.py
```diff
@@ -0,0 +1,435 @@
...
+        block_name = "realistic_sample_code_block"
+        logging.info(f"Starting timing for code block: {block_name}")
+        
+        try:
+            with ec_timer.Timer() as timer:
+                # Sample code block to be timed with additional complexity
+                logging.info("Executing 'realistic' sample code block with helper functions")
+                data = self.helper_function_1()
+                processed_data = self.helper_function_2(data)
+                self.helper_function_3(processed_data)
```
JGreenlee, line 152: I don't see a reason to create 3 functions whose only purposes are to perform fake operations. One fake operation like `{k: v.upper() for k, v in data.items()}` or just a call to `sleep()` is fine

emission/tests/funcTests/TestFunctionTiming.py
JGreenlee, line 1: I don't think you need to run your `test_function_*`s more than once, because currently you call them one-by-one in individual tests and multiple times in one test. Just do "multiple".

It might sound silly to cut this out after you've just written it, but I just think when a test file is this large it will be hard to maintain in future (by someone who is probably not you)

TeachMeTW:
Tests now run like this:
```
########### Starting test_code_block_instrumentation ###########
Starting timing for 'sample_code_block'.
Executing sample_code_block with simulated operations
Step 1: Data retrieval from mock service
Data retrieved: {'user_id': 123, 'action': 'login'}
Step 2: Data processing
Processed data: {'user_id': '123', 'action': 'LOGIN'}
Step 3: Storing processed data
Processed data stored successfully
'sample_code_block' executed successfully in 2014.81 ms.
Verifying entry for 'sample_code_block' under key 'stats/dashboard_time'.
Stored Document for 'sample_code_block': reading=2014.8075409999997 ms, name=sample_code_block
'sample_code_block': Expected elapsed time = 2014.81 ms, Stored reading = 2014.8075409999997 ms.
########### Finished test_code_block_instrumentation ###########

Cleaned up database entries after test based on metadata keys.
.
########### Starting test_faulty_function_timing ###########
Starting timing for 'faulty_function'.
Executing faulty_function
'faulty_function' failed after 1001.40 ms with error: Simulated error in faulty_function
Verifying entry for 'faulty_function' under key 'stats/dashboard_error'.
Stored Document for 'faulty_function': reading=1001.4026669999998 ms, name=faulty_function
'faulty_function': Expected elapsed time = 1001.40 ms, Stored reading = 1001.4026669999998 ms.
########### Finished test_faulty_function_timing ###########

Cleaned up database entries after test based on metadata keys.
.
########### Starting test_multiple_functions_timing ###########
Starting timing for 'function_one'.
Executing function_one
'function_one' executed successfully in 1004.34 ms.
Verifying entry for 'function_one' under key 'stats/dashboard_time'.
Stored Document for 'function_one': reading=1004.3434590000002 ms, name=function_one
'function_one': Expected elapsed time = 1004.34 ms, Stored reading = 1004.3434590000002 ms.
Starting timing for 'function_two'.
Executing function_two
'function_two' executed successfully in 1500.24 ms.
Verifying entry for 'function_two' under key 'stats/dashboard_time'.
Stored Document for 'function_two': reading=1500.2447080000004 ms, name=function_two
'function_two': Expected elapsed time = 1500.24 ms, Stored reading = 1500.2447080000004 ms.
########### Finished test_multiple_functions_timing ###########

Cleaned up database entries after test based on metadata keys.
.
########### Starting test_single_function_timing ###########
Starting timing for 'sample_function'.
Executing sample_function
'sample_function' executed successfully in 2005.10 ms.
Verifying entry for 'sample_function' under key 'stats/dashboard_time'.
Stored Document for 'sample_function': reading=2005.0953330000007 ms, name=sample_function
'sample_function': Expected elapsed time = 2005.10 ms, Stored reading = 2005.0953330000007 ms.
########### Finished test_single_function_timing ###########

Cleaned up database entries after test based on metadata keys.
.
Ran 4 tests in 7.690s

OK
```


Lines cut from 400 -> 200

JGreenlee:
https://github.com/TeachMeTW/e-mission-server/pull/1/commits/a24b03bb38d24f39652e217bf484915b99354248

JGreenlee:
@TeachMeTW After you have adjusted comments and docstrings, can you provide an overall summary of the changes in this PR?
Since this PR is 40+ commits I fully anticipate that we will squash merge. Shankari can copy your summary into the commit message of the squash so that we have a clean and meaningful git history.

TeachMeTW:
@shankari @JGreenlee 
Modified the PR description and heres an overview:


Add function timing and error logging with unit tests

- Implemented `store_dashboard_time` and `store_dashboard_error` functions to log execution times and errors in the database.
- Added unit tests in `TestStatsQuery` to verify correct timing and error logging for individual, multiple, and faulty function executions.

JGreenlee (approved):
LGTM, great work!

I think the amount of comments is still more verbose than I would have done, but I'll get off my soapbox and let @shankari state a preference (if any)

TeachMeTW:
Yes your right it does seem a bit trivial on the comments side; looking back now it seems repetitive

shankari (approved):

shankari:
I just merged this, but didn't squash before merging. Argh!
Trying to fix this now by modifying the git history (since I don't want to have commit messages with `resolve`

JGreenlee:
@shankari for squash message

> Add function timing and error logging with unit tests
> 
> * Implemented `store_dashboard_time` and `store_dashboard_error` functions to log execution times and errors in the database.
> * Added unit tests in `TestStatsQuery` to verify correct timing and error logging for individual, multiple, and faulty function executions.

PR #990, opened 2024-11-01
Add per-operation timing to segment_current_trips using ect.Timer

TeachMeTW:
- Wrapped each significant operation within the `segment_current_trips` function with `ect.Timer` context managers.
- Named each timer using the pattern `ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/operation"` for consistent identification.
- After each timed block, recorded the elapsed time by calling `esds.store_pipeline_time` with the appropriate parameters.
- Ensured that only timing-related code was added without altering existing logic, error handling, or formatting.

This enhancement enables granular performance monitoring of the trip segmentation process, allowing for better identification of potential bottlenecks and optimization opportunities.

### DIFF
--- a/emission/analysis/intake/segmentation/trip_segmentation.py
+++ b/emission/analysis/intake/segmentation/trip_segmentation.py
@@ -7,6 +7,7 @@ standard_library.install_aliases()
 from builtins import *
 from builtins import object
 import logging
+import time
 
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.storage.decorations.place_queries as esdp
@@ -23,6 +24,9 @@ import emission.core.wrapper.untrackedtime as ecwut
 import emission.analysis.intake.segmentation.restart_checking as eaisr
 
 import emission.core.common as ecc
+import emission.storage.decorations.stats_queries as esds
+import emission.core.timer as ect
+import emission.core.wrapper.pipelinestate as ecwp
 
 class TripSegmentationMethod(object):
     def segment_into_trips(self, timeseries, time_query):
@@ -47,54 +51,76 @@ class TripSegmentationMethod(object):
         pass
     
 def segment_current_trips(user_id):
-    ts = esta.TimeSeries.get_time_series(user_id)
-    time_query = epq.get_time_range_for_segmentation(user_id)
+    with ect.Timer() as t_get_time_series:
+        ts = esta.TimeSeries.get_time_series(user_id)
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_time_series", time.time(), t_get_time_series.elapsed)
+
+    with ect.Timer() as t_get_time_range:
+        time_query = epq.get_time_range_for_segmentation(user_id)
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_time_range_for_segmentation", time.time(), t_get_time_range.elapsed)
 
     import emission.analysis.intake.segmentation.trip_segmentation_methods.dwell_segmentation_time_filter as dstf
     import emission.analysis.intake.segmentation.trip_segmentation_methods.dwell_segmentation_dist_filter as dsdf
-    dstfsm = dstf.DwellSegmentationTimeFilter(time_threshold = 5 * 60, # 5 mins
-                                              point_threshold = 9,
-                                              distance_threshold = 100) # 100 m
 
-    dsdfsm = dsdf.DwellSegmentationDistFilter(time_threshold = 10 * 60, # 10 mins
-                                              point_threshold = 9,
-                                              distance_threshold = 50) # 50 m
+    with ect.Timer() as t_create_time_filter:
+        dstfsm = dstf.DwellSegmentationTimeFilter(time_threshold=5 * 60,  # 5 mins
+                                                 point_threshold=9,
+                                                 distance_threshold=100)  # 100 m
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/create_time_filter", time.time(), t_create_time_filter.elapsed)
+
+    with ect.Timer() as t_create_dist_filter:
+        dsdfsm = dsdf.DwellSegmentationDistFilter(time_threshold=10 * 60,  # 10 mins
+                                                 point_threshold=9,
+                                                 distance_threshold=50)  # 50 m
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/create_dist_filter", time.time(), t_create_dist_filter.elapsed)
 
     filter_methods = {"time": dstfsm, "distance": dsdfsm}
     filter_method_names = {"time": "DwellSegmentationTimeFilter", "distance": "DwellSegmentationDistFilter"}
+
     # We need to use the appropriate filter based on the incoming data
     # So let's read in the location points for the specified query
-    loc_df = ts.get_data_df("background/filtered_location", time_query)
+    with ect.Timer() as t_get_data_df:
+        loc_df = ts.get_data_df("background/filtered_location", time_query)
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_data_df", time.time(), t_get_data_df.elapsed)
+
     if len(loc_df) == 0:
         # no new segments, no need to keep looking at these again
         logging.debug("len(loc_df) == 0, early return")
         epq.mark_segmentation_done(user_id, None)
         return
 
-    out_of_order_points = loc_df[loc_df.ts.diff() < 0]
-    if len(out_of_order_points) > 0:
-        logging.info("Found out of order points!")
-        logging.info("%s" % out_of_order_points)
-        # drop from the table
-        loc_df = loc_df.drop(out_of_order_points.index.tolist())
-        loc_df.reset_index(inplace=True)
-        # invalidate in the database.
-        out_of_order_id_list = out_of_order_points["_id"].tolist()
-        logging.debug("out_of_order_id_list = %s" % out_of_order_id_list)
-        for ooid in out_of_order_id_list:
-            ts.invalidate_raw_entry(ooid)
+    with ect.Timer() as t_handle_out_of_order:
+        out_of_order_points = loc_df[loc_df.ts.diff() < 0]
+        if len(out_of_order_points) > 0:
+            logging.info("Found out of order points!")
+            logging.info("%s" % out_of_order_points)
+            # drop from the table
+            loc_df = loc_df.drop(out_of_order_points.index.tolist())
+            loc_df.reset_index(inplace=True)
+            # invalidate in the database.
+            out_of_order_id_list = out_of_order_points["_id"].tolist()
+            logging.debug("out_of_order_id_list = %s" % out_of_order_id_list)
+            for ooid in out_of_order_id_list:
+                ts.invalidate_raw_entry(ooid)
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/handle_out_of_order_points", time.time(), t_handle_out_of_order.elapsed)
 
-    filters_in_df = loc_df["filter"].dropna().unique()
+    with ect.Timer() as t_get_filters:
+        filters_in_df = loc_df["filter"].dropna().unique()
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_filters_in_df", time.time(), t_get_filters.elapsed)
+
     logging.debug("Filters in the dataframe = %s" % filters_in_df)
     if len(filters_in_df) == 1:
         # Common case - let's make it easy
-        
-        segmentation_points = filter_methods[filters_in_df[0]].segment_into_trips(ts,
-            time_query)
+        with ect.Timer() as t_segment_trips:
+            segmentation_points = filter_methods[filters_in_df[0]].segment_into_trips(ts, time_query)
+        esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/segment_into_trips", time.time(), t_segment_trips.elapsed)
     else:
-        segmentation_points = get_combined_segmentation_points(ts, loc_df, time_query,
-                                                               filters_in_df,
-                                                               filter_methods)
+        with ect.Timer() as t_get_combined_segmentation:
+            segmentation_points = get_combined_segmentation_points(ts, loc_df, time_query,
+                                                                   filters_in_df,
+                                                                   filter_methods)
+        esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_combined_segmentation_points", time.time(), t_get_combined_segmentation.elapsed)
+
     # Create and store trips and places based on the segmentation points
     if segmentation_points is None:
         epq.mark_segmentation_failed(user_id)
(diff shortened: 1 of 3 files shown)


TeachMeTW:
@shankari  Please review;
- added a timer to the "ad-hoc" pipeline step ✅ 
- add timers to parts of the trip segmentation, with suffixes ✅ 
- tested with a user intake, updates show when querying ✅

shankari:
@TeachMeTW Thank you for the clean PR!

I don't think that the trip segmentations stats are fine-grained enough. 
Did you look at the logs/stats from the "tested with a user intake, updates show when querying"?
My expectation is that the `create_places_and_trips` timer will account for the bulk of the overall time for the step (e.g. if the step takes 100ms, `create_places_and_trips` will account for 95ms). And you don't have any timers within `create_places_and_trips`, so you won't know where we are spending the time, and what we should try to optimize.
Further, I don't think we need timers for simple `if/then` statements - e.g. the below

```
        elif len(segmentation_points) == 0:
            with ect.Timer(ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/early_return_no_segmentation") as timer_early_return_no_segmentation:
                # no new segments, no need to keep looking at these again
                logging.debug("len(segmentation_points) == 0, early return")
                epq.mark_segmentation_done(user_id, None)
            esds.store_pipeline_time(
                user_id,
                ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/early_return_no_segmentation",
                time.time(),
                timer_early_return_no_segmentation.elapsed
            )
```

There are literally two statements within the timer in the code snippet above - a log statement and a `mark_segmentation_done`, neither of which are expected to be heavyweight.

I would:
1. look at the logs from running the single user intake,
2. correlate that to the trip segmentation codebase, and
3. add timers for the sections of code (within `create_places_and_trips`) that are taking the most time.

as a nice-to-have, can you also split commit into two:
- easy: one for adding instrumentation to the `_get_and_store_range`?
- more complex: one for adding instrumentation to trip_segmentation?

You can even create two PRs so I can merge the easy one while reviewing the more complex one.
Feel free to close this PR if that makes it easier!

shankari:
@TeachMeTW is there an ETA for addressing my comments?
I see that you have split the commit, but I still don't see the "easy: one for adding instrumentation to the _get_and_store_range?" PR and you haven't addressed my comment around "I don't think that the trip segmentations stats are fine-grained enough." in this PR.

In particular, "easy: one for adding instrumentation to the _get_and_store_range?" is the main change that we want to get into production soon.

TeachMeTW:
> @TeachMeTW is there an ETA for addressing my comments? I see that you have split the commit, but I still don't see the "easy: one for adding instrumentation to the _get_and_store_range?" PR and you haven't addressed my comment around "I don't think that the trip segmentations stats are fine-grained enough." in this PR.
> 
> In particular, "easy: one for adding instrumentation to the _get_and_store_range?" is the main change that we want to get into production soon.

It is almost finished but I need to test it; is there a uuid in ca_ebike that has trips that are segmentable? How do I search for that? The tests I've done stops at:
```
    if len(loc_df) == 0:
        # no new segments, no need to keep looking at these again
        logging.debug("len(loc_df) == 0, early return")
        epq.mark_segmentation_done(user_id, None)
        return
        
  ```
  
  because there are no new segments hence I cannot see if the fine grained timers are being called.

**Comment by shankari** [comment]
you can reset the pipeline using `bin/reset_pipeline.py`

**Comment by TeachMeTW** [comment]
@shankari I have added more timings, specifically for `create_places_and_trips`

**Comment by TeachMeTW** [comment]
@shankari ready for feedback; is this in depth timers enough; I also only noticed 3 things that take the most times; filter methods, create_places_and_trips, single/multi filter

File: emission/pipeline/intake_stage.py [hunk 1]
```diff
@@ -198,7 +198,13 @@ def run_intake_pipeline_for_user(uuid, skip_if_no_new_data):
...
                                  time.time(), crt.elapsed)
 
-        _get_and_store_range(uuid, "analysis/composite_trip")
+        with ect.Timer() as gsr:
+            logging.info("*" * 10 + "UUID %s: generating store and range " % uuid + "*" * 10)
+            print(str(arrow.now()) + "*" * 10 + "UUID %s: generating store and range " % uuid + "*" * 10)
+            _get_and_store_range(uuid, "analysis/composite_trip")
+
+        esds.store_pipeline_time(uuid, 'GENERATE_STORE_AND_RANGE',
+                                 time.time(), gsr.elapsed)
```
shankari, line 207: @TeachMeTW this is fine. As I suggested, can you please pull this out to a separate PR so that I can accept and merge it immediately? This will unblock us for collecting baseline data for this function before refactoring it next week.

From https://github.com/e-mission/e-mission-server/pull/990#issuecomment-2452811296

> You can even create two PRs so I can merge the easy one while reviewing the more complex one.

or https://github.com/e-mission/e-mission-server/pull/990#issuecomment-2453223948

> In particular, "easy: one for adding instrumentation to the _get_and_store_range?" is the main change that we want to get into production soon.

shankari (changes requested):

emission/pipeline/intake_stage.py
TeachMeTW: @shankari see #991 991

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,68 +51,172 @@ def segment_into_trips(self, timeseries, time_query):
...
+        filter_methods = {"time": dstfsm, "distance": dsdfsm}
+        filter_method_names = {"time": "DwellSegmentationTimeFilter", "distance": "DwellSegmentationDistFilter"}
+        # We need to use the appropriate filter based on the incoming data
+        # So let's read in the location points for the specified query
+    esds.store_pipeline_time(
+        user_id,
+        ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/SEGMENT_CURRENT_TRIPS/setup_filter_methods",
+        time.time(),
+        timer_setup_filter_methods.elapsed
+    )
```
shankari, line 101: > Further, I don't think we need timers for simple if/then statements - e.g. the below

For example, this literally has a timer for creating two dictionaries.

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,68 +51,172 @@ def segment_into_trips(self, timeseries, time_query):
...
+                esds.store_pipeline_time(
+                    user_id,
+                    ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/SEGMENT_CURRENT_TRIPS/mark_segmentation_done",
+                    time.time(),
+                    timer_mark_done.elapsed
+                )
+            except:
+                with ect.Timer() as timer_handle_failure:
+                    logging.exception("Trip generation failed for user %s" % user_id)
+                    epq.mark_segmentation_failed(user_id)
```
shankari, line 206: similarly, this is just a log and a `mark_segmentation` step

shankari (changes requested):
@TeachMeTW no, this doesn't actually address my comments.

> And you don't have any timers within create_places_and_trips, so you won't know where we are spending the time, and what we should try to optimize.

I miswrote; I would actually expect most of the time to be in `segment_into_trips`. I think that is what you found as well.

From https://github.com/e-mission/e-mission-server/pull/990#issuecomment-2453572191
> I also only noticed 3 things that take the most times; filter methods, create_places_and_trips, single/multi filter

what are those times?

I would expect to see fine-grained instrumentation in `segment_into_trips` as well.

**Note also that this has broken several tests.**

As I said earlier, it looks like this will take some time to implement, so please move 
https://github.com/e-mission/e-mission-server/pull/990#discussion_r1827228153 
to a separate PR

TeachMeTW:
@shankari  Addressed comments

shankari:
@TeachMeTW tests are still failing

https://github.com/e-mission/e-mission-server/pull/990#pullrequestreview-2412265795

> Note also that this has broken several tests.

I am not going to review unless the tests are passing.

shankari:
@JGreenlee since this is no longer the weekend, can you review this before it comes to me?

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,54 +51,76 @@ def segment_into_trips(self, timeseries, time_query):
...
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_filters_in_df", time.time(), t_get_filters.elapsed)
+
     logging.debug("Filters in the dataframe = %s" % filters_in_df)
     if len(filters_in_df) == 1:
         # Common case - let's make it easy
-        
-        segmentation_points = filter_methods[filters_in_df[0]].segment_into_trips(ts,
-            time_query)
+        with ect.Timer() as t_segment_trips:
+            segmentation_points = filter_methods[filters_in_df[0]].segment_into_trips(ts, time_query)
```
JGreenlee, line 115: This is an example of a whitespace change I am fine with because you were making a change to this line anyway, and you put it into a less awkward format

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -181,68 +209,90 @@ def create_places_and_trips(user_id, segmentation_points, segmentation_method_na
...
+        with ect.Timer() as t_start_new_chain:
+            last_place = start_new_chain(user_id)
+            last_place.source = segmentation_method_name
+            last_place_entry = ecwe.Entry.create_entry(user_id,
+                                    "segmentation/raw_place", last_place, create_id=True)
+        esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/create_places_and_trips/start_new_chain", time.time(), t_start_new_chain.elapsed)
     else:
         last_place = last_place_entry.data
-
+
```
JGreenlee, line 230: unneeded spaces

JGreenlee (commented):
This looks pretty clean! We should be able to assess bottlenecks in segmentation; I think this will allow for even more granular instrumentation than I had in mind.

TeachMeTW:
@shankari @JGreenlee and I discovered a confusing behavior with the filters. I tested with both an ios and android user -- but the main issue stems from the fact that both only use the distance filter, while the time filter is left unused. What are your thoughts? We are both stumped in regards to this. @JGreenlee can add more to this thread if I have not covered everything.

shankari:
> I tested with both an ios and android user -- but the main issue stems from the fact that both only use the distance filter, while the time filter is left unused. 

I would like to see proof of this. Please use <details> and obscure any identifying information.

JGreenlee:
There is no bug. We were testing against 2 iOS users when we thought it was 1 Android and 1 iOS user.
I was correct that the reason @TeachMeTW wasn't seeing the time filter be used is because he was testing a user on one platform and needed to test the other platform. However, I mixed it up and thought `android` -> `distance` and `ios` -> `time`, when it is actually the opposite.
After choosing an Android user, @TeachMeTW was indeed able to see readings from the time filter.

TeachMeTW:
@JGreenlee Please review the two changes. I do not believe the failing ubuntu test has anything to do with the changes I made.

emission/analysis/intake/segmentation/trip_segmentation_methods/dwell_segmentation_dist_filter.py
```diff
@@ -46,102 +50,201 @@ def segment_into_trips(self, timeseries, time_query):
...
 
-        self.last_ts_processed = None
+        with ect.Timer() as t_initialize:
+            self.last_ts_processed = None
+        esds.store_pipeline_time(
+            user_id,
+            ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/segment_into_trips/initialize_last_ts_processed",
+            time.time(),
+            t_initialize.elapsed
+        )
```
JGreenlee, line 93: This is overkill. Why would this need to be its own step?
I don't think there's any situation where assigning a primitive value is going to take a significant amount of time

shankari: I had also linked to several others earlier (single log, creating a dictionary)

JGreenlee (approved):
I think this is fine to merge.
Some of these Timer blocks might still be overkill, but it's not easy to tell without digging into every function called. We can always remove some later after assessing on prod and seeing what is and isn't a bottleneck.
At this stage, better too granular than not granular enough

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,54 +51,76 @@ def segment_into_trips(self, timeseries, time_query):
         pass
     
 def segment_current_trips(user_id):
-    ts = esta.TimeSeries.get_time_series(user_id)
-    time_query = epq.get_time_range_for_segmentation(user_id)
+    with ect.Timer() as t_get_time_series:
+        ts = esta.TimeSeries.get_time_series(user_id)
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_time_series", time.time(), t_get_time_series.elapsed)
```
shankari, line 56: creating a python object

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,54 +51,76 @@ def segment_into_trips(self, timeseries, time_query):
...
+
+    with ect.Timer() as t_get_time_range:
+        time_query = epq.get_time_range_for_segmentation(user_id)
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/get_time_range_for_segmentation", time.time(), t_get_time_range.elapsed)
```
shankari, line 60: simple database query

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,54 +51,76 @@ def segment_into_trips(self, timeseries, time_query):
...
 
-    dsdfsm = dsdf.DwellSegmentationDistFilter(time_threshold = 10 * 60, # 10 mins
-                                              point_threshold = 9,
-                                              distance_threshold = 50) # 50 m
+    with ect.Timer() as t_create_time_filter:
+        dstfsm = dstf.DwellSegmentationTimeFilter(time_threshold=5 * 60,  # 5 mins
+                                                 point_threshold=9,
+                                                 distance_threshold=100)  # 100 m
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/create_time_filter", time.time(), t_create_time_filter.elapsed)
+
```
shankari, line 70: created python object

emission/analysis/intake/segmentation/trip_segmentation.py
```diff
@@ -47,54 +51,76 @@ def segment_into_trips(self, timeseries, time_query):
...
+    with ect.Timer() as t_create_dist_filter:
+        dsdfsm = dsdf.DwellSegmentationDistFilter(time_threshold=10 * 60,  # 10 mins
+                                                 point_threshold=9,
+                                                 distance_threshold=50)  # 50 m
+    esds.store_pipeline_time(user_id, ecwp.PipelineStages.TRIP_SEGMENTATION.name + "/create_dist_filter", time.time(), t_create_dist_filter.elapsed)
```
shankari, line 75: created python object

shankari (approved):
I agree with @JGreenlee that more fine-grained is better than less, but I also want to point out that there is a cost to storing a lot of stats in terms of database usage. This is particularly true for the pipeline stats which are generated ~  every hour, as opposed to the admin dashboard stats, which are only generated when the admin user logs in to the dashboard.

I can deploy this to staging, but we should use the staging results to strip out the bottom 89-90% of stats before we move to production (even the limited 3 environment production). We can add them back if we resolve all the issues with the top 10-20% of readings!

In particular, I do not anticipate that any of the stats below 👇  will be relevant. They are literally just creating python objects.

PR #991, opened 2024-11-04
added instrumentation to the _get_and_store_range

TeachMeTW:
Separated from #990

### DIFF
--- a/emission/pipeline/intake_stage.py
+++ b/emission/pipeline/intake_stage.py
@@ -198,7 +198,13 @@ def run_intake_pipeline_for_user(uuid, skip_if_no_new_data):
         esds.store_pipeline_time(uuid, ecwp.PipelineStages.CREATE_COMPOSITE_OBJECTS.name,
                                  time.time(), crt.elapsed)
 
-        _get_and_store_range(uuid, "analysis/composite_trip")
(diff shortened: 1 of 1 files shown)


emission/pipeline/intake_stage.py
```diff
@@ -198,7 +198,13 @@ def run_intake_pipeline_for_user(uuid, skip_if_no_new_data):
         esds.store_pipeline_time(uuid, ecwp.PipelineStages.CREATE_COMPOSITE_OBJECTS.name,
                                  time.time(), crt.elapsed)
 
-        _get_and_store_range(uuid, "analysis/composite_trip")
+        with ect.Timer() as gsr:
+            logging.info("*" * 10 + "UUID %s: generating store and range " % uuid + "*" * 10)
+            print(str(arrow.now()) + "*" * 10 + "UUID %s: generating store and range " % uuid + "*" * 10)
+            _get_and_store_range(uuid, "analysis/composite_trip")
+
+        esds.store_pipeline_time(uuid, 'GENERATE_STORE_AND_RANGE',
```
shankari, line 206: I have merged this for now, but I think that this name is incorrect and should be fixed after the current round of optimizations

shankari (commented):
Cleanup change

PR #993, opened 2024-11-08
Enhance `_get_and_store_range` Function to Include Trip Statistics and Last API Call Tracking

TeachMeTW:
### Overview
Enhances `_get_and_store_range` by adding `total_trips`, `labeled_trips`, and `last_call` fields to the user profile, enabling better insights into user activities and API usage.

### Changes

1. **Extended User Profile Updates:**
   - **`total_trips`**: Total number of trips recorded for the user.
   - **`labeled_trips`**: Number of labeled/annotated trips.

2. **Integrated Last API Call Tracking:**
   - Accesses `stats/server_api_time` to retrieve timestamps of the last `GET` and `PUT` API calls.
   - Determines the latest call and stores it as `last_call` in the user's profile in UNIX format.

### Testing

1. Picked a user, cleared pipeline, re-ran pipeline, verified user profile was updated.

**Please review and provide feedback or approval.**

### DIFF
--- a/emission/analysis/result/user_stat.py
+++ b/emission/analysis/result/user_stat.py
@@ -1,0 +1,112 @@
+# emission/analysis/result/user_stats.py
+
+import logging
+import pymongo
+import arrow
+from typing import Optional, Dict, Any
+import emission.storage.timeseries.abstract_timeseries as esta
+import emission.core.wrapper.user as ecwu
+
+TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss'
+
+def count_trips(ts: esta.TimeSeries, key_list: list, extra_query_list: Optional[list] = None) -> int:
+    """
+    Counts the number of trips based on the provided query.
+
+    :param ts: The time series object.
+    :type ts: esta.TimeSeries
+    :param key_list: List of keys to filter trips.
+    :type key_list: list
+    :param extra_query_list: Additional queries, defaults to None.
+    :type extra_query_list: Optional[list], optional
+    :return: The count of trips.
+    :rtype: int
+    """
+    count = ts.find_entries_count(key_list=key_list, extra_query_list=extra_query_list)
+    logging.debug(f"Counted {len(key_list)} trips with additional queries {extra_query_list}: {count}")
+    return count
+
+
+def get_last_call_timestamp(ts: esta.TimeSeries) -> Optional[int]:
+    """
+    Retrieves the last API call timestamp.
+
+    :param ts: The time series object.
+    :type ts: esta.TimeSeries
+    :return: The last call timestamp or None if not found.
+    :rtype: Optional[int]
+    """
+    last_call_ts = ts.get_first_value_for_field(
+        key='stats/server_api_time',
+        field='data.ts',
+        sort_order=pymongo.DESCENDING
+    )
+    logging.debug(f"Last call timestamp: {last_call_ts}")
+    return None if last_call_ts == -1 else last_call_ts
+
+
+def update_user_profile(user_id: str, data: Dict[str, Any]) -> None:
+    """
+    Updates the user profile with the provided data.
+
+    :param user_id: The UUID of the user.
+    :type user_id: str
+    :param data: The data to update in the user profile.
+    :type data: Dict[str, Any]
+    :return: None
+    """
+    user = ecwu.User.fromUUID(user_id)
+    user.update(data)
+    logging.debug(f"User profile updated with data: {data}")
+    logging.debug(f"New profile: {user.getProfile()}")
+
+
+def get_and_store_user_stats(user_id: str, trip_key: str) -> None:
+    """
+    Aggregates and stores user statistics into the user profile.
+
+    :param user_id: The UUID of the user.
+    :type user_id: str
+    :param trip_key: The key representing the trip data in the time series.
+    :type trip_key: str
+    :return: None
+    """
+    try:
+        logging.info(f"Starting get_and_store_user_stats for user_id: {user_id}, trip_key: {trip_key}")
+
+        ts = esta.TimeSeries.get_time_series(user_id)
+        start_ts_result = ts.get_first_value_for_field(trip_key, "data.start_ts", pymongo.ASCENDING)
+        start_ts = None if start_ts_result == -1 else start_ts_result
+
+        end_ts_result = ts.get_first_value_for_field(trip_key, "data.end_ts", pymongo.DESCENDING)
+        end_ts = None if end_ts_result == -1 else end_ts_result
+
+        total_trips = count_trips(ts, key_list=["analysis/confirmed_trip"])
+        labeled_trips = count_trips(
+            ts,
+            key_list=["analysis/confirmed_trip"],
+            extra_query_list=[{'data.user_input': {'$ne': {}}}]
+        )
+
+        logging.info(f"Total trips: {total_trips}, Labeled trips: {labeled_trips}")
+        logging.info(f"user_id type: {type(user_id)}")
+
+        last_call_ts = get_last_call_timestamp(ts)
+        logging.info(f"Last call timestamp: {last_call_ts}")
+
+        update_data = {
+            "pipeline_range": {
+                "start_ts": start_ts,
+                "end_ts": end_ts
+            },
+            "total_trips": total_trips,
+            "labeled_trips": labeled_trips,
+            "last_call_ts": last_call_ts
+        }
+
+        update_user_profile(user_id, update_data)
+
+        logging.debug("User profile updated successfully.")
+
+    except Exception as e:
+        logging.error(f"Error in get_and_store_user_stats for user_id {user_id}: {e}")
--- a/emission/analysis/result/user_stat.py
+++ b/emission/analysis/result/user_stat.py
@@ -7,26 +7,6 @@ from typing import Optional, Dict, Any
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.core.wrapper.user as ecwu
 
-TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss'
-
-def count_trips(ts: esta.TimeSeries, key_list: list, extra_query_list: Optional[list] = None) -> int:
-    """
-    Counts the number of trips based on the provided query.
-
-    :param ts: The time series object.
-    :type ts: esta.TimeSeries
-    :param key_list: List of keys to filter trips.
-    :type key_list: list
-    :param extra_query_list: Additional queries, defaults to None.
-    :type extra_query_list: Optional[list], optional
-    :return: The count of trips.
-    :rtype: int
-    """
-    count = ts.find_entries_count(key_list=key_list, extra_query_list=extra_query_list)
-    logging.debug(f"Counted {len(key_list)} trips with additional queries {extra_query_list}: {count}")
-    return count
-
-
 def get_last_call_timestamp(ts: esta.TimeSeries) -> Optional[int]:
     """
     Retrieves the last API call timestamp.
@@ -81,9 +61,8 @@ def get_and_store_user_stats(user_id: str, trip_key: str) -> None:
         end_ts_result = ts.get_first_value_for_field(trip_key, "data.end_ts", pymongo.DESCENDING)
         end_ts = None if end_ts_result == -1 else end_ts_result
 
-        total_trips = count_trips(ts, key_list=["analysis/confirmed_trip"])
-        labeled_trips = count_trips(
-            ts,
+        total_trips = ts.find_entries_count(key_list=["analysis/confirmed_trip"])
+        labeled_trips = ts.find_entries_count(
             key_list=["analysis/confirmed_trip"],
             extra_query_list=[{'data.user_input': {'$ne': {}}}]
         )
--- a/emission/pipeline/intake_stage.py
+++ b/emission/pipeline/intake_stage.py
@@ -40,6 +40,7 @@ import emission.net.ext_service.habitica.executor as autocheck
 import emission.storage.decorations.stats_queries as esds
 
 import emission.core.wrapper.user as ecwu
+import emission.analysis.result.user_stat as eaurs
 
 def run_intake_pipeline(process_number, uuid_list, skip_if_no_new_data=False):
     """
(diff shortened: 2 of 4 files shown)


JGreenlee (changes requested):

emission/pipeline/intake_stage.py
```diff
@@ -207,14 +208,92 @@ def run_intake_pipeline_for_user(uuid, skip_if_no_new_data):
...
+
+        # Retrieve trip entries as an iterator
+        trip_entries = ts.find_entries([trip_key], time_query=None)
+
+        # Iterate through trip_entries once to count total_trips and labeled_trips
+        for trip in trip_entries:
+            total_trips += 1
+            if trip.get('data', {}).get('user_input'):
+                labeled_trips += 1
+        logging.info(f"Total trips: {total_trips}, Labeled trips: {labeled_trips}")
```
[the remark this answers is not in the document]
TeachMeTW: @JGreenlee I get this error:
`extra query would overwrite keys ['user_id']` using op-admins implementation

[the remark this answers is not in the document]
JGreenlee: You can remove the `user_id` filter because `ts` here is already the timeseries **for that user**

TeachMeTW:
@JGreenlee Comments addressed, I split the function into a different file as suggested -- though it may be overkill. Seeking thoughts

JGreenlee (commented):
Good refactor! I agree that the smaller functions are overkill.
Sometimes it's just a matter of finding the balance between tidiness and complexity.

TeachMeTW:
@JGreenlee Comments addressed

emission/analysis/result/user_stat.py
```diff
@@ -0,0 +1,111 @@
...
+        logging.info(f"Last call timestamp: {last_call_ts}")
+
+        update_data = {
+            "pipeline_range": {
+                "start_ts": start_ts,
+                "end_ts": end_ts
+            },
+            "total_trips": total_trips,
+            "labeled_trips": labeled_trips
+        }
```
JGreenlee, line 104: Are we forgetting to store `last_call_ts` ?

TeachMeTW:
@JGreenlee Addressed and rebased

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self.assertEqual(pipeline_range["start_ts"], expected_start_ts,
+                         f"Expected start_ts to be {expected_start_ts}, got {pipeline_range['start_ts']}")
+        self.assertEqual(pipeline_range["end_ts"], expected_end_ts,
+                         f"Expected end_ts to be {expected_end_ts}, got {pipeline_range['end_ts']}")
+
+        # Verify last_call_ts
+        expected_last_call_ts = -
+        self.assertEqual(profile["last_call_ts"], expected_last_call_ts,
+                         f"Expected last_call_ts to be {expected_last_call_ts}, got {profile['last_call_ts']}")
```
JGreenlee, line 127: If we want to validate last_call_ts getting filled in, we will need to pretend we accessed the API. I think you can do this by calling `after_request()` from `emission.net.api.cfc_webapp`

I recommend extracting this to a different function `test_last_call_ts`.

After calling `after_request()`, `runIntakePipeline` again. Then get the profile and `last_call_ts` should be filled in with the time at which you called `after_request()`

TeachMeTW:
@JGreenlee I realized my code was NOT pushed and you may have reviewed the wrong thing.... my bad, I will take your notes into account with the latest.

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,218 @@
...
+        self.assertIsNotNone(profile, "User profile should exist after storing stats.")
+
+        # Verify that the expected fields are present
+        self.assertIn("total_trips", profile, "User profile should contain 'total_trips'.")
+        self.assertIn("labeled_trips", profile, "User profile should contain 'labeled_trips'.")
+        self.assertIn("pipeline_range", profile, "User profile should contain 'pipeline_range'.")
+        self.assertIn("last_call_ts", profile, "User profile should contain 'last_call_ts'.")
+
+        expected_total_trips = -
+        expected_labeled_trips = -
```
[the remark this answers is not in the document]
TeachMeTW: @JGreenlee load any user inputs how so? Like loading from the example or what?

TeachMeTW:
@JGreenlee Comments addressed, ready for re-review

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,141 @@
...
+import emission.tests.common as etc
+
+
+class TestUserStats(unittest.TestCase):
+    def setUp(self):
+        """
+        Set up the test environment by loading real example data for both Android and  users.
+        """
+        # Configure logging for the test
+        etc.configLogging()
```
JGreenlee, line 35: can you move this to the end of file below `if __name__ == '__main__':` ?

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,141 @@
...
+            self.fail("No users found in the database to perform tests.")
+        test_user_id = get_example.iloc[0].uuid
+        self.testUUID = test_user_id
+        self.UUID = test_user_id
+        # Load example entries from a JSON file
+        with open("emission/tests/data/real_examples/shankari_2015-aug-27") as fp:
+            self.entries = json.load(fp, object_hook=esj.wrapped_object_hook)
+
+        # Set up the real example data with entries
+        etc.setupRealExampleWithEntries(self)
```
JGreenlee, line 49: Just calling `setupRealExample` (not `setupRealExampleWithEntries`) will do all of this for you

It will create one UUID for you, assigned to `self.testUUID`. Use that instead of `self.UUID`

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Retrieve the user profile
+        profile = edb.get_profile_db().find_one({"user_id": self.UUID})
+        if profile is None:
+            # Initialize the profile if it does not exist
+            edb.get_profile_db().insert_one({"user_id": self.UUID})
```
JGreenlee, line 55: What happens if you take this out? Does something not work without it?

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,141 @@
...
+        pipeline_db = edb.get_pipeline_state_db()
+        pipeline_db.delete_many({"user_id": self.UUID})
+
+        # Delete all analysis time series entries for  users
+        analysis_ts_db = edb.get_analysis_timeseries_db()
+        analysis_ts_db.delete_many({"user_id": self.UUID})
+
+        # Delete user profiles
+        profile_db = edb.get_profile_db()
+        profile_db.delete_one({"user_id": self.UUID})
```
JGreenlee, line 80: This can be more concise, no need to overexplain. I would simply do:

```python
edb.get_timeseries_db().delete_many({"user_id": self.UUID})
edb.get_pipeline_state_db().delete_many({"user_id": self.UUID})
edb.get_analysis_timeseries_db().delete_many({"user_id": self.UUID})
edb.get_profile_db().delete_one({"user_id": self.UUID})
```

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def testGetAndStoreUserStats(self):
```
JGreenlee, line 82: this test looks good

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -0,0 +1,141 @@
...
+        expected_end_ts = 1440729142.709
+
+        self.assertEqual(pipeline_range["start_ts"], expected_start_ts,
+                         f"Expected start_ts to be {expected_start_ts}, got {pipeline_range['start_ts']}")
+        self.assertEqual(pipeline_range["end_ts"], expected_end_ts,
+                         f"Expected end_ts to be {expected_end_ts}, got {pipeline_range['end_ts']}")
+
+    def testLastCall(self):
+        # Call the function with all required arguments
+        enac.store_server_api_time(self.UUID, "test_call", 1440729142.709, 69)
```
JGreenlee, line 123: This ts could be anything, but using the same ts as the end_ts could be confusing to people in the future.
You could grab the current time and call it `test_call_ts`

```suggestion
        test_call_ts = time.time()
        enac.store_server_api_time(self.UUID, "test_call", test_call_ts)
```


Then you can also use `test_call_ts` below instead of declaring `expected_last_call_ts`

JGreenlee (changes requested):
Almost there!

emission/tests/analysisTests/intakeTests/TestUserStat.py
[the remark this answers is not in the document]
JGreenlee: The `real_examples` files like `shankari_2015-aug-27` only have raw sensed data; i.e. what the phone automatically picked while traveling

For some of the days, e.g. `shankari_2016-06-20` we have an additional file `shankari_2016-06-20.user_inputs`, which does not contain sensed data but it has "manual inputs" or "user inputs". These entries represent labels or survey responses that the user has recorded while using the app.

For a more in-depth test you could load both `shankari_2016-06-20` and `shankari_2016-06-20.user_inputs` and `expected_labeled_trips` would update

emission/tests/analysisTests/intakeTests/TestUserStat.py
TeachMeTW: The setup or  just `enc.configLogging()` ?

emission/tests/analysisTests/intakeTests/TestUserStat.py
TeachMeTW: Noted, that does seem way simpler

emission/tests/analysisTests/intakeTests/TestUserStat.py
TeachMeTW: When I take this out, the profile is empty.
```
  File "/Users/rsimpson/e-mission-server/emission/tests/analysisTests/intakeTests/TestUserStat.py", line 81, in testGetAndStoreUserStats
    self.assertIsNotNone(profile, "User profile should exist after storing stats.")
AssertionError: unexpectedly None : User profile should exist after storing stats.
```

Meaning for some reason the entry is not being inserted. Does that hint to an issue with my additions or existing?

TeachMeTW:
@JGreenlee Addressed, see response -- in need of clarification

emission/tests/analysisTests/intakeTests/TestUserStat.py
JGreenlee: I just meant configLogging 
The other tests have it at the bottom (at least the majority)

emission/analysis/result/user_stat.py
```diff
@@ -0,0 +1,112 @@
...
+import logging
+import pymongo
+import arrow
+from typing import Optional, Dict, Any
+import emission.storage.timeseries.abstract_timeseries as esta
+import emission.core.wrapper.user as ecwu
+
+TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss'
+
+def count_trips(ts: esta.TimeSeries, key_list: list, extra_query_list: Optional[list] = None) -> int:
```
JGreenlee, line 12: nitpick: I had an earlier comment here that was resolved but I don't think was addressed. I don't see the point of this function when it is just a wrapper around `ts.find_entries_count` with the exact same params and no added functionality 

In the event that a suggested change doesn't work for you or you have a good reason to do otherwise, that's fine; just reply and we can discuss

Or, if you just missed my comment, make sure not to resolve conversations until you have made a change that addresses them. When I have my code reviewed I like to react with thumbs up and reply "TODO" so I can keep track of what I have seen vs what I have actually implemented and pushed up

JGreenlee (changes requested):
@TeachMeTW Code looks good but the new tests appear to be failing in GH Actions 
Do they pass for you locally?

emission/analysis/result/user_stat.py
TeachMeTW: I think I did address it, but I've been trying to rebase and do all this github branch magic and maybe it got reverted -- will do react and/or reply on future comments before resolving

OR I believe a better option is for the reviewer to resolve the comment? what are your thoughts on that approach in addition?

JGreenlee:
I don't know why it's passing locally and not here, but it is related to this PR because the test that's failing is the one you added. It's complaining about timestamps that look a little bit off.

TeachMeTW:
> I don't know why it's passing locally and not here, but it is related to this PR because the test that's failing is the one you added. It's complaining about timestamps that look a little bit off.

Shall I add delta/assert if almost equal?

JGreenlee:
> > I don't know why it's passing locally and not here, but it is related to this PR because the test that's failing is the one you added. It's complaining about timestamps that look a little bit off.
> 
> Shall I add delta/assert if almost equal?

No we should investigate why it's different.
The pipeline is supposed to be deterministic, i.e. 100% reproducible 

What happens if you checkout master and run that test data through the pipeline?
Is it the same as the CI or the same as you got on your branch?

TeachMeTW:
@JGreenlee  I checkout master and run the test data through pipeline and its the same as the branch NOT the CI. I believe running ALL the tests must affect the ts. I tried to test runAllTest.sh but it crashes my console/vscode, perhaps you could verify

JGreenlee:
I will try to verify tomorrow. I will also pass this along in the review process for Shankari's input.

If `runAllTests` performs differently than running an individual test file, it is concerning as it means we do not have proper isolation between the tests. Maybe one of the tests does not `tearDown` properly. It may be difficult to track down the culprit.

JGreenlee:
If we can confirm that the issue is with one of the other tests, this looks good to me to merge.

shankari:
> I tried to test runAllTest.sh but it crashes my console/vscode, perhaps you could verify

You should resolve the crash and then test. We already have one CI failing for "unknown reasons", having the second also fail makes the CI not be meaningful.

TeachMeTW:
@shankari  @JGreenlee  

I discovered two things.

### Crashing
The first one in regards to the crashing runAllTests.sh

When the terminal started crashing during script execution, I redirected the output to a log file to figure out what was going on. That’s when I saw the error: ValueError: I/O operation on closed file. Initially, the code opened the file without using a with statement and didn’t explicitly close it causing a different ValueError where the file was never closed. This was causing unpredictable behavior, as the file could be closed unexpectedly by the system or Python's garbage collector.

I updated the code to use a with statement to ensure the file was properly closed. However, the error persisted because there was still a redundant line after the with block trying to read the file, even though it had already been closed. Once I removed that redundant line and ensured the file was read entirely within the with block, the issue was resolved, and the terminal stopped crashing.

I could make a new issue in regards to this fix.

### Timing differences

There is indeed a difference with single testing and runAll. The single test has the ts normal but the runAll brings the same error as the gitlab runner where it is off. I thought fixing the crash would fix this discrepancy but it seems deeper than that.

shankari:
> My hunch is that a teardown is broken somewhere and this UUID is being reused. Perhaps I can use a different uuid as a bandaid fix or can dig deeper into the root cause.

Feel free to use either option. I will not merge until this is resolved.

JGreenlee:
This specific problem has happened before.

I used Github search to search the `e-mission` org for the problematic timestamp and got exactly 1 hit https://github.com/e-mission/e-mission-server/pull/509#issuecomment-367734717

Reading through that PR, we cannot pinpoint what the cause was.

TeachMeTW:
Commenting out all the test cases in TestTripSegmentation and replacing it with a dummy test with a print statment still invokes the same errors. Might be something with setup and teardown.

TeachMeTW:
However, if one runs `TestTripSegmentation` or `TestSectionSegmentation` **AFTER** `TestGetUserStats` there is **NO ERROR**

TeachMeTW:
Alternatively, using `emission/tests/data/real_examples/shankari_2015-aug-21` rather than `emission/tests/data/real_examples/shankari_2015-aug-27` resulted in **NO ERRORS** when running all.

image

<br></br>

We have a couple options. 

1. We could do above and use Aug 21 dataset for UserStats, call it a day for now.
2. We could dig deeper as to why this is happening.

Thoughts?

shankari:
We should do both.
We should change the test to Aug 21 for UserStats so that we can merge this and continue to make progress.
We should then file a new issue to dig deeper into why this is happening

TeachMeTW:
@shankari Modified and pushed accordingly. Will make an issue as suggested.

JGreenlee:
> Will make an issue as suggested.

@TeachMeTW I figured out what was causing it and made a PR https://github.com/e-mission/e-mission-server/pull/1003

TeachMeTW:
Nice find!

shankari (approved):
I am going to roll forward and merge to unblock us; @TeachMeTW please address these comments in a follow-on PR

shankari:
@TeachMeTW there is now a merge conflict after I have pulled from master.
Screenshot 2024-12-21 at 4 58 28 PM

I think this may be because your original base did not include 
https://github.com/e-mission/e-mission-server/commit/2a327421d20b4dd5e3989cdd042df9e645bd1707

You need to pull changes from master periodically, particularly when you are making changes to the same files

shankari:
@JGreenlee @TeachMeTW this needs to be refactored further for it to work properly.
Note that we have an early return if there is no new data, so this code is effectively in the "else"

```
        if skip_if_no_new_data and new_entry_count == 0:
            print("No new entries, and skip_if_no_new_data = %s, skipping the rest of the pipeline" % skip_if_no_new_data)
            return
        else:
            print("New entry count == %s and skip_if_no_new_data = %s, continuing" %
                (new_entry_count, skip_if_no_new_data))
```

That's fine as long as we were only caching pipeline state, because it wouldn't change if the pipeline was skipped
But now that we are caching the `last_call_ts` the current location will not work.
Concretely, if the app is contacting the server but is not uploading any data, `last_call_ts` will never be changed.

shankari:
I have fixed this in #1005, @TeachMeTW please update the unit tests as outlined there

emission/analysis/result/user_stat.py
```diff
@@ -0,0 +1,91 @@
...
+        ts = esta.TimeSeries.get_time_series(user_id)
+        start_ts_result = ts.get_first_value_for_field(trip_key, "data.start_ts", pymongo.ASCENDING)
+        start_ts = None if start_ts_result == -1 else start_ts_result
+
+        end_ts_result = ts.get_first_value_for_field(trip_key, "data.end_ts", pymongo.DESCENDING)
+        end_ts = None if end_ts_result == -1 else end_ts_result
+
+        total_trips = ts.find_entries_count(key_list=["analysis/confirmed_trip"])
+        labeled_trips = ts.find_entries_count(
+            key_list=["analysis/confirmed_trip"],
```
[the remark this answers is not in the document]
TeachMeTW: will be addressing in follow up pr; no reason it should be hard coded

PR #997, opened 2024-11-23
Add Timing Instrumentation to `has_trip_ended` Function

TeachMeTW:
## Summary

Adds timing instrumentation to the `has_trip_ended` function, improving observability and allowing for detailed performance monitoring of key operations.

## Changes

1. **Timing Instrumentation**:
   - Added timing measurements for the following operations within the `has_trip_ended` function:
     - **Time Calculations**: Measures the computation time for variables like `timeDelta`, `distDelta`, and `speedDelta`.
     - **Motion Check**: Tracks the time taken to evaluate ongoing motion in the specified time range.
     - **Distance Threshold Check**: Times the evaluation of `last5MinsDistances` and `last10PointsDistances` against their thresholds.


TeachMeTW:
Was part of the bottom 80%, no longer needed

PR #999, opened 2024-12-09
Fix `DB_RESULT_LIMIT` Parsing to Allow Integer Configuration

TeachMeTW:
## Summary

Fixes an issue where the `DB_RESULT_LIMIT` configuration was being treated as a string, preventing it from being set as an integer dynamically. The update ensures the value is properly parsed as an integer, enabling correct configuration usage.

## Changes

- Ensures `DB_RESULT_LIMIT` is parsed as an integer.
- Retains the default value of `250000` if the configuration is not explicitly set.

## Why This Fix?

- Previously, the value of `DB_RESULT_LIMIT` was treated as a string, leading to unexpected behavior when attempting to use or override it.
- This change resolves the type mismatch and ensures the value can be correctly parsed and used as an integer.

## Testing
- Tested with no `DB_RESULT_LIMIT` set: defaults to `250000`.
- Tested with `DB_RESULT_LIMIT` set to a valid integer: properly reflects the configured value.
- Tested with invalid `DB_RESULT_LIMIT` (non-numeric string): verified that it raises an appropriate error -- defaults to 250,000 if error

image

### DIFF
--- a/emission/core/get_database.py
+++ b/emission/core/get_database.py
@@ -23,7 +23,7 @@ for key in ["DB_HOST", "DB_RESULT_LIMIT"]:
         db_config[key] = None
 print("Retrieved config: %s" % db_config)
 url = config.get("DB_HOST", "localhost")
-result_limit = config.get("DB_RESULT_LIMIT", 250000)
+result_limit = int(config.get("DB_RESULT_LIMIT", 250000))
 
 try:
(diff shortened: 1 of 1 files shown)


JGreenlee (approved):
LGTM!

shankari:
LTGM! As general feedback, it would be good to have screenshots of all the tests that you did, maybe in a table, but I am not going to hold up the release for that.

PR #1003, opened 2024-12-19
clear AND reload analysis config on tests’ tearDown

JGreenlee:
There are several tests that call `etc.set_analysis_config` in their `setUp()` to set config options in certain testing scenarios. This creates a config file, and in tearDown of each test the config file is deleted via os.remove.

However, just deleting the file is not enough!

When emission.analysis.config is intialized, we read the config and cache it in a variable `config_data` https://github.com/e-mission/e-mission-server/blob/4034533bfc95dbd7c976146341acd3527f9bc7c9/emission/analysis/config.py#L19 When get_config() is called while the pipeline is running, it returns the value of `config_data`, which can be out of sync with the contents of the config file if the config file was added/removed/modified and `reload_config()` was not called. So instead of just calling `os.remove()` these tests also need to call `reload_config()` in their `tearDown()`.

Added this by way of a new function in `emission.tests.common`

Also, I made the config file pathnames into constants for tidiness and to eliminate the risk of typo-related bugs

### DIFF
--- a/emission/analysis/config.py
+++ b/emission/analysis/config.py
@@ -1,17 +1,21 @@
 import json
 import os
 
+ANALYSIS_CONF_PATH = "conf/analysis/debug.conf.json"
+ANALYSIS_CONF_PROD_PATH = "conf/analysis/debug.conf.prod.json"
+ANALYSIS_CONF_DEV_PATH = "conf/analysis/debug.conf.dev.json"
+
 def get_config_data():
     try:
         print("Trying to open debug.conf.json")
-        config_file = open('conf/analysis/debug.conf.json')
+        config_file = open(ANALYSIS_CONF_PATH)
     except:
         if os.getenv("PROD_STAGE") == "TRUE":
             print("In production environment, config not overridden, using default production debug.conf")
-            config_file = open('conf/analysis/debug.conf.prod.json')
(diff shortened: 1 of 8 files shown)


shankari:
@JGreenlee I have merged this since it is clearly the right thing to do, but I don't understand how the analysis config would affect the pipeline timestamps, which is what you reported in 

https://github.com/e-mission/e-mission-server/pull/993#issuecomment-2549462483
> We do not understand why the pipeline start ts comes up as 1440688707.867 during runAllTests.sh.

Is the pipeline not run properly because of the configs? But in that case, the pipeline range should not be set at all (it should be like a new user), right? Can you explain this further?

JGreenlee:
> Is the pipeline not run properly because of the configs? But in that case, the pipeline range should not be set at all (it should be like a new user), right? Can you explain this further?

The pipeline runs properly but behaves differently depending on `"intake.cleaning.filter_accuracy.enable"` in the analysis config.
When `True`, the first trip on Aug 27 starts at `1440688707.867`.
When `False`, that trip starts a bit later at `1440688739.672`.

When the test was run in isolation, `"intake.cleaning.filter_accuracy.enable"` was `False`.
When the test was run during runAllTests.sh, it was `True`.

PR #1004, opened 2024-12-20
fix EnvironmentLocationNotFound error during test-with-manual-install

JGreenlee:
"Teardown the test environment" has been consistently failing for the last month or so.
```
Run source setup/teardown_tests.sh
Removing environment from
CondaError: Run 'conda init' before 'conda deactivate'
EnvironmentLocationNotFound: Not a conda environment: /usr/share/miniconda/envs/emissiontest
Error: Process completed with exit code 1.
```

I noticed that test-with-manual-install.yml follows a similar teardown procedure (with 'emission' as the environment instead of 'emissiontest') I saw that test-with-manual-install.yml invokes activates_conda.sh again right before calling the teardown script

### DIFF
--- a/.github/workflows/test-with-manual-install.yml
+++ b/.github/workflows/test-with-manual-install.yml
@@ -70,4 +70,6 @@ jobs:
 
     - name: Teardown the test environment
       shell: bash -l {0}
-      run: source setup/teardown_tests.sh
+      run: |
+        source setup/activate_conda.sh
+        source setup/teardown_tests.sh


shankari (approved):

shankari:
@JGreenlee thanks so much for handling this! The failure was annoying me too, and I took a brief look at it, but couldn't immediately figure out why the environment was not being deleted given that it had just been used to run the tests. Not activating will do that!

It was also confusing that this broke although the last change to the file was from Apr 2023. 

```
$ git log .github/workflows/test-with-manual-install.yml
commit 6d17182bfc3089dccbcf62edbd1a4ac227b5c7dd
Merge: 54721a19 022182f7
Author: Shankari <shankari@eecs.berkeley.edu>
Date:   Mon Apr 17 10:40:20 2023 -0700
```

I suspect that what happened is that the default version of conda on the github runner changed so that it no longer supports `conda env`. Before that change, since we didn't call activate, we used the default conda. The old default conda appeared to treat deactivating a nonexisting environment as a NOP, the new default conda appears to fail.

PR #1005, opened 2024-12-22
Always run the computation of the user profile stats

shankari:
In https://github.com/e-mission/e-mission-server/pull/993 we expanded `_get_and_store_range` to store entries other than the pipeline range, notably the `last_call_ts`.

However, adding it to the end of the pipeline causes it to be skipped most of the time.

We have an early return from `run_intake_pipeline_for_user` if there is no new data, so the end of the pipeline (the code in #993) is effectively in the "else"

That's fine as long as we were only caching pipeline state, because it wouldn't change if the pipeline was skipped
But now that we are caching the last_call_ts the current location will not work.  Concretely, if the app is contacting the server but is not uploading any data, last_call_ts will never be changed.

Since this is a fix that I am implementing on a holiday weekend, I have not added unit tests to check this. @TeachMeTW please enhance the unit tests to verify that the stats are generated in both cases: (i) when there is new data and (ii) when there is no new data.

### DIFF
--- a/emission/pipeline/intake_stage.py
+++ b/emission/pipeline/intake_stage.py
@@ -76,6 +76,12 @@ def run_intake_pipeline(process_number, uuid_list, skip_if_no_new_data=False):
 
         try:
             run_intake_pipeline_for_user(uuid, skip_if_no_new_data)
+            with ect.Timer() as gsr:
+                logging.info("*" * 10 + "UUID %s: storing user stats " % uuid + "*" * 10)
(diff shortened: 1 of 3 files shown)


shankari:
Tested on staging, stats should be computed now.

```
2024-12-22T05:09:52.399086+00:00**********UUID 45c15db9-97a0-4162-a633-82668100aba8: moving to long term**********
No new entries, and skip_if_no_new_data = True, skipping the rest of the pipeline
2024-12-22T05:09:53.252319+00:00**********UUID 45c15db9-97a0-4162-a633-82668100aba8: storing user stats **********
```

PR #1006, opened 2024-12-24
Fix for User Stats Pipeline to Handle Both Scenarios (New and No New Data)

TeachMeTW:
## Summary
This PR addresses a fix to ensure that the user stats pipeline behaves as expected in two specific scenarios:
1. **When no new data is added**: The stats should remain unchanged after re-running the pipeline.
2. **When new data is added**: The stats should update correctly to reflect the new data.

While implementing the fix, I identified the need for unit tests to ensure these behaviors are consistently verified.

## Changes Made
- Added checks for both scenarios to improve pipeline reliability and prevent unexpected stat updates.

## Unit Tests
1. **Scenario (i):** Stats remain unchanged when the pipeline is re-run without new data.
2. **Scenario (ii):** Stats update correctly when new data is added, ensuring consistency with expected values.

The logic for these tests can be based on the example methods provided:
- `testGetAndStoreUserStatsSecondRunNoNewData`: Verifies no changes to stats when re-run without new data.
- `testGetAndStoreUserStatsNewData`: Verifies correct updates when new data is added.

See #1005

### DIFF
--- a/emission/tests/analysisTests/intakeTests/TestUserStat.py
+++ b/emission/tests/analysisTests/intakeTests/TestUserStat.py
@@ -57,11 +57,20 @@ class TestUserStats(unittest.TestCase):
         edb.get_analysis_timeseries_db().delete_many({"user_id": self.testUUID})
         edb.get_profile_db().delete_one({"user_id": self.testUUID})
 
-    def testGetAndStoreUserStats(self):
+    def testGetAndStoreUserStatsDefault(self):
         """
-        Test get_and_store_user_stats for the user to ensure that user statistics
+        Case (i): Test get_and_store_user_stats for the user to ensure that user statistics
         are correctly aggregated and stored in the user profile.
         """
+        # Load in new data
+        
+        with open("emission/tests/data/real_examples/shankari_2015-aug-27") as fp:
+            self.entries = json.load(fp, object_hook=esj.wrapped_object_hook)
+            
+        etc.setupRealExampleWithEntries(self)
+
+        # Rerun Pipeline
+        etc.runIntakePipeline(self.testUUID)
 
         # Retrieve the updated user profile from the database
         profile = edb.get_profile_db().find_one({"user_id": self.testUUID})
@@ -75,7 +84,7 @@ class TestUserStats(unittest.TestCase):
         self.assertIn("pipeline_range", profile, "User profile should contain 'pipeline_range'.")
         self.assertIn("last_call_ts", profile, "User profile should contain 'last_call_ts'.")
 
-        expected_total_trips = 5
+        expected_total_trips = 13
         expected_labeled_trips = 0
 
         self.assertEqual(profile["total_trips"], expected_total_trips,
@@ -89,15 +98,58 @@ class TestUserStats(unittest.TestCase):
         self.assertIn("end_ts", pipeline_range, "Pipeline range should contain 'end_ts'.")
 
         expected_start_ts = 1440168891.095
-        expected_end_ts = 1440209488.817
+        expected_end_ts = 1440729142.709
 
         self.assertEqual(pipeline_range["start_ts"], expected_start_ts,
                          f"Expected start_ts to be {expected_start_ts}, got {pipeline_range['start_ts']}")
         self.assertEqual(pipeline_range["end_ts"], expected_end_ts,
                          f"Expected end_ts to be {expected_end_ts}, got {pipeline_range['end_ts']}")
+        
+        test_call_ts = time.time()
+        enac.store_server_api_time(self.testUUID, "test_call_ts", test_call_ts, 69420)
+        etc.runIntakePipeline(self.testUUID)
 
-    def testLastCall(self):
-        # Call the function with all required arguments
+        # Retrieve the profile from the database
+        profile = edb.get_profile_db().find_one({"user_id": self.testUUID})
+
+        # Verify that last_call_ts is updated correctly
+        expected_last_call_ts = test_call_ts
+        actual_last_call_ts = profile.get("last_call_ts")
+
+        self.assertEqual(
+            actual_last_call_ts,
+            expected_last_call_ts,
+            f"Expected last_call_ts to be {expected_last_call_ts}, got {actual_last_call_ts}"
+        )
+
+    def testGetAndStoreUserStatsSecondRunNoNewData(self):
+        """
+        Case (ii): Verify stats remain unchanged if we run the pipeline again
+        without adding new data.
+        """
+        # Check stats after the initial run (from setUp()).
+        initial_profile = edb.get_profile_db().find_one({"user_id": self.testUUID})
+        self.assertIsNotNone(initial_profile, "User profile should exist after first run.")
+        initial_total_trips = initial_profile["total_trips"]
+        initial_labeled_trips = initial_profile["labeled_trips"]
+
+        # Run the pipeline again, but don't add any new data
+        etc.runIntakePipeline(self.testUUID)
+
+        # Stats should remain the same
+        updated_profile = edb.get_profile_db().find_one({"user_id": self.testUUID})
+        self.assertIsNotNone(updated_profile, "Profile should still exist.")
+        self.assertEqual(
+            updated_profile["total_trips"], 
+            initial_total_trips,
+            f"Expected total_trips to remain {initial_total_trips}, got {updated_profile['total_trips']}"
+        )
+        self.assertEqual(
+            updated_profile["labeled_trips"], 
+            initial_labeled_trips,
+            f"Expected labeled_trips to remain {initial_labeled_trips}, got {updated_profile['labeled_trips']}"
+        )
+
         test_call_ts = time.time()
         enac.store_server_api_time(self.testUUID, "test_call_ts", test_call_ts, 69420)
         etc.runIntakePipeline(self.testUUID)


shankari:
@TeachMeTW just to clarify, this PR does not incorporate the fix to the two use cases - that was fixed earlier.
This PR adds two new test cases, and passes in the trip_key as an input instead of hardcoding `analysis/confirmed_trip`
- Please split the two unrelated changes into two separate commits
- In the current two commits that you have, the first one adds `testGetAndStoreUserStatsNewData`, and the second one removes it. **Please clean up the commit history to avoid unnecessary churn.**

TeachMeTW:
@shankari I thought incorporating the fix meant fixing the unit test, hence the two new test cases. 

I have split the two unrelated changes into two commits and cleaned up the commit history as requested.

shankari:
This is correct, so I am happy to merge it.
But you also need to test what happens when these entries are not available in the profile.
e.g. between the time that the user registers and the pipeline runs for the first time, if the dashboard tries to access it, what happens?

shankari:
>  I thought incorporating the fix meant fixing the unit test, hence the two new test cases.

Well, the fix is already done, but I guess you can argue that if it is not tested, it is not done 😄

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -75,7 +75,7 @@ def testGetAndStoreUserStats(self):
         self.assertIn("pipeline_range", profile, "User profile should contain 'pipeline_range'.")
         self.assertIn("last_call_ts", profile, "User profile should contain 'last_call_ts'.")
 
-        expected_total_trips = 5
```
shankari, line 78: the commit that added this (https://github.com/e-mission/e-mission-server/pull/1006/commits/8a9939cf7d4274e016668b08b6bf818442d7a598) should have had a short explanation justifying the change.

I _thought_ this was because you are running the pipeline a second time (`etc.runIntakePipeline(self.testUUID)`). But you are not loading any new data in these changes *and* we check the variable before the new pipeline run
https://github.com/e-mission/e-mission-server/pull/1006/files#diff-69c9cbc3241bd93c4c5cae6866fec4189a9aa32c8964e86b2bdc3749bae4d2b2L82

So why does this need to change? Please investigate and provide justification.

shankari (changes requested):

shankari:
Aha! After I pushed c77c5c549d1839dbcb56b0921126e769edfdebe6 related to using the input key, the tests started failing
https://github.com/e-mission/e-mission-server/actions/runs/12638957212/job/35216284230

with this error

```
web-server-1  | ======================================================================
web-server-1  | FAIL: testGetAndStoreUserStats (analysisTests.intakeTests.TestUserStat.TestUserStats)
web-server-1  | Test get_and_store_user_stats for the user to ensure that user statistics
web-server-1  | ----------------------------------------------------------------------
web-server-1  | Traceback (most recent call last):
web-server-1  |   File "/src/e-mission-server/emission/tests/analysisTests/intakeTests/TestUserStat.py", line 81, in testGetAndStoreUserStats
web-server-1  |     self.assertEqual(profile["total_trips"], expected_total_trips,
web-server-1  | AssertionError: 8 != 5 : Expected total_trips to be 5, got 8
web-server-1  | 
web-server-1  | ----------------------------------------------------------------------
```

@TeachMeTW if a test fails, you cannot arbitrarily modify the tests so that it passes. That breaks the entire point of having automated tests, which is to catch regressions.

We need to investigate and indentify why we are getting different results for confirmed and composite trips

@TeachMeTW do you have an ETA for the investigation to be complete??

TeachMeTW:
@shankari I am looking into it at the moment

TeachMeTW:
@shankari That is because we WERE using `confirmed_trip` rather than `composite_trip`
```
        total_trips = ts.find_entries_count(key_list=["analysis/confirmed_trip"])
        labeled_trips = ts.find_entries_count(
            key_list=["analysis/confirmed_trip"],
            extra_query_list=[{'data.user_input': {'$ne': {}}}]
        )
```
During your refactor, you modified the statement to use `composite_trip` whereas the `get_store_user_stats` was hardcoded to `confirmed_trip` hence the mismatch:
```
        try:
            run_intake_pipeline_for_user(uuid, skip_if_no_new_data)
            with ect.Timer() as gsr:
                logging.info("*" * 10 + "UUID %s: storing user stats " % uuid + "*" * 10)
                print(str(arrow.now()) + "*" * 10 + "UUID %s: storing user stats " % uuid + "*" * 10)
                eaurs.get_and_store_user_stats(uuid, "analysis/composite_trip")
            esds.store_pipeline_time(uuid, 'STORE_USER_STATS',
                                time.time(), gsr.elapsed)
```

shankari:
@TeachMeTW the data model is such that we should have a 1:1 mapping between confirmed and composite trips.
We create composite trips by taking confirmed trips and adding on the other components of trips.

> By changing the internal get_and_store_user_stat to analysis/composite_trip rather than analysis/confirmed_trips, it matches with the 8 expected. Hence I believe 8 is the ideal value and not 5.

This is not expected and needs more explanation

TeachMeTW:
> @TeachMeTW **the data model is such that we should have a 1:1 mapping between confirmed and composite trips.** We create composite trips by taking confirmed trips and adding on the other components of trips.
> 
> > By changing the internal get_and_store_user_stat to analysis/composite_trip rather than analysis/confirmed_trips, it matches with the 8 expected. Hence I believe 8 is the ideal value and not 5.
> 
> This is not expected and needs more explanation
> 

I will look into it.

TeachMeTW:
@shankari  So as we
> should have a 1:1 mapping between confirmed and composite trips. We create composite trips by taking confirmed trips 
> and adding on the other components of trips.

Does that mean confirmed trips is the ground truth in terms of length/amount?

TeachMeTW:
I added debugging and logging with my changes and got to this conclusion:
```
INFO:root:User UUID-xyz-abc-bla-bla-bla Summary:
INFO:root:Confirmed Places: 9
INFO:root:Confirmed Trips: 5
INFO:root:Composite Trips: 5
```

TeachMeTW:
# Changes:


```triplikeEntries = list(ts.find_entries([esda.CONFIRMED_TRIP_KEY, esda.CONFIRMED_UNTRACKED_KEY], time_query=time_query))```

to 

 ```triplikeEntries = list(ts.find_entries([esda.CONFIRMED_TRIP_KEY], time_query=time_query))```
 
 Removed esda.CONFIRMED_UNTRACKED_KEY to ensure that only Confirmed Trips are processed for creating Composite Trips.
 
 <br></br>

 ```
 statuscheck["curr_confirmed_trip_count"] = db.count_documents({
    "metadata.key": esda.CONFIRMED_TRIP_KEY,
    "user_id": user_id
})
```

 Introduced a new count to track the number of Confirmed Trips per user.
  <br></br>
  
  ```
  if needs_hack:
    assert statuscheck["curr_composite_trip_count"] == 0, "Current composite trip count is not zero."
    estbt.BuiltinTimeSeries.update(ecwe.Entry(ct))
```
Added an assertion to verify that the current composite trip count is zero before updating, preventing duplication and ensuring alignment.
<br></br>
`[esda.CONFIRMED_TRIP_KEY, esda.CONFIRMED_UNTRACKED_KEY]` to `[esda.CONFIRMED_TRIP_KEY]`
Eliminated the inclusion of esda.CONFIRMED_UNTRACKED_KEY to prevent Composite Trips from being created based on untracked trip data.
<br></br>


## Questions
@shankari I just had a question in regards to the 1:1 mapping? If we are adding to it, how is it keeping a 1:1 ratio? I am not sure if ommiting the untracked_key is the good way to go about it. Let me know your thoughts before I push this commit. Currently the test works with 5 as the expected trip. The debug statement was above.

**Comment by shankari** [comment]
@TeachMeTW first, identify the problem, then think about potential fixes, and **then** make the changes.

Aha! This is the problem. Good job identifying it.

```
triplikeEntries = list(ts.find_entries([esda.CONFIRMED_TRIP_KEY, esda.CONFIRMED_UNTRACKED_KEY], time_query=time_query))
```

One potential fix is indeed to change the `triplikeEntries`. But the app also uses composite trips (which is why we want to keep it warm), so changing this data structure will break untracked time. 

Do you have thoughts for an alternate fix? I can think of at least one other option. Please think through all the options, and list their pros and cons first.

**Please do not make changes without thinking through several pros and cons**

**Comment by TeachMeTW** [comment]
@shankari I will take a look at these options and explore the pros and cons as suggested; will see if I can uncover a pssobile fix

**Comment by TeachMeTW** [comment]
@shankari 

Here are some options, let me know your thoughts. This is based on how I understood the issue at hand:
## Add a metadata tag
- When creating a composite trip, add metadata indicating whether it originated from a confirmed trip or an untracked period.
- When counting or comparing confirmed trips to composite trips, filter out composite trips that originated from untracked data.

### Pros:
- Selective Counting: Allows accurate matching of confirmed trips to their composite counterparts without losing untracked trip processing.
- Flexibility: You can still generate composite trips for untracked data, but they won't interfere with the confirmed trip count.
- Transparency: Admins can later view details about which composites came from untracked time if needed.

### Cons:
- Additional Complexity: Requires adding a metadata field and updating filtering logic.
- Implementation Overhead: Slight changes to data structure and queries are necessary.

## Maintain Separate Counts and Labels
- Continue creating composite trips from both confirmed and untracked data.
- Adjust the reporting interface to display two distinct numbers:
- One for confirmed trips (and their composites).
- Another for untracked time composites.

### Pros:
- Clarity in Data: Admins can see how many composites come from confirmed trips versus untracked periods.
- No Data Filtering Needed: No need to change the creation logic or add metadata; just adjust the reporting.

### Cons:
- It gets muddy by multiple similar metrics, potentially leading to confusion.

## Change Labeling to "Trips and Untracked Time"
- Leave the data structure as is.
- Simply change the labels in the admin reports to reflect that the count includes both trips and untracked time.

### Pros:
- Minimal Changes: Only label/text changes; no changes to logic or structure.
- Quick Fix: Fast to implement

### Cons:
- Misleading Metric: The label might still be confusing or misleading to admins, as it lumps two different kinds of data together.
- Lack of Granularity: Does not resolve the underlying issue of distinguishing between confirmed and untracked data.

**Comment by shankari** [comment]
@TeachMeTW
1. this still fairly generic and what I would expect from ChatGPT if I said "tell me now to solve a problem where I need to filter two types of trips". I would encourage you to carefully **look at the current implementation** instead of saying "add a metadata tag" What would it take to add the tag? How much filtering of the logic is needed? You are writing code within the context of an existing, complex codebase. You want to start with seeing what exists and how your solution can integrate with it, instead of starting de novo.
2. If you think that option 1 is best, but it is too complex, you need to know **how** complex so you can weigh the options.

**Comment by TeachMeTW** [comment]
> @TeachMeTW
> 
>     1. this still fairly generic and what I would expect from ChatGPT if I said "tell me now to solve a problem where I need to filter two types of trips". I would encourage you to carefully **look at the current implementation** instead of saying "add a metadata tag"
>        
>        * What would it take to add the tag?
>        * How much filtering of the logic is needed?
>          You are writing code within the context of an existing, complex codebase. You want to start with seeing what exists and how your solution can integrate with it, instead of starting de novo.
> 
>     2. If you think that option 1 is best, but it is too complex, you need to know **how** complex so you can weigh the options.

@shankari I believe from reading the code, we already have a dict for `composite_trip_data` which as keys like locations, sections, and we already have `metadata` which has an `origin_key` which is from ct[metadata][key] which has stuff like `analysis/confirmed_tracked`, `confirmed_trip`, `confirmed_untracked` and whatnot based on my digging.

I would think just to do add to it by doing `composite_trip_entry['metadata']['origin_type'] = confirmed or untracked or something to differentiate.

Now with that, we can modify the db.count_document() to filter for metadata.origin_type

ie:

metadata.key: esda.COMPOSITE_TRIP_KEY
metadata.origin_type: confirmed


So: 
> What would it take to add the tag?
I believe my iteration above would suffice by adding to the composite trip entry so that line
> How much filtering of the logic is needed?
We could use the count_documents and filter with the new metadata.origin so itd be like:

```
db.count_documents({
       metadata.key: esda.COMPOSITE_TRIP_KEY
       metadata.origin_type: confirmed
})
```

Would you say this explanation and train of thought suffices?

**Comment by TeachMeTW** [comment]
> @TeachMeTW first, identify the problem, then think about potential fixes, and **then** make the changes.
> 
> 
> 
> Aha! This is the problem. Good job identifying it.
> 
> 
> 
> ```
> 
> triplikeEntries = list(ts.find_entries([esda.CONFIRMED_TRIP_KEY, esda.CONFIRMED_UNTRACKED_KEY], time_query=time_query))
> 
> ```
> 
> 
> 
> One potential fix is indeed to change the `triplikeEntries`.
> 
> 
> 
> Do you have thoughts for an alternate fix? I can think of at least one other option. Please think through all the options, and list their pros and cons first.
> 
> 
> 
> **Please do not make changes without thinking through several pros and cons**

@shankari I also wanted to clarify by what you mean by "But the app also uses composite trips (which is why we want to keep it warm), so changing this data structure will break untracked time."

How are we keeping it warm and how will changing the structure break things?

**Comment by TeachMeTW** [comment]
Well thinking about it more -- it begs the question what things are effected with this change. Furthermore, why wasn't this 1:1 issue caught before? Does this mean it was broken already? If so why didn't the tests catch it?

**Comment by TeachMeTW** [comment]
Since I saw that we have origin_keys already, I was thinking modifying filtering to:
```
composite_trip_count = db.count_documents({
    "metadata.key": esda.COMPOSITE_TRIP_KEY,
    "user_id": user_id,
    "metadata.origin_key": esda.CONFIRMED_TRIP_KEY # (or hardcode it for this manner i suppose)
})
```
We could filter by confirmed this way?

But this seems too good to be true. How do you suggest me testing this @shankari -- what depends on the composite objects? I know as you said the phone does and I don't want to affect that and whatnot

**Comment by TeachMeTW** [comment]
We set `composite_trip_entry["metadata"]["origin_key"] = origin_key` in `def create_composite_trip(ts, ct):` but we never use it in the composite_trip creation. Perhaps we could utilize this

**Comment by TeachMeTW** [comment]
@shankari 
In essence, would total trips be the larger of confirmed trips and composites or would total trips just be confirmed trips?

**Comment by JGreenlee** [comment]
It is not perfectly 1:1. All confirmed trips have a corresponding composite trip and all untracked have a corresponding composite trip.
If there are any untracked, there will be more composite trips than there are confirmed trips.
The # of composite trips should equal # of confirmed trips + # of untracked

I can explain in more detail if you want to get on a call later

**Comment by JGreenlee** [comment]
I can also provide context for

> How are we keeping it warm and how will changing the structure break things?

**Comment by shankari** [comment]
@TeachMeTW here's a concrete example working more independently. If you look at this PR, you have asked me 4 times in ~ 12 hours! Neither Jack nor I can respond at that rate and still get our own work done. The goal is for you to slow down, read the code, explore the data and figure it out yourself.

It is more efficient for _you_ to ask us when you have a question, but it is less efficient for _us_.

You have been added to the internal issue where we are discussing scalability changes.
You have access to the full codebase.
You have access to a dataset and have been taught how to load it.
Poke around at all of those and answer your own questions, including by making modifications if necessary.

Learning how to read existing code is a critical skill.

**Comment by TeachMeTW** [comment]
@shankari After clearing my thoughts, rereading and building off from my theories and approaches from yesterday, it seems that Option 1 -- the metadata was ALREADY in the codebase so it was a simple fix once I clarified with Jack with what composite trips are and why and how they are used in conjunction with the app.

My notes:

Untracked: 
- Time when phone is off or not tracking location, gaps in the location spanning x time, any point in time for a user there should be a place, trip, or untracked — so there’s no overlap. So people can see the gaps. At any time there will always be one of the three
- The ground truth is confirmed trips
- We want to query with composite_trips — what gets sent to the phone (has place, trip or untracked) — created from untracked or places — embedded


Keeping composite warm:
- Pipeline runs every hour
- We want composite_trips to be queried so it stays warm

Filter by origin_key

Total trips = confirmed trips

Thus, the solution should just be filtering by origin which I suspected yesterday, and I did so by adding an extra query: `extra_query_list=[{'metadata.origin_key': 'analysis/confirmed_trip'}]`

**Comment by TeachMeTW** [comment]
Now the tests passes and we keep a 1:1 ratio WITHOUT changing the existing structure. Furthermore, I edited the existing commits to ensure no commit muddle/churn

**Comment by shankari** [comment]
> that Option 1 -- the metadata was ALREADY in the codebase so it was a simple fix

Yes. this was my point with 

> I can think of at least one other option

Again, the goal here is not to write a ton of code. The goal is understand the existing system, and to make changes that are consistent with it.

> once I clarified with @JGreenlee with what composite trips are and why and how they are used in conjunction with the app.

I'm glad you were able to get clarity from Jack on this, and I appreciate Jack taking the time to help you, but I don't want you to ask Jack 5 questions in 12 hours either - that won't let **him** do any work.

File: emission/tests/analysisTests/intakeTests/TestUserStat.py [hunk 2]
```diff
@@ -95,9 +95,52 @@ def testGetAndStoreUserStats(self):
...
+        enac.store_server_api_time(self.testUUID, "test_call_ts", test_call_ts, 69420)
+        etc.runIntakePipeline(self.testUUID)
+
+        # Retrieve the profile from the database
+        profile = edb.get_profile_db().find_one({"user_id": self.testUUID})
+
+        # Verify that last_call_ts is updated correctly
+        expected_last_call_ts = test_call_ts
+        actual_last_call_ts = profile.get("last_call_ts")
+
```
shankari, line 109: I don't see where this is using new trip data. You are re-running the pipeline, but have not loaded any new data before doing so.

shankari (changes requested):

emission/tests/analysisTests/intakeTests/TestUserStat.py
TeachMeTW: @shankari I misunderstood and thought that loading from an empty profile/uuid would constitute as loading new trip data. I went ahead and fixed that.

emission/analysis/result/user_stat.py
```diff
@@ -61,9 +61,9 @@ def get_and_store_user_stats(user_id: str, trip_key: str) -> None:
         end_ts_result = ts.get_first_value_for_field(trip_key, "data.end_ts", pymongo.DESCENDING)
         end_ts = None if end_ts_result == -1 else end_ts_result
 
-        total_trips = ts.find_entries_count(key_list=["analysis/confirmed_trip"])
+        total_trips = ts.find_entries_count(key_list=[trip_key], extra_query_list=[{'metadata.origin_key': 'analysis/confirmed_trip'}])
```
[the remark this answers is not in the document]
TeachMeTW: @shankari From what I see, every implementation of `get_and_store_user_stats` currently passes in trip_key as analysis/composite_trip.

Given this, it feels less intuitive to hardcode it as analysis/confirmed_trip within the function. Instead, wouldn't it make more sense to ensure consistency by relying on the trip_key passed as a parameter? This way, we maintain flexibility for future cases where trip_key might need to vary and avoid unexpected behavior.

Alternatively, if hardcoding is essential for the specific use case of composite trips, perhaps we could document this behavior clearly and explain why it is done to avoid confusion.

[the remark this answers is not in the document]
JGreenlee: @TeachMeTW 
You must either pass in both the trip key and the extra query as arguments, or have them both hardcoded.

If the extra query is hardcoded and the trip key is anything other than `analysis/composite_trip`, it will always return 0 entries because nothing other than composite trips have the `origin_key` field.
In that scenario, the `trip_key` argument would be pointless, and not do what it seems to do, which is bound to cause frustration later on.

Yes, it would be nice to be able to swap out the trip key if we decide later that we don't need to keep composite trips warm and we want to switch the query to confirmed trips.
BUT it is better to not have a feature/customization than to pretend we have one and not have it work the way you'd expect, because that's where dangerous mistakes originate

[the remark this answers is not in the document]
TeachMeTW: @JGreenlee What if we `def get_and_store_user_stats(trip_key, extra_query_list=None):`? 

And have logic like:
```
if extra_query_list is None and trip_key == "analysis/composite_trip":
        extra_query_list = [{'metadata.origin_key': 'analysis/confirmed_trip'}]
```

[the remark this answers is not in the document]
JGreenlee: You could do something like that and have `get_and_store_user_stats` adapt to whatever is thrown at it

But here's another reason to define the trip key inside `get_and_store_user_stats` rather than as an argument.
Currently `get_and_store_user_stats` called in a few places. If in a few months we want to switch this to confirmed trips, how many places in the codebase do we want to make changes? Ideally, 1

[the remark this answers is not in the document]
TeachMeTW: Got it thank you for the clarification.

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -57,11 +57,23 @@ def tearDown(self):
...
+        # Load in new data
+        with open("emission/tests/data/real_examples/shankari_2015-aug-27") as fp:
+            additional_entries = json.load(fp, object_hook=esj.wrapped_object_hook)
+
+        # Store in the DB
+        for entry in additional_entries:
+            # Adjust the entry's user_id to match the existing user's UUID if needed
+            entry['user_id'] = self.testUUID  
+            edb.get_timeseries_db().insert_one(entry)
+
```
shankari, line 74: why is this not just `setupRealExampleWithEntries`?
**please look at existing code and use it instead of starting over and re-implementing**
You are working with an existing codebase
You need to learn how to read, modify and apply existing code.

TeachMeTW: @shankari You're absolutely right, and I appreciate the feedback. I knew about `setupRealExampleWithEntries` and `setupRealExample`, but I autopiloted into a naive solution instead of stopping and thinking if this was already in the codebase using it. That was a mistake on my part, and I’ll refactor my code to leverage the existing function and slow down on jumping the gun and trying to implement things on the fly.

shankari: @TeachMeTW note also that, with the review cycle, jumping the gun actually results in _slower_ completion and more work for everybody. If you had spent the additional 10 minutes to look through the existing codebase, I could have merged this during my review.

Since you didn't, I had to put in review comments, you had to make more changes, and I have to now look at it again today when I have time to review. So instead of the code being "done" yesterday, it will be done tonight, with an additional time cost for me.

"Measure twice, cut once" https://en.wiktionary.org/wiki/measure_twice_and_cut_once

shankari:
Closing this since it is subsumed by https://github.com/e-mission/e-mission-server/pull/1023 https://github.com/e-mission/e-mission-server/pull/1024 https://github.com/e-mission/e-mission-server/pull/1025

PR #1009, opened 2025-01-10
Add `composite_trip` Migration Scripts

JGreenlee:
### add _common.py with run_on_all_deployments for migration scripts

`run_on_all_deployments` can be used byother migration scripts, causing them to be run on all production deployments
It does this by scanning the `nrel-openpath-deploy-configs` repo for a list of deployments and modifying the DB_HOST accordingly in between each time the function is called

### add trim_fluff_from_composite_trips.py

"""
Trim unnecessary fields from composite trips in the analysis_timeseries_db.
The shape of the remaining fields is unchanged.
"""

### DIFF
--- a/bin/historical/migrations/_common.py
+++ b/bin/historical/migrations/_common.py
@@ -2,28 +2,29 @@ import os
 import subprocess
 import importlib
 import logging
+import tempfile
+import time
 
 import emission.core.get_database as edb
 
-DB_HOST_TEMPLATE = "mongodb://localhost:27017/openpath_prod_REPLACEME"
+DB_HOST_TEMPLATE = os.environ.get('DB_HOST_TEMPLATE', "mongodb://localhost:27017/openpath_prod_REPLACEME")
 
-proc = subprocess.run(
-      'rm -rf nrel-openpath-deploy-configs && ' +
-      'git clone --no-checkout https://github.com/e-mission/nrel-openpath-deploy-configs.git && ' +
-      'cd nrel-openpath-deploy-configs && ' +
-      'git ls-tree -r --name-only HEAD | grep configs/',
-      shell=True,
-      capture_output=True,
-      text=True)
-filenames = proc.stdout.replace("configs/", "").split("\n")
+if 'PROD_LIST' in os.environ:
+    PROD_LIST=os.environ['PROD_LIST'].split(",")
+else:
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        print(f"created {tmpdirname=} to find list of configs")
+        os.chdir(tmpdirname)
+        proc = subprocess.run(
+              f"git clone https://github.com/e-mission/nrel-openpath-deploy-configs.git", shell=True)
+        filenames = os.listdir(f"nrel-openpath-deploy-configs/configs/")
 
-PROD_LIST = [
-    fname.split(".")[0]
-    for fname in filenames
-      if fname and 'dev-' not in fname and 'stage-' not in fname
-]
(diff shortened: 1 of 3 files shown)


bin/historical/migrations/_common.py
```diff
@@ -0,0 +1,41 @@
+import os
+import subprocess
+import importlib
+import logging
+
+import emission.core.get_database as edb
+
+DB_HOST_TEMPLATE = "mongodb://localhost:27017/openpath_prod_REPLACEME"
```
shankari, line 8:
```suggestion
DB_HOST_TEMPLATE = os.environ["DB_HOST_TEMPLATE"]
```

The template for DB_HOST on DocumentDB is very different. It has a different host (not `localhost`), and includes user+password

bin/historical/migrations/_common.py
```diff
@@ -0,0 +1,41 @@
...
+proc = subprocess.run(
+      'rm -rf nrel-openpath-deploy-configs && ' +
+      'git clone --no-checkout https://github.com/e-mission/nrel-openpath-deploy-configs.git && ' +
+      'cd nrel-openpath-deploy-configs && ' +
+      'git ls-tree -r --name-only HEAD | grep configs/',
+      shell=True,
+      capture_output=True,
+      text=True)
+filenames = proc.stdout.replace("configs/", "").split("\n")
+
```
shankari, line 19: I assume this works, and I don't want the perfect to be the enemy of the good, but this seems overly convoluted. Why isn't this just

```suggestion
      'rm -rf nrel-openpath-deploy-configs && ' +
      'git clone https://github.com/e-mission/nrel-openpath-deploy-configs.git)'
filenames = os.listdir('nrel-openpath-deploy-configs')
```

It seems much cleaner than capturing stdout and mucking around with it.

shankari:
```suggestion
  with tempfile.TemporaryDirectory() as tmpdirname:
      print('created temporary directory', tmpdirname)
      proc = subprocess.run('git clone --no-checkout https://github.com/e-mission/nrel-openpath-deploy-configs.git ' + tmpdirname)
filenames = os.listdir(tmpdirnme+"/configs")
```

shankari (changes requested):
Given that this was ready yesterday afternoon, and it is pressing, I am just going to make the changes now so we can run the script overnight

bin/historical/migrations/_common.py
JGreenlee: I did this to avoid downloading the actual files since all we needed were the filenames. May have been overengineered

bin/historical/migrations/_common.py
shankari: Fixed in https://github.com/e-mission/e-mission-server/pull/1009/commits/947492d1b7ec509e7afc0f7b6af0a9eca52ac93a

bin/historical/migrations/_common.py
shankari: I thought it may be something like that (no-checkout) but yes, I thought it was overly complex, and having the full configs might be helpful for some functions (e.g. maybe the function will be sensitive to the variation in the configs).

shankari: Also fixed in https://github.com/e-mission/e-mission-server/pull/1009/commits/947492d1b7ec509e7afc0f7b6af0a9eca52ac93a

shankari:
I stayed up until 2am-ish last night so I will merge and run this tomorrow, when I will be more careful. 
I don't want to run scripts against production when I am burned out

PR #1010, opened 2025-01-22
✂️ Add script to purge inactive users from decommissioned deployments

JGreenlee:
The following conditions must be met before this script should be run on a program:

- The program’s collection period (per MOU) has passed, and we have confirmed with program admins that it is ok to shut down
- Data has already been archived in TSDC

I tested this on my local dump of `nrel-commute`:

```
(emission) jgreenle@jgreenle-34794s e-mission-server % PROD_LIST=nrel-commute DB_HOST=mongodb://localhost:27017/openpath_prod_nrel_commute PYTHONPATH=. ./e-mission-py.bash bin/historical/migrations/purge_inactive_users.py
Retrieved config: {'DB_HOST': 'mongodb://localhost:27017/openpath_prod_nrel_commute', 'DB_RESULT_LIMIT': None}
Connecting to database URL mongodb://localhost:27017/openpath_prod_nrel_commute
PROD_LIST: ['nrel-commute']
About to run purge_inactive_users on 1 deployments. Proceed? [y/n]
y
Running purge_inactive_users for nrel-commute on DB nrel_commute
Retrieved config: {'DB_HOST': 'mongodb://localhost:27017/openpath_prod_nrel_commute', 'DB_RESULT_LIMIT': None}
Connecting to database URL mongodb://localhost:27017/openpath_prod_nrel_commute
Total users: 51
Finding inactive users...
Checking activity for user (uuid)
for user <uuid>, last call was 1734401813.8308148
(...continued)
Of 51 users, found 46 inactive users:
(list of UUIDs)
Purging inactive users...
Purging user (uuid)
(...continued)
```

I also ran it a second time to ensure nothing bad happened:
```
Of 5 users, found 0 inactive users:
[]
Purging inactive users...
```

I inspected the DB and confirmed that 5 users remain.

Then, I referenced the UUIDs table on the admin dashboard and verified that those 5 users have a recent last_call_ts, while the other 46 users do not.

### DIFF
--- a/bin/historical/migrations/purge_inactive_users.py
+++ b/bin/historical/migrations/purge_inactive_users.py
@@ -1,0 +1,64 @@
+import arrow
+import pymongo
+import emission.core.get_database as edb
+import emission.storage.timeseries.abstract_timeseries as esta
+import bin.debug.common as common
+from _common import run_on_all_deployments
+
+SECONDS_90_DAYS = 60 * 60 * 24 * 90
+
+NOW_SECONDS = arrow.now().timestamp()
+
+def find_inactive_uuids(uuids_entries):
+    '''
+    Users are inactive if:
+     - no API calls in the last 90 days
+      AND
+     - no locations in the last 90 days
+    '''
+    inactive_uuids = []
+    for u in uuids_entries:
+        print(f'Checking activity for user {u["uuid"]}')
+        ts = esta.TimeSeries.get_time_series(u['uuid'])
+
+        last_call_ts = ts.get_first_value_for_field(
+            key='stats/server_api_time',
+            field='data.ts',
+            sort_order=pymongo.DESCENDING
+        )
+        print(f'for user {u["uuid"]}, last call was {last_call_ts}')
+        if last_call_ts > NOW_SECONDS - SECONDS_90_DAYS:
+            continue
+
+        last_loc_ts = ts.get_first_value_for_field(
+            key='background/location',
+            field='data.ts',
+            sort_order=pymongo.DESCENDING
+        )
+        print(f'for user {u["uuid"]}, last location was {last_loc_ts}')
+        if last_loc_ts > NOW_SECONDS - SECONDS_90_DAYS:
+            continue
+
+        print(f'User {u["uuid"]} is inactive')
+        inactive_uuids.append(u['uuid'])
+
+    return inactive_uuids
+
+
+def purge_inactive_users():
+    total_users = edb.get_uuid_db().count_documents({})
+    print(f'Total users: {total_users}')
+    uuids_entries = edb.get_uuid_db().find()
+    print('Finding inactive users...')
+    inactive_uuids = find_inactive_uuids(uuids_entries)
+    print(f'Of {total_users} users, found {len(inactive_uuids)} inactive users:')
+    print(inactive_uuids)
+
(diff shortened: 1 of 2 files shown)


bin/historical/migrations/purge_inactive_users.py
```diff
@@ -0,0 +1,64 @@
...
+
+NOW_SECONDS = arrow.now().timestamp()
+
+def find_inactive_uuids(uuids_entries):
+    '''
+    Users are inactive if:
+     - no API calls in the last 90 days
+      AND
+     - no locations in the last 90 days
+    '''
```
shankari, line 18: @JGreenlee is this consistent with the definition of "active user" in the admin dashboard and the public dashboard? I think we should ensure that our definition of "active user" is consistent throughout the platform.

bin/historical/migrations/purge_inactive_users.py
```diff
@@ -0,0 +1,64 @@
...
+    print(inactive_uuids)
+
+    print("Purging inactive users...")
+    for u in inactive_uuids:
+        print(f'Purging user {u}')
+        common.purge_entries_for_user(u, True)
+
+
+if __name__ == '__main__':
+    run_on_all_deployments(purge_inactive_users)
```
shankari, line 64: This is useful, but I think that a first step is just a script to identify the number of inactive users in each program so that we can also identify inactive programs.

Such inactive programs are:
- candidates for experimenting with the new architecture, AND
- proactively reaching out to ask if we can shut down, even if the MOU is still active

Since the new proposed script would also use `find_inactive_uuids`, you can have it be a variant of this (controlled via a command line argument that you get using argparse) or a separate script that imports `find_inactive_uuids`

shankari (changes requested):

bin/historical/migrations/purge_inactive_users.py
catarial: In the public dashboard, active users are the number of users that were active on a **specific day**

https://github.com/e-mission/em-public-dashboard/blob/00b70c96495f2bcc07837c4d89ff073bc30fdcc1/viz_scripts/generic_timeseries.ipynb#L186

catarial: In the admin dashboard, active users are users that have had activity in the last day

https://github.com/e-mission/op-admin-dashboard/blob/33d41d16ce9cd175922aa5cc6cf4264aee247911/pages/home.py#L237

bin/historical/migrations/purge_inactive_users.py
catarial: Would this essentially be the same script but without the part where it actually removes the users?

shankari: Sure.

I would make the default be to print, and actually remove the users only when the appropriate command line arg is passed in.

bin/historical/migrations/purge_inactive_users.py
catarial: the admin dashboard also only uses the timestamp of API calls and no location timestamps

shankari: I think we should use both API calls and locations. We can change this for now, but then encode this into https://github.com/JGreenlee/e-mission-common and change the others to use it.

PR #1011, opened 2025-01-22
⏪ Go back to generating trip statistics from the confirmed trips

shankari:
We deployed the change to switch to composite trips around Jan 5th, at which time I/O usage tripled. Since it looks like DocumentDB does not use the index, and instead pulls all data into memory to filter there, the switch to composite trips would likely result in more data being read, which may have resulted in greater I/O.

The change was originally to keep the cache warm, but per testing, it looks like it doesn't really help with performance. And it increases the I/O rate, which also affects the overall cost.

So we revert the change, and rely on trimming down object sizes to improve performance instead.

We were always reading composite trips to determine the pipeline range added in https://github.com/e-mission/e-mission-server/commit/0403ae0ecc8985911451f16fac085fcf0d6328de

We then read the number of confirmed trips, (not composite) as part of https://github.com/e-mission/e-mission-server/commit/39a33586e727d6e14ed7d0773723ad4e0d2cbaa8

The change to run the user stats every time instead of only when there was new data was in https://github.com/e-mission/e-mission-server/commit/9d1f414b067cbb8411214fed88d45aaa8c6d5eae

Testing done:
- One liner, no testing

### DIFF
--- a/emission/tests/analysisTests/intakeTests/TestUserStat.py
+++ b/emission/tests/analysisTests/intakeTests/TestUserStat.py
@@ -43,8 +43,7 @@ class TestUserStats(unittest.TestCase):
             # Initialize the profile if it does not exist
             edb.get_profile_db().insert_one({"user_id": self.testUUID})
 
-        #etc.runIntakePipeline(self.testUUID)
-        epi.run_intake_pipeline_for_user(self.testUUID, skip_if_no_new_data = False)
+        etc.runIntakePipeline(self.testUUID)
         logging.debug("UUID = %s" % (self.testUUID))
 
     def tearDown(self):
--- a/emission/tests/analysisTests/intakeTests/TestUserStat.py
+++ b/emission/tests/analysisTests/intakeTests/TestUserStat.py
@@ -44,7 +44,7 @@ class TestUserStats(unittest.TestCase):
             edb.get_profile_db().insert_one({"user_id": self.testUUID})
 
         #etc.runIntakePipeline(self.testUUID)
-        etc.runIntakePipeline(self.testUUID)
+        epi.run_intake_pipeline_for_user(self.testUUID, skip_if_no_new_data = False)
         logging.debug("UUID = %s" % (self.testUUID))
 
     def tearDown(self):
--- a/emission/analysis/result/user_stat.py
+++ b/emission/analysis/result/user_stat.py
@@ -41,9 +41,10 @@ def update_user_profile(user_id: str, data: Dict[str, Any]) -> None:
     logging.debug(f"New profile: {user.getProfile()}")
 
 
-def get_and_store_user_stats(user_id: str, trip_key: str) -> None:
+def get_and_store_pipeline_dependent_user_stats(user_id: str, trip_key: str) -> None:
     """
-    Aggregates and stores user statistics into the user profile.
+    Aggregates and stores pipeline dependent into the user profile.
+    These are statistics based on analysed data such as trips or labels.
 
     :param user_id: The UUID of the user.
     :type user_id: str
@@ -52,7 +53,7 @@ def get_and_store_user_stats(user_id: str, trip_key: str) -> None:
     :return: None
     """
     try:
-        logging.info(f"Starting get_and_store_user_stats for user_id: {user_id}, trip_key: {trip_key}")
+        logging.info(f"Starting get_and_store_pipeline_dependent_user_stats for user_id: {user_id}, trip_key: {trip_key}")
 
         ts = esta.TimeSeries.get_time_series(user_id)
         start_ts_result = ts.get_first_value_for_field(trip_key, "data.start_ts", pymongo.ASCENDING)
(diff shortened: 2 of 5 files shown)


shankari:
@JGreenlee @TeachMeTW I also spent some time poking around at when we originally switched from confirmed to composite trips, and it looks like we always used composite trips for determining the pipeline range https://github.com/e-mission/e-mission-server/commit/0403ae0ecc8985911451f16fac085fcf0d6328de

The difference at the time was that:
- we used to only update the pipeline range when the pipeline completed AND
- we used to only return the first and last composite trip

When we started computing user stats, we used composite trips (https://github.com/e-mission/e-mission-server/commit/0403ae0ecc8985911451f16fac085fcf0d6328de) to determine the pipeline range. This ensures that the stored pipeline range is aligned with the UI. If we change this to confirmed trips, we run the risk that an error in generating composite trips will result in no trips being visible to users, since we only query for draft trips outside the pipeline range.

However, at the time, we used to only update the pipeline range when the pipeline ran, which was only when new data came in. We started running it every time when we added `last_call_ts`, because the last call could be updated independently of the pipeline.

However, queries against the timeseries, which consists of many small objects, are fast so the real fix is to split into two updates - one for the pipeline range and the number of trips, and one for the last_call_ts. I am 99% sure I have highlighted this as an optimization earlier, but I can't find it now.

The `pipeline_dependent_stats` only need to be updated when the pipeline is run with new data, so can be more heavyweight, the `pipeline_independent_stats` need to be lightweight but can be run more frequently.

Given that issues with I/O, and the back and forth that is typically required for changes from @TeachMeTW I am going to go ahead and make the change myself.

shankari:
@JGreenlee for visibility and potential review. I will merge and deploy tomorrow!

shankari:
Testing done:
- Launched a server with these changes
- Connected an emulator to it
- Faked a trip using the location simulation

```
START 2025-01-23 00:30:20.575034 POST /usercache/put
START 2025-01-23 00:30:20.597320 POST /usercache/get
END 2025-01-23 00:30:20.600963 POST /usercache/get 414ee37a-732e-4751-a6f6-0d83b41b27f0 0.002919912338256836
END 2025-01-23 00:30:21.948563 POST /usercache/put 414ee37a-732e-4751-a6f6-0d83b41b27f0 1.373345136642456
```

There was only one profile

```
>>> edb.get_profile_db().count_documents({})
1
```

Before running the pipeline, it did not have any of these user stats

```
>>> edb.get_profile_db().find_one()
{'_id': ObjectId('6791fc13b4b3fa6db902c42b'), 'user_id': UUID('414ee37a-732e-4751-a6f6-0d83b41b27f0'), 'mode': {}, 'mpg_array': [32.044384997811726], 'purpose': {}, 'source': 'Shankari', 'update_ts': datetime.datetime(2025, 1, 23, 0, 21, 39, 712000), 'client_app_version': '1.9.6', 'client_os_version': '18.1', 'curr_platform': 'ios', 'manufacturer': 'Apple', 'phone_lang': 'en'}
```

- and after the pipeline ran, it did

```
>>> edb.get_profile_db().find_one()
{'_id': ObjectId('6791fc13b4b3fa6db902c42b'), 'user_id': UUID('414ee37a-732e-4751-a6f6-0d83b41b27f0'), 'mode': {}, 'mpg_array': [32.044384997811726], 'purpose': {}, 'source': 'Shankari', 'update_ts': datetime.datetime(2025, 1, 23, 0, 21, 39, 712000), 'client_app_version': '1.9.6', 'client_os_version': '18.1', 'curr_platform': 'ios', 'manufacturer': 'Apple', 'phone_lang': 'en', 'labeled_trips': 0, 'pipeline_range': {'start_ts': 1737620409.0311022, 'end_ts': 1737621020.497178}, 'total_trips': 1, 'last_call_ts': 1737621021.948551}
```

emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -44,7 +44,7 @@ def setUp(self):
             edb.get_profile_db().insert_one({"user_id": self.testUUID})
 
         #etc.runIntakePipeline(self.testUUID)
-        etc.runIntakePipeline(self.testUUID)
+        epi.run_intake_pipeline_for_user(self.testUUID, skip_if_no_new_data = False)
```
JGreenlee, line 47: This test is failing because `epi.run_intake_pipeline_for_user` never calls `eaurs.get_and_store_pipeline_independent_user_stats`, and these tests 

I think with `etc.runIntakePipeline`, everything will run including the pipeline and pipeline-independent stats

```suggestion
        etc.runIntakePipeline(self.testUUID)
```

shankari: Yup, just made that change and pushed it. I spent some time trying to figure out why we had commented out `etc.runIntakePipeline` in favor of `epi.run_intake_pipeline_for_user` in the first place, but it was part of the initial commit

shankari:
Also fixed a few tests:
- adjusted `etc.runIntakePipeline` to call both the pipeline dependent and pipeline independent user stats
- switched the pipeline call in the setup from `epi.run_intake_pipeline_for_user` to `etc.runIntakePipeline`. The previous call had `etc.runIntakePipeline` commented out, but it was part of the original commit, so I am not sure why it was commented out in the first place. But we need to invoke the pipeline independent stats, which are called outside of `epi.run_intake_pipeline_for_user` since they need to be generated even if the pipeline was skipped, so I think this is a meaningful change for this test.

@TeachMeTW do you remember why `etc.runIntakePipeline` was commented out?

shankari:
@JGreenlee will wait for ~ 30 mins in case you want to re-review and then merge

JGreenlee (approved):
LGTM

PR #1012, opened 2025-01-24
add option to exclude users from pipeline reset

JGreenlee:


shankari:
@JGreenlee did you try testing this? I think this was copied over from the ad-hoc instance, but we should ensure that it is not bitrotted.

PR #1014, opened 2025-01-30
Trip Segmentation Optimization

JGreenlee:


shankari (approved):
This change seems fine; it is a pretty straightforward change in which we pre-load a bunch of values and then re-use them instead of loading them lazily.

I am fine with merging this so we can see the impact on the DB queries.

As part of cleanup, I would like to see:
- [ ] what was justification for this change? In particular, what were the timing results from our sample programs (stage, ccebike, smart commute, stm-community) that led you to focus on these areas?
- [ ] I see that we are querying for loc_df in `trip_segmentation` and passing it in the time/distance filters. And then we are querying for transition and motion activity inside `segment_into_trips`. Why? It seems like we can just read all input data from that time range and pass it into `segment_into_trips` at the same time, in three different dataframes. This will not have an impact on performance, but it is cleaner and easier to understand.

shankari:
I checked staging, and this seems to be working correctly. The logs show a successful run

```
2025-02-06 03:09:22,152:INFO:140272741308224:For stage PipelineStages.TRIP_SEGMENTATION, start_ts is None
2025-02-06 03:09:23,516:INFO:140272741308224:For stage PipelineStages.TRIP_SEGMENTATION, last_ts_processed = 2025-02-06T02:16:41.052000

2025-02-06 05:09:04,586:INFO:140433989416768:For stage PipelineStages.TRIP_SEGMENTATION, start_ts = 2025-02-06T02:16:41.052000
2025-02-06 05:09:10,450:INFO:140433989416768:For stage PipelineStages.TRIP_SEGMENTATION, last_ts_processed = 2025-02-06T04:59:49.876000
```

And I see two composite trips retrieved by the UI

```
10623,1738860615.793,2025-02-06T08:50:15.793000-08:00,"js : getRawEntries, args: [[""analysis/composite_trip""],1738810605.3665705,1738817817.996,""data.end_ts""];
      
got 2 entries"
10624,1738860615.798,2025-02-06T08:50:15.798000-08:00,"js : Timeline: readCompositePromise resolved with 2 trips; 
      readUnprocessedPromise resolved with 0 trips"

```

I can't see them in the UI because of https://github.com/e-mission/e-mission-docs/issues/1108

@JGreenlee can you verify that you saw the trips from yesterday (Wednesday) in the UI?

shankari:
@TeachMeTW as an aside, while evaluating this fix, I notice that we store the `has_trip_ended` stat for every point considered. Given that, we may want to add them up in addition to looking at the individual checks

JGreenlee:
> @JGreenlee can you verify that you saw the trips from yesterday (Wednesday) in the UI?

Yes, it is there and processed on all the phones I brought

TeachMeTW:
> @TeachMeTW as an aside, while evaluating this fix, I notice that we store the `has_trip_ended` stat for every point considered. Given that, we may want to add them up in addition to looking at the individual checks

@shankari to clarify, summarize how many times has_trip_ended was True?

shankari:
Checked nrel-commute as well, no errors except for habitica, and "stuck" pipelines - e.g.

```
2025-02-08 21:30:56,624:ERROR:139951841285952:Found error curr_state.curr_run_ts = 1731022050.2421896 while processing pipeline for user [redacted], skipping
```

Moving this to production as well

PR #1015, opened 2025-02-02
Refactor User Stats Tests to Pipeline-Dependent and Independent Updates

TeachMeTW:
# Description
Updates our tests for user statistics to reflect the recent refactor that separates pipeline-dependent statistics (such as trip counts and pipeline range) from pipeline-independent statistics (such as last_call_ts). Simplified the comments for brevity and to match the new changes.

# Changes

## Pipeline-Dependent Stats:
The test now verifies that after running the intake pipeline via etc.runIntakePipeline, the user profile is updated with:
- total_trips
- labeled_trips
- pipeline_range (with both start_ts and end_ts)

## Pipeline-Independent Stats:
- The test verifies that when a server API call is stored, the last_call_ts field is updated correctly in the user profile.

## Testing
python -m unittest emission/tests/analysisTests/intakeTests/TestUserStat.py

### DIFF
--- a/emission/tests/analysisTests/intakeTests/TestUserStat.py
+++ b/emission/tests/analysisTests/intakeTests/TestUserStat.py
@@ -16,95 +16,100 @@ import emission.storage.json_wrappers as esj
 
 # Our imports
 import emission.core.get_database as edb
-import emission.storage.timeseries.timequery as estt
 import emission.storage.timeseries.abstract_timeseries as esta
-import emission.storage.decorations.analysis_timeseries_queries as esda
 import emission.core.wrapper.user as ecwu
-import emission.net.api.stats as enac
-import emission.pipeline.intake_stage as epi
+import emission.analysis.result.user_stat as user_stats
 
 # Test imports
 import emission.tests.common as etc
-
+import emission.net.api.stats as enac
 
 class TestUserStats(unittest.TestCase):
     def setUp(self):
         """
-        Set up the test environment by loading real example data for both Android and  users.
+        Set up the test environment by loading real example data.
         """
-        # Set up the real example data with entries
         self.testUUID = uuid.uuid4()
+        # Load example entries (this file must exist and contain valid test data)
         with open("emission/tests/data/real_examples/shankari_2015-aug-21") as fp:
-            self.entries = json.load(fp, object_hook = esj.wrapped_object_hook)
-        # Retrieve the user profile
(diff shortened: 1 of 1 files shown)


emission/tests/analysisTests/intakeTests/TestUserStat.py
```diff
@@ -16,95 +16,100 @@
...
+            self.entries = json.load(fp, object_hook=esj.wrapped_object_hook)
         etc.setupRealExampleWithEntries(self)
+        # Ensure that a profile exists for this user.
         profile = edb.get_profile_db().find_one({"user_id": self.testUUID})
         if profile is None:
-            # Initialize the profile if it does not exist
             edb.get_profile_db().insert_one({"user_id": self.testUUID})
-
+        
+        # Run the intake pipeline via the function from 'etc'
```
JGreenlee, line 42: I think this is a good example of a pointless comment. Supposing a Python developer who knows nothing about e-mission, what added value does this comment provide?
The function is already named "runIntakePipeline". The module is already called "etc". You are just describing what the code already says at face value

JGreenlee (approved):
Approved, given that it passes and the stats are now described accurately

Non-critical feedback below for learning. I know it is a hard habit to break, but there are still a lot of unnecessary inline comments.

There are some good reasons for inline comments:
- documenting decisions and trade-offs; why one approach was taken over another, especially if the other approach may be relevant later/elsewhere in the project
- flagging future changes or improvements to be made (TODO)
- offering a deeper explanation to what code does **iff it is not already obvious at a surface level**

However, comments that restate the code in words (e.g. `i += 1 # increment i by 1`) add noise without value

Remember, the ideal code needs little to no comments; it is already readable due to splitting code into understandable chunks and choosing meaningful variable and function names

emission/tests/analysisTests/intakeTests/TestUserStat.py
TeachMeTW: Got it, I had this mindset where open source meant it was better to leave laymans comments.

JGreenlee: You can assume that your audience knows the programming language, but not project-specific details

shankari:
I believe that, since we have implemented https://github.com/e-mission/e-mission-server/pull/1023, these tests are no longer valid. I believe I already updated the tests as part of that PR, but if we want to add additional tests, we should create a new PR.

PR #1019, opened 2025-02-07
Inactive users script

catarial:
Extension of https://github.com/e-mission/e-mission-server/pull/1010

A command line python script that identifies and optionally removes inactive users.

The amount of time that defines an active user can be specified with the `-t` flag.

The threshold time is compared to the timestamps of API calls and locations in the database.

I decided to pull out the part where it get's run on all deployments in favor of more configurability.

### DIFF
--- a/bin/historical/migrations/inactive.py
+++ b/bin/historical/migrations/inactive.py
@@ -1,0 +1,69 @@
+import arrow
+import pymongo
+import emission.core.get_database as edb
+import emission.storage.timeseries.abstract_timeseries as esta
+import bin.debug.common as common
+
+NOW_SECONDS = arrow.now().timestamp()
+
+def find_inactive_uuids(uuids_entries, threshold):
+    inactive_uuids = []
+    for u in uuids_entries:
+        print(f'Checking activity for user {u["uuid"]}')
+        ts = esta.TimeSeries.get_time_series(u['uuid'])
+
+        last_call_ts = ts.get_first_value_for_field(
+            key='stats/server_api_time',
+            field='data.ts',
+            sort_order=pymongo.DESCENDING
+        )
+        print(f'for user {u["uuid"]}, last call was {last_call_ts}')
+        if last_call_ts > NOW_SECONDS - threshold:
+            continue
+
+        last_loc_ts = ts.get_first_value_for_field(
+            key='background/location',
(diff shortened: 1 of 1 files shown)


shankari:
> I decided to pull out the part where it get's run on all deployments in favor of more configurability.

This needs to be restored. I am not going to sit and edit the `DB_HOST` variable 40 times and manually run the script 40 times.
I don't see why it is an either-or for configurability versus running on all deployments.

shankari:
Also, how did you test this? Can you please update "Testing done"?
Please also put this into the correct column in the project (PR for review by Shankari) once you are done

catarial:
> > I decided to pull out the part where it get's run on all deployments in favor of more configurability.
> 
> This needs to be restored. I am not going to sit and edit the `DB_HOST` variable 40 times and manually run the script 40 times. I don't see why it is an either-or for configurability versus running on all deployments.

I misunderstood the _common.py file and I didn't realize you could pass in the arguments of the function to execute with the `*` operator. I changed it back now

catarial:
Testing was done with the private database dump "openpath-stage-snapshot-investigate-demographics".

Since this dump is over a year old, running the script with a threshold of one day marks all the users as inactive

```
./e-mission-py.bash bin/historical/migrations/inactive.py -t 1                                 
Retrieved config: {'DB_HOST': None, 'DB_RESULT_LIMIT': None}
PROD_LIST: ['stage']
About to run start_inactive(86400, False) on 1 deployments. Proceed? [y/n]
...
Of 76 users, found 76 inactive users:
```

Setting the threshold to 500 days gives us some active users

```
./e-mission-py.bash bin/historical/migrations/inactive.py -t 500
...
Of 76 users, found 54 inactive users:
```

Let's remove these users

```
./e-mission-py.bash bin/historical/migrations/inactive.py -t 500 -p
...
About to remove 54 users. Proceed? [y/n]
y
```

Now there are no more inactive users with a threshold of 500 days

```
./e-mission-py.bash bin/historical/migrations/inactive.py -t 500
...
Of 22 users, found 0 inactive users:
```

shankari (approved):

shankari:
@catarial I am merging this right now, but could you submit a separate PR that generates a csv of the results for future analysis?

PR #1021, opened 2025-02-10
🚑️ Do not iterate over map keys while modifying them

shankari:
🐛 Do not iterate over map keys while modifying them

In https://github.com/e-mission/e-mission-server/pull/1008 (and in particular, https://github.com/e-mission/e-mission-server/commit/1ec706e7e0d51cee3dd5942f87b45f3a9b9dd86d), we munged incoming keys with dots so that we could store them properly in mongodb.

We did this by iterating over all the keys, "popping" the one with dots and
storing the value into the munged key, consistent with https://stackoverflow.com/a/11277439

However, this lead to an issue where apparently the key order is also changed,
so that modified keys are iterated over twice, and some keys are not even
checked. And since they are not checked, they are not munged.

This refactors the code to two stages:
- find the `keys_to_munge` (aka `ktm`) by iterating over the map
- iterate over the `keys_to_munge` list and munge the keys

Testing done:

- Before this, the tests in e94140ea55a78c201d3f4c10d4833841769d4cac were failing

```
Ran 8 tests in 28.153s

FAILED (failures=2, errors=1)
```

- After this, the tests in e94141ea55a78c201d3f4c10d4833841769d4cac were passing

```
Ran 8 tests in 28.969s

OK
```

### DIFF
--- a/emission/net/api/usercache.py
+++ b/emission/net/api/usercache.py
@@ -30,17 +30,26 @@ def sync_server_to_phone(uuid):
     return retrievedData
 
 def _remove_dots(entry_doc):
+    keys_to_munge = []
     for key in entry_doc:
         # print(f"Checking {key=}")
         if isinstance(entry_doc[key], dict):
             # print(f"Found dict for {key=}, recursing")
             _remove_dots(entry_doc[key])
         if '.' in key:
-            munged_key = key.replace(".", "_")
-            logging.info(f"Found {key=} with dot, munged to {munged_key=}")
-            # Get and delete in one swoop
-            # https://stackoverflow.com/a/11277439
-            entry_doc[munged_key] = entry_doc.pop(key, None)
+            logging.info(f"Found {key=} with dot, adding to {keys_to_munge=}")
+            keys_to_munge.append(key)
+
+    logging.info(f"Before modifying, {keys_to_munge=}")
+
+    for ktm in keys_to_munge:
+        munged_key = ktm.replace(".", "_")
+        # Get and delete in one swoop
+        # https://stackoverflow.com/a/11277439
+        logging.info(f"Replacing original dotted key {ktm} with {munged_key=}")
+        entry_doc[munged_key] = entry_doc.pop(ktm, None)
+
+    logging.info(f"(After modifying, {entry_doc.keys()=}")
 
 def sync_phone_to_server(uuid, data_from_phone):
     """
--- a/emission/tests/netTests/TestBuiltinUserCacheHandlerInput.py
+++ b/emission/tests/netTests/TestBuiltinUserCacheHandlerInput.py
@@ -345,6 +345,122 @@ class TestBuiltinUserCacheHandlerInput(unittest.TestCase):
         self.assertEqual(len(self.uc1.getMessage()), 0)
         self.assertEqual(len(list(self.ts1.find_entries())), 33)
 
+    def testRemoteDotsIterateWhileModifyingSingleKey(self):
+        test_with_single_key = {'ts': 1734661726.167,
+            'client_app_version': '1.9.4',
+            'name': 'open_notification',
+            'client_os_version': '18.1.1',
+            'reading': {'message': 'Please label your recent trips',
+            'title': 'Trip labels requested',
+            'additionalData': {
+                'coldstart': True, 'title': 'Trip labels requested',
+                'message': 'Please label your recent trips',
+                'foreground': False,
+                'google.c.fid': 'TEST_FID',
+                'gcm_message_id': 'TEST_MESSAGE',
+                'google_c_sender_id': 'TEST_SENDER',
+                'google_c_a_e': '1'}}}
+        self.assertEqual(len(test_with_single_key["reading"]["additionalData"]), 8)
+        self.assertIn("google.c.fid",
+            test_with_single_key["reading"]["additionalData"])
+        mauc._remove_dots(test_with_single_key)
+        self.assertEqual(len(test_with_single_key["reading"]["additionalData"]), 8)
+        self.assertIn("google_c_fid",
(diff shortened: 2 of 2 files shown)


shankari:
@JGreenlee @Abby-Wheelis in case you would like to review this. I will merge and get this out to staging today, open access tomorrow morning (so we can verify the fix), and to the rest of production later this week.

shankari:
We can probably remove `testRemoteDotsIterateWhileModifyingMultiKey2` after it moves to production; it seems identical to `testRemoteDotsIterateWhileModifyingMultiKey`. I just wanted to be paranoid and copy all the examples that we found during investigation

shankari:
wrt

> However, this lead to an issue where apparently the key order is also changed, so that modified keys are iterated over twice, and some keys are not even checked. 

Here are the outputs that were generated during testing after adding additional logs

I just tried to run this manually, and it seems like it works properly. After enabling all the logs and converting them to `print` statements, I see

```
>>> from uuid import UUID
>>> import emission.net.api.usercache as enau
>>> temp_json = {'data': {'ts': 1734661726.167, 'client_app_version': '1.9.4', 'name': 'open_notification', 'client_os_version': '18.1.1', 'reading': {'message': 'Please label your recent trips', 'title': 'Trip labels requested', 'additionalData': {'coldstart': True, 'title': 'Trip labels requested', 'message': 'Please label your recent trips', 'foreground': False, 'gcm.message_id': 'REDACTED', 'google.c.sender.id': 'REDACTED', 'google_c.a.e': '1', 'google.c.fid': 'REDACTED'}}}, 'metadata': {'time_zone': 'America/Denver', 'plugin': 'none', 'write_ts': 1734661726.174125, 'platform': 'ios', 'read_ts': 0, 'key': 'stats/client_time', 'type': 'message'}, 'user_id': UUID(REDACTED)}
>>> enau._remove_dots(temp_json)
Checking key='data', check for '.' is False
Found dict for key='data', recursing
Checking key='ts', check for '.' is False
Checking key='client_app_version', check for '.' is False
Checking key='name', check for '.' is False
Checking key='client_os_version', check for '.' is False
Checking key='reading', check for '.' is False
Found dict for key='reading', recursing
Checking key='message', check for '.' is False
Checking key='title', check for '.' is False
Checking key='additionalData', check for '.' is False
Found dict for key='additionalData', recursing
Checking key='coldstart', check for '.' is False
Checking key='title', check for '.' is False
Checking key='message', check for '.' is False
Checking key='foreground', check for '.' is False
Checking key='gcm.message_id', check for '.' is True
Found key='gcm.message_id' with dot, munged to munged_key='gcm_message_id'
Checking key='google.c.sender.id', check for '.' is True
Found key='google.c.sender.id' with dot, munged to munged_key='google_c_sender_id'
Checking key='google_c.a.e', check for '.' is True
Found key='google_c.a.e' with dot, munged to munged_key='google_c_a_e'
Checking key='google_c_a_e', check for '.' is False
Checking key='metadata', check for '.' is False
Found dict for key='metadata', recursing
Checking key='time_zone', check for '.' is False
Checking key='plugin', check for '.' is False
Checking key='write_ts', check for '.' is False
Checking key='platform', check for '.' is False
Checking key='read_ts', check for '.' is False
Checking key='key', check for '.' is False
Checking key='type', check for '.' is False
Checking key='user_id', check for '.' is False
```

JGreenlee (commented):
This appears to work well, but is there a reason for iterating twice rather than taking an "immutable" approach like this?

```python
def _remove_dots(d):
    new_dict = {}
    for k, v in d.items():
        if isinstance(v, dict):
            v = _remove_dots(v)
        new_dict[k.replace('.', '_')] = v
    return new_dict
```

```python
data = _remove_dots(data)
```

The difference may turn out to not be negligible, but given that this will be called for every usercache document (likely thousands or millions), is there a chance that iterating twice could take a noticeable amount of time?

shankari:
@JGreenlee I thought about something like that, but this uses extra memory by creating a copy of every incoming entry (this is the version you have shown above) and we are currently more memory constrained than CPU constrained.

Screenshot 2025-02-10 at 1 42 02 PM

Given that the `USERCACHE` step is currently not the bottleneck, I thought this would be fine to unblock us

If it does turn out to be the bottleneck, we can use your approach, or fix it by munging on the phone before saving so that we can remove this check on the server.

JGreenlee (approved):
Perfect, thanks for the explanation!

shankari:
No issues on staging except for some log spew. Moving this to open-access so that we can verify that it works.

```
2025-02-11T03:57:06.131-08:00 2025-02-11 11:57:06,131:DEBUG:140076466042432:suspended_subgroups=[]
2025-02-11T03:57:06.141-08:00 2025-02-11 11:57:06,140:INFO:140076466042432:Before modifying, keys_to_munge=[]
2025-02-11T03:57:06.141-08:00 2025-02-11 11:57:06,141:INFO:140076466042432:(After modifying, entry_doc.keys()=dict_keys(['key', 'platform', 'read_ts', 'time_zone', 'type', 'write_ts'])
2025-02-11T03:57:06.141-08:00 2025-02-11 11:57:06,141:INFO:140076466042432:Before modifying, keys_to_munge=[]
2025-02-11T03:57:06.141-08:00 2025-02-11 11:57:06,141:INFO:140076466042432:(After modifying, entry_doc.keys()=dict_keys(['connectUrl', 'aggregate_call_auth'])
2025-02-11T03:57:06.141-08:00 2025-02-11 11:57:06,141:INFO:140076466042432:Before modifying, keys_to_munge=[]
2025-02-11T03:57:06.141-08:00 2025-02-11 11:57:06,141:INFO:140076466042432:(After modifying, entry_doc.keys()=dict_keys(['deployment_partner_name', 'deployment_name', 'summary_line_1
```

PR #1023, opened 2025-02-11
⚡️ Update the call-specific metrics inline instead of querying the da…

shankari:
…tabase

We have made several changes to cache summary information in the user profile. This summary information can be used to improve the scalability of the admin dashboard (https://github.com/e-mission/op-admin-dashboard/issues/145) but will also be used to determine dormant deployment and potentially in a future OpenPATH-wide dashboard.

These changes started by making a single call to cache both trip and call stats https://github.com/e-mission/e-mission-server/pull/1005

This resulted in all the composite trips being read every hour, so we split the stats into pipeline-dependent and pipeline-independent stats, in 88bb35a79fda4026842e1b59d2e2e85a73893b09
(part of https://github.com/e-mission/e-mission-server/pull/1005)

The idea was that, since the composite object query was slow because the composite trips were large, we could run only the queries for the server API stats every hour, and read the composite trips only when they were updated.

However, after the improvements to the trip segmentation pipeline (https://github.com/e-mission/e-mission-server/pull/1014, results in https://github.com/e-mission/e-mission-docs/issues/1105#issuecomment-2631914594) reading the server stats is now the bottleneck.

Checking the computation time on the big deployments (e.g. ccebikes), although the time taken has significantly improved as the database load has gone down, even in the past two days, we see a median of ~ 10 seconds and a max of over two minutes.

And we don't really need to query this data to generate the call summary statistics. Instead of computing them on every run, we can compute them only when _they_ change, which happens when we receive calls to the API.

So, in the API's after_request hook, in addition to adding a new stat, we can also just update the `last_call_ts` in the profile. This potentially makes every API call a teeny bit slower since we are adding one more DB write, but it significantly lowers the DB load, so should make the system as a whole faster.

Testing done:
- the modified `TestUserStat` test for the `last_call_ts` passes

### DIFF
--- a/emission/analysis/result/user_stat.py
+++ b/emission/analysis/result/user_stat.py
@@ -7,12 +7,29 @@ from typing import Optional, Dict, Any
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.core.wrapper.user as ecwu
 
+def update_upload_timestamp(user_id: str, stat_name: str, ts: float) -> None:
+    """
+    Updates the upload timestamps in the profile
+
+    :param user_id: The user's UUID
+    :type user_id: str
+    :param stat_name: The field name that is updated
+    :type stat_name: str
+    :param ts: The timestamp to store (may not always be 'now')
+    :type ts: float
+    :return: None
+    """
+    update_data = {
+        stat_name: ts
+    }
+    update_user_profile(user_id, update_data)
+
 def update_last_call_timestamp(user_id: str, call_path: str) -> Optional[int]:
     """
     Updates the user profile with server call starts
 
     :param user_id: The user's UUID
-    :type ts: str
+    :type user_id: str
     :param call_path: Can be used to store different call stats
     :type ts: str
     :return: None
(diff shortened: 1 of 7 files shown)


shankari:
Some stats on the time taken for the `STORE_PIPELINE_INDEPENDENT_USER_STATS` calls:

Variation across users:
image

Variation across time, focused on the month of February:
image

Summary stats across time, focused on Feb (x axis is the day of the month)
image

Summary stats across time, focused on the last two days (after https://github.com/e-mission/e-mission-server/pull/1014) was moved to production)
image

 Note that the max is still well above 2 minutes, and even the median is around 10 seconds, for a query that we don't need to run.

shankari:
image

Example slow query:

```
{"find":"Stage_timeseries", "filter":{
"invalid":{"$exists":false},
"user_id":{"$binary":{"base64":"[redacted]","subType":"03"}},
"$or":[{"metadata.key":"stats/server_api_time"}]},
"sort":{"data.ts":{"$numberInt":"-1"}},"limit":{"$numberInt":"1"},
"lsid":{"id":{"$binary":{"base64":"[redacted]","subType":"04"}}},
"$db":"...ccebikes","$readPreference":{"mode":"primaryPreferred"}}
```

shankari:
@JGreenlee @TeachMeTW now that we have these additional fields, can we please put them into the admin dashboard?
https://github.com/e-mission/op-admin-dashboard/blob/01af84d177626a152d6a3968e10542db146f7431/utils/db_utils.py#L439

PR #1026, opened 2025-02-14
🚀 Perform overpass queries in batches

JGreenlee:
> The existing implementation predicted modes for sections one by one (if the section's sensed mode was IN_VEHICLE or UNKNOWN, we queried transit stops at the start and end locations respectively)
> 
> This was inefficient in 2 ways:
> - in the case of consecutive IN_VEHICLE / UNKNOWN sections, we queried twice for the same point because the first section's end is the same as the second section's start
> - we queried everything one-by-one, which led to a lot of small queries in succession
> 
> This can be optimized by doing the predictions in 2 steps.
> On the first pass, we make the predictions we can without any transit stops.
> Then we query overpass for stops near all the remaining sections' locations
> Then we do a second pass on those sections, with the transit stops at each location having now been retrieved
> 
> in match_stops.py, get_stops_near now accepts a list of locations rather than just one location. Updated TestOverpass to reflect this, and also made test_get_stops_near a little more comprehensive
> 
> Testing done:
> Tests pass. I also manually inspected the created `analysis/inferred_section` entries for `shankari_2016-07-27` and `shankari_2016-08-04` on this branch vs. master to ensure the sensed_modes were the same


JGreenlee:
TODO:
- Add limit to how many locations are in one query
  - While batching multiple locations into one query does appear to be faster overall, the current implementation does all sections at once and we risk creating a query so large that it consistently times out
- Address configurability regressions (if they are important)
  - `section.endStopRadius` is not used. Can it be removed or might it be used in the future?
  - The query template is now defined inline instead of a separate file in `conf/`. Is it important for this to be configurable?

TeachMeTW:
> * Add limit to how many locations are in one query

Added MAX_BBOXES_PER_QUERY

>   * The query template is now defined inline instead of a separate file in `conf/`. Is it important for this to be configurable?

Added optionality to load from config else use default inline

emission/net/ext_service/transit_matching/match_stops.py
```diff
@@ -67,28 +61,89 @@ def make_request_and_catch(overpass_query):
...
+    '''
+
+def get_query_for_bbox(bbox):
+    bbox_string = "%s,%s,%s,%s" % bbox
+    template = load_overpass_query_template()
+    return template.format(bbox=bbox_string)
+
+
+def get_query_for_bboxes(bboxes):
+    query = '[out:json][timeout:25];\n'
```
JGreenlee, line 99: The hard part about making the query configurable is that the query now has 2 distinct parts: the header which is defined here, and the body which is repeated for X number of locations
The current solution only makes the body configurable. Now that I am thinking about it, I am not even sure if we need the body to be configurable because the code expects those specific features when it parses the result. If someone wants to change what features are queried, they will have to make code changes anyway

If there is one thing that definitely should be configurable, I think it is the timeout threshold which goes in the header and is currently 25

emission/individual_tests/TestOverpass.py
```diff
@@ -44,19 +52,171 @@ def test_overpass(self):
...
+                             msg=f"Batch size {batch_size} produced {len(stops)} chunks; expected {expected_chunks}.")
+            for chunk in stops:
+                self.assertEqual(len(chunk), 1)
+                self.assertEqual(chunk[0]['tags'], {'dummy': True})
+
+        # Restore original settings.
+        enetm.MAX_BBOXES_PER_QUERY = original_max
+        enetm.make_request_and_catch = original_make_request_and_catch
+
+    def test_get_predicted_transit_mode_different_sizes(self):
```
JGreenlee, line 189: Can you explain what this one is supposed to do?

emission/individual_tests/TestOverpass.py
```diff
@@ -44,19 +52,171 @@ def test_overpass(self):
...
+        original_make_request_and_catch = enetm.make_request_and_catch
+        def dummy_make_request_and_catch(query):
+            # Return a dummy response: one dummy node and a "count" marker.
+            return [{'id': 100, 'type': 'node', 'tags': {'dummy': True}},
+                    {'type': 'count'}]
+        enetm.make_request_and_catch = dummy_make_request_and_catch
+
+        stops = enetm.get_stops_near(coords, 150.0)
+        # Expect one chunk per coordinate = 20 chunks.
+        self.assertEqual(len(stops), 20)
```
JGreenlee, line 115: This works but not for the reason you think it does.

Looking at your implementation of `get_stops_near`, the return value is not separated by chunk; the chunks have already been merged together.
The length of this list will match the length of coords (in this case 20), regardless of what `MAX_BBOXES_PER_QUERY` is.
So this isn't really validating the chunking at all.

Come up with a better way to test the chunking and do it with `MAX_BBOXES_PER_QUERY`=10

Hint: maybe you can measure how many API calls were made?

emission/individual_tests/TestOverpass.py
```diff
@@ -44,19 +52,171 @@ def test_overpass(self):
...
+        data_exact = list(range(1, 10))  # [1, 2, ..., 9]
+        chunk_size = 3
+        chunks_exact = list(enetm.chunk_list(data_exact, chunk_size))
+        expected_chunks_exact = [[1, 2, 3], [4, 5, 6], [7, 8, 9]]
+        self.assertEqual(chunks_exact, expected_chunks_exact)
+
+        # Case 3: Empty list
+        data_empty = []
+        chunks_empty = list(enetm.chunk_list(data_empty, chunk_size))
+        self.assertEqual(chunks_empty, [])
```
JGreenlee, line 91: I don't see much value in testing this function by itself.
Validate the behavior of `get_public_transit_stops` and that will tell us whether the chunking is working correctly

emission/individual_tests/TestOverpass.py
JGreenlee, line 1: If there are going to be a lot of new tests and they are specific to functions in `match_stops.py`, then we should probably move those to a new file called TestMatchStops or similar.

However, I don't think we really need all these new tests; we just need one that ensures the correct number of Overpass requests are made (e.g. 1 call for <=10 locations, 2 calls for <=20 locations, etc)

emission/individual_tests/TestOverpass.py
TeachMeTW: @JGreenlee So you suggest that with MAX_BBOXES_PER_QUERY=10 and 20 dummy coordinates, we expect 2 API calls?

emission/individual_tests/TestOverpass.py
TeachMeTW: The goal here is to verify that the function behaves correctly regardless of the number of stops provided. It checks that the function scales correctly and returns the expected result for a range of input sizes. Though it could be argued that is is redundant since we have test_get_predicted_transit_mode_many_chunks

JGreenlee:
Ok, I think the old `master` used `pipeline.py` and a separate branch `gis-based-mode-detection` used `rule_engine.py` (I vaguely remember)
As of 2023, `gis-based-mode-detection` is now `master` so `pipeline.py` is unused.

`gis-based-mode-detection`, created by https://github.com/e-mission/e-mission-server/pull/712, never had tests for `rule_engine.py`

So `intake_stage.py` uses `eacimr` (`rule_engine.py`) https://github.com/e-mission/e-mission-server/blob/869796b81b75f25fdc883c6926bfe5d37052d063/emission/pipeline/intake_stage.py#L166
while the tests' `runIntakePipeline` uses `eacimp` (`pipeline.py`) https://github.com/e-mission/e-mission-server/blob/869796b81b75f25fdc883c6926bfe5d37052d063/emission/tests/common.py#L210

If there's a regression with `rule_engine.py`, our tests are not going to pick it up. So this should definitely be cleaned up.

JGreenlee:
Rebased this onto `TeachMeTW:test_coverage_ms_re`
Will come back to this when https://github.com/e-mission/e-mission-server/pull/1039 is merged

shankari:
Note that the `pipeline_range.end_ts` in https://github.com/e-mission/e-mission-server/pull/1026#issuecomment-3055416673 was `1659476829.9993272` = `2022-08-02T21:47:09.999327+00:00`

So there must be other entries to process in subsequent runs. Checking these, we see a consistent pattern of ~ 30 mins to handle the full 250k batch of location points

```
2025-07-09T19:06:00.554-07:00 2025-07-10 02:06:00,554:INFO:139873138087744:**********UUID redacted-user-2: moving to long term**********
2025-07-09T19:11:59.469-07:00 2025-07-10 02:11:59,469:INFO:139873138087744:**********UUID redacted-user-2: updating incoming user inputs**********
2025-07-09T19:11:59.487-07:00 2025-07-10 02:11:59,487:INFO:139873138087744:**********UUID redacted-user-2: filter accuracy if needed**********
2025-07-09T19:11:59.495-07:00 2025-07-10 02:11:59,495:INFO:139873138087744:**********UUID redacted-user-2: segmenting into trips********** 
2025-07-09T19:14:11.345-07:00 2025-07-10 02:14:11,345:INFO:139873138087744:**********UUID redacted-user-2: segmenting into sections**********
2025-07-09T19:17:53.760-07:00 2025-07-10 02:17:53,760:INFO:139873138087744:**********UUID redacted-user-2: smoothing sections**********
2025-07-09T19:19:27.383-07:00 2025-07-10 02:19:27,383:INFO:139873138087744:**********UUID redacted-user-2: cleaning and resampling timeline**********
2025-07-09T19:23:45.712-07:00 2025-07-10 02:23:45,712:INFO:139873138087744:**********UUID redacted-user-2: inferring transportation mode**********
2025-07-09T19:27:26.662-07:00 2025-07-10 02:27:26,662:INFO:139873138087744:**********UUID redacted-user-2: inferring labels**********
2025-07-09T19:28:31.502-07:00 2025-07-10 02:28:31,502:INFO:139873138087744:**********UUID redacted-user-2: populating expectations**********
2025-07-09T19:28:32.005-07:00 2025-07-10 02:28:32,005:INFO:139873138087744:**********UUID redacted-user-2: creating confirmed objects **********
2025-07-09T19:29:02.567-07:00 2025-07-10 02:29:02,567:INFO:139873138087744:**********UUID redacted-user-2: creating composite objects **********
2025-07-09T19:29:13.960-07:00 2025-07-10 02:29:13,959:INFO:139873138087744:**********UUID redacted-user-2: storing user stats **********
```

Or to summarize:

```
2025-07-09T20:06:00.965-07:00 2025-07-10 03:06:00,965:INFO:139926104946496:**********UUID redacted-user-2: moving to long term**********
2025-07-09T20:30:43.986-07:00 2025-07-10 03:30:43,985:INFO:139926104946496:**********UUID redacted-user-2: storing user stats **********
    
2025-07-09T21:05:56.215-07:00 2025-07-10 04:05:56,215:INFO:139718333790016:**********UUID redacted-user-2: moving to long term**********
2025-07-09T21:19:24.988-07:00 2025-07-10 04:19:24,988:INFO:139718333790016:**********UUID redacted-user-2: inferring transportation mode**********
2025-07-09T21:30:51.277-07:00 2025-07-10 04:30:51,276:INFO:139718333790016:**********UUID redacted-user-2: inferring labels**********
2025-07-09T21:33:00.241-07:00 2025-07-10 04:33:00,241:INFO:139718333790016:**********UUID redacted-user-2: storing user stats **********
```

The current pipeline range is still only up to march 2023

```
2025-07-09T21:33:00.235-07:00 2025-07-10 04:33:00,235:INFO:139718333790016:For stage PipelineStages.CREATE_COMPOSITE_OBJECTS, last_ts_processed = 2023-03-28T15:28:06.970836
```

PR #1031, opened 2025-02-26
Section Optimization

TeachMeTW:
https://github.com/e-mission/e-mission-docs/issues/1105#issuecomment-2673073221

rewrite section segmentation also to read all entries from the database upfront, instead of reading on a point-by-point basis.


TeachMeTW:
I believe this change -- although not as impactful performance-wise -- the original approach might perform a separate (potentially expensive) DB query for each segmentation point—so if we have m segmentation points, we could end up with m DB calls. In the refactored version, we execute one bulk query to load all n filtered location entries (in the relevant time range), and then each segmentation point does an O(1) dictionary lookup.
 
 In Big-O terms:
Original: Approximately O(m) DB calls (each call having a constant overhead but involving I/O).

Refactored: O(n) for the initial bulk load plus O(m) for dictionary lookups, where both dictionary lookups and iterating over the loaded entries are very fast in-memory operations (which doesnt seem the case as its the same speed as original).

Even though the theoretical complexity might look like O(n + m), in practice this change SHOULD reduces the number of costly DB calls, leading to significantly better performance when m is large.

TeachMeTW:
Per Jack's comment: Refactored should be O(1) for DB calls and O(m) for dictionary lookups
 
Our assumption is that on production, dictionary lookups will significantly outperform DB lookups, even if that is not the case on a local MongoDB instance

TeachMeTW:
Originally I discovered that there were more db calls after the transition, why is that?
image

Despite it using the preloaded:
image

TeachMeTW:
I then decided to look at the stack trace:
image
image

TeachMeTW:
With Jack's assistance, it was noted that In builtin_timeseries when find_entries is called, both timeseries_db and analysis_timeseries_db get called
 
While with get_entry_at_ts only timeseries_db is called

Now I set up a temp solution:
```
    def find_entries_only_timeseries(self, key_list=None, time_query=None, geo_query=None, extra_query_list=None):
```
 
 This may not be ideal

TeachMeTW:
However, the problem of extra calls is solved:

image


The only issue is that we didnt reduce any calls from the original:
image

emission/analysis/intake/segmentation/section_segmentation.py
```diff
@@ -99,14 +108,42 @@ def segment_trip_into_sections(user_id, trip_entry, trip_source):
...
 
-    get_loc_for_ts = lambda time: ecwl.Location(ts.get_entry_at_ts("background/filtered_location", "data.ts", time)["data"])
+    # ------------------------------------------------------------------
+    # Define a function to use preloaded filtered locations instead of querying on each point.
+    # Logs indicate whether the preloaded lookup or the fallback method is used.
+    # Looks like preloaded lookup is used for all points, so the fallback method is not used -- well according to tests that is
+    # ------------------------------------------------------------------
+    def get_loc_for_ts(time):
+        if time in filtered_loc_lookup:
+            print("Using preloaded filtered location for time: %s" % time)
```
TeachMeTW, line 118: print will be changed to logging if kept; this was just for notebook analysis

shankari:
> With Jack's assistance, it was noted that In builtin_timeseries when find_entries is called, both timeseries_db and analysis_timeseries_db get called

This is not true. Look carefully at the implementation with the split keys.
That is why you are not seeing the difference https://github.com/e-mission/e-mission-server/pull/1031#issuecomment-2684169753

TeachMeTW:
With further optimization in get_location_streams, managed to get lower db calls by combining queries and doing the filtering in memory:
image

JGreenlee:
> > With Jack's assistance, it was noted that In builtin_timeseries when find_entries is called, both timeseries_db and analysis_timeseries_db get called
> 
> This is not true. Look carefully at the implementation with the split keys. That is why you are not seeing the difference [#1031 (comment)](https://github.com/e-mission/e-mission-server/pull/1031#issuecomment-2684169753)

For the record, this is the motivation behind https://github.com/e-mission/e-mission-server/pull/1032

TeachMeTW:
image


Reduced the DB calls to less than half of the original

emission/analysis/intake/segmentation/section_segmentation.py
```diff
@@ -66,8 +66,42 @@ def segment_trip_into_sections(user_id, trip_entry, trip_source):
...
+        if k == "background/bluetooth_ble":
+            ble_entries_during_trip.append(entry)
+        elif k == "background/filtered_location":
+            filtered_loc_entries.append(entry)
+        elif k == "background/location":
+            unfiltered_loc_entries.append(entry)
+        elif k == "background/motion_activity":
+            motion_entries.append(entry)
+        else:
+            pass
```
JGreenlee, line 100: Extract this to a separate function

emission/analysis/intake/segmentation/section_segmentation.py
```diff
@@ -99,14 +133,42 @@ def segment_trip_into_sections(user_id, trip_entry, trip_source):
...
+    # Looks like preloaded lookup is used for all points, so the fallback method is not used -- well according to tests that is
+    # ------------------------------------------------------------------
+    def get_loc_for_ts(time):
+        if time in filtered_loc_lookup:
+            logging.info("Using preloaded filtered location for time: %s" % time)
+            return ecwl.Location(filtered_loc_lookup[time]["data"])
+        else:
+            logging.info("Using fallback get_entry_at_ts for time: %s" % time)
+            entry = ts.get_entry_at_ts("background/filtered_location", "data.ts", time)
+            return ecwl.Location(entry["data"])
```
JGreenlee, line 148: We just preloaded all the filtered_location entries. What situation would we need this fallback?

emission/analysis/intake/segmentation/section_segmentation.py
```diff
@@ -99,14 +133,42 @@ def segment_trip_into_sections(user_id, trip_entry, trip_source):
...
+    def get_loc_for_row(row):
+        ts_val = row["ts"]
+        if ts_val in filtered_loc_lookup:
+            print("Using preloaded filtered location for row with ts: %s" % ts_val)
+            return ecwl.Location(filtered_loc_lookup[ts_val]["data"])
+        else:
+            print("Using fallback df_row_to_entry for row with ts: %s" % ts_val)
+            entry = ts.df_row_to_entry("background/filtered_location", row)
+            return ecwl.Location(entry["data"])
+
```
JGreenlee, line 168: same here

emission/analysis/intake/segmentation/section_segmentation_methods/smoothed_high_confidence_motion.py
```diff
@@ -46,7 +48,7 @@ def is_filtered(self, curr_activity_doc):
             logging.debug("%s, returning False" % activity_dump)
             return False
 
-    def segment_into_motion_changes(self, timeseries, time_query):
+    def segment_into_motion_changes(self, timeseries, time_query, preload = None):
```
JGreenlee, line 51: When will `preload` be None? Aren't we preloading everything now?

JGreenlee (changes requested):
If we are reading everything upfront, we should not need all these functions to support both preloaded and non-preloaded implementations.
Is there a reason for this choice?

emission/analysis/intake/segmentation/section_segmentation_methods/smoothed_high_confidence_motion.py
JGreenlee: (same applies for all the functions below that have `preload=None`)

emission/analysis/intake/segmentation/section_segmentation.py
TeachMeTW: @JGreenlee 
```
    # Group entries by key
    entries_by_key = {k: [] for k in keys}
    for entry in combined_entries_during_trip:
        key = entry["metadata"]["key"]
        if key in entries_by_key:
            entries_by_key[key].append(entry)

    # Now extract individual lists as needed
    ble_entries_during_trip      = entries_by_key["background/bluetooth_ble"]
    filtered_loc_entries         = entries_by_key["background/filtered_location"]
    unfiltered_loc_entries       = entries_by_key["background/location"]
    motion_entries               = entries_by_key["background/motion_activity"]
```
    
would this suffice over needing a separate function?

emission/analysis/intake/segmentation/section_segmentation_methods/smoothed_high_confidence_motion.py
TeachMeTW: @JGreenlee unless we modify this test:
```
    def testSegmentationPointsSmoothedHighConfidenceMotion(self):
        ts = esta.TimeSeries.get_time_series(self.androidUUID)
        tq = estt.TimeQuery("metadata.write_ts", 1440695152.989, 1440699266.669)
        shcmsm = shcm.SmoothedHighConfidenceMotion(60, 100, [ecwm.MotionTypes.TILTING,
                                                        ecwm.MotionTypes.UNKNOWN,
                                                        ecwm.MotionTypes.STILL])
        segmentation_points = shcmsm.segment_into_sections(ts, 0, tq)

        for (start, end, motion) in segmentation_points:
            logging.info("section is from %s (%f) -> %s (%f) using mode %s" %
                (start.fmt_time, start.ts, end.fmt_time, end.ts, motion))
        self.assertIsNotNone(segmentation_points)
        self.assertEqual(len(segmentation_points), 2)
        self.assertEqual([start.ts for (start, end, motion) in segmentation_points],
                          [1440695873.453, 1440698306.892])
        self.assertEqual([end.ts for (start, end, motion) in segmentation_points],
                          [1440698066.704, 1440699234.834])
                          
  ```
  It will fail since nothing is being preloaded into `segmentation_points = shcmsm.segment_into_sections(ts, 0, tq)`
  Shall I modify the test or keep the `None`?

Comment (TeachMeTW) [reply emission/analysis/intake/segmentation/section_segmentation_methods/smoothed_high_confidence_motion.py:51]: I just modified the test by adding the preloaded call.

**Comment by TeachMeTW** [comment]
Current tests fails, working on figuring out why

**Comment by TeachMeTW** [comment]
For example, the `FAIL: testGetDataDf (storageTests.TestTimeSeries.TestTimeSeries)` is due to the new metadata_key i added in _to_data_df.

**Comment by TeachMeTW** [comment]
I am going to try adding incremental changes to a new branch to see when it fails.

**Comment by TeachMeTW** [comment]
My hunch is the bts change I made:

My reasoning follows:
```
2025-03-03T04:38:17.5221352Z web-server-1  | ======================================================================
2025-03-03T04:38:17.5221809Z web-server-1  | FAIL: testUserInputRealData (storageTests.TestTripQueries.TestTripQueries)
2025-03-03T04:38:17.5222291Z web-server-1  | ----------------------------------------------------------------------
2025-03-03T04:38:17.5222645Z web-server-1  | Traceback (most recent call last):
2025-03-03T04:38:17.5223285Z web-server-1  |   File "/src/e-mission-server/emission/tests/storageTests/TestTripQueries.py", line 212, in testUserInputRealData
2025-03-03T04:38:17.5223856Z web-server-1  |     self.assertEqual(len(ct_df), 4)
2025-03-03T04:38:17.5224148Z web-server-1  | AssertionError: 3 != 4
```


If I go to that test, we run accross this block:

```
        # Let's retrieve them
        ts = esta.TimeSeries.get_time_series(self.testUUID)
        ct_df = ts.get_data_df("analysis/cleaned_trip", time_query=None)
        self.assertEqual(len(ct_df), 4)
```

If we go into `get_data_df`, the only thing changed there was my change to `_to_df_entry`

**Comment by TeachMeTW** [comment]
Aha!

After removing the metadata_key in _to_df_entry that I added and testing on a problematic test, it passed:

Ran 1 test in 3.775s

OK




I will now try the other tests

File: emission/analysis/intake/segmentation/section_segmentation.py [hunk 6]
```diff
@@ -84,8 +110,8 @@ def segment_trip_into_sections(user_id, trip_entry, trip_source):
                                                         ecwm.MotionTypes.STILL,
                                                         ecwm.MotionTypes.NONE, # iOS only
                                                         ecwm.MotionTypes.STOPPED_WHILE_IN_VEHICLE]) # iOS only
-        
-    segmentation_points = shcmsm.segment_into_sections(ts, distance_from_place, time_query)
+    preload = unfiltered_loc_entries + filtered_loc_entries + motion_entries
```
JGreenlee, line 113: You sort the entries by key above, lines 89-92, but then you concatenate them all together again, so you re-sort them in the `SmoothedHighConfidence*` classes, which is inefficient.
Keep them separate to avoid doing extra work

TeachMeTW: Separated as in diff inputs? That makes sense I just didn't wish to bloat the function params

emission/analysis/intake/segmentation/section_segmentation.py
```diff
@@ -66,8 +66,34 @@ def segment_trip_into_sections(user_id, trip_entry, trip_source):
...
+    # ---------------------------------------------------------------
+    # Make ONE call for multiple keys: BLE, filtered_location, location
+    # ---------------------------------------------------------------
+    keys_we_need = [
+        "background/bluetooth_ble",
+        "background/filtered_location",
+        "background/location",
+        "background/motion_activity"
+    ]
+    combined_entries_during_trip = ts.find_entries(keys_we_need, time_query)
```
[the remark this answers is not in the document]
TeachMeTW: @JGreenlee  Hmm naively moving it to `segment_current_sections` results in an error:
```
  File "/Users/rsimpson/e-mission-server/emission/analysis/intake/segmentation/section_segmentation.py", line 66, in segment_current_sections
    combined_entries = ts.find_entries(keys_we_need, time_query)
  File "/Users/rsimpson/e-mission-server/emission/storage/timeseries/builtin_timeseries.py", line 213, in find_entries
    (orig_ts_db_count, orig_ts_db_result) = self._get_entries_for_timeseries(self.timeseries_db,
  File "/Users/rsimpson/e-mission-server/emission/storage/timeseries/builtin_timeseries.py", line 242, in _get_entries_for_timeseries
    ts_db_count = tsdb.count_documents(ts_query, hint=hint_arr)
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/collection.py", line 1842, in count_documents
    return self._retryable_non_cursor_read(_cmd, session)
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/collection.py", line 1848, in _retryable_non_cursor_read
    return client._retryable_read(func, self._read_preference_for(s), s)
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/_csot.py", line 105, in csot_wrapper
    return func(self, *args, **kwargs)
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/mongo_client.py", line 1448, in _retryable_read
    return func(session, server, sock_info, read_pref)
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/collection.py", line 1837, in _cmd
    result = self._aggregate_one_result(sock_info, read_preference, cmd, collation, session)
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/collection.py", line 1703, in _aggregate_one_result
    result = self._command(
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/collection.py", line 292, in _command
    return sock_info.command(
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/pool.py", line 767, in command
    return command(
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/network.py", line 166, in command
    helpers._check_command_response(
  File "/Users/rsimpson/anaconda3/envs/emission_copy/lib/python3.9/site-packages/pymongo/helpers.py", line 181, in _check_command_response
    raise OperationFailure(errmsg, code, response, max_wire_version)
pymongo.errors.OperationFailure: error processing query: ns=Stage_database.Stage_timeseriesTree: $and
    user_id $eq BinData(3, C76A04877E5A3B17A44947BE666B36F6)
    data.end_ts $lte 1741218481.444603
    metadata.key $in [ "background/bluetooth_ble" "background/filtered_location" "background/location" "background/motion_activity"]
    $not
        invalid exists
Sort: {}
Proj: {}
Collation: { locale: "simple" }
```

Does this have something to do with the new hints changes? I will investigate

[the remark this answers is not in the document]
TeachMeTW: Hmm no it cant be it because my old version inside `segment_trips_into_sections` still works.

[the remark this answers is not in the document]
TeachMeTW: Alright got it to work.

JGreenlee:
I opened a new PR #1036 because i) I don't have access to push commits to this one and ii) the earlier commits here are irrelevant anyway after a few rounds of refinement

This PR can be closed

shankari:
Closing this PR since it has been superceded by #1036

PR #1028, opened 2025-02-24
⚡️ Specify the appropriate hints on the indices to use for various qu…

shankari:
…eries

While looking into an upgrade to mongo 8.0.4, we found that the queries were super slow. So running the pipeline after reset for a specified set of three users took ~ 15 mins against mongo 4.4.0 but over a day against mongo 8.0.4 https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2676415132

Investigating further, including by looking at the slow queries, we saw that this was due to the `metadata.key` index being used for several of the queries. Using the `metadata.key` instead of `data.ts` index for location queries, for example, took 3x the time.
https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2677074300

I think that the medium-term fix is to use a composite index. However, we can also specify a "hint" while running the query, which will encourage mongodb to use the correct index.
https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2677093069

However, the `data.ts` hint only seems to work if used with a `data.ts` filter in the query - e.g. for `background/location`. While reading the `manual/*` entries, the filter query works with `metadata.write_ts`, and then, querying by both `data.ts` and `metadata.write_ts` is slow. The only index that is moderately fast is `metadata.key`. As an aside, I assume that is why the default index was set to `metadata.key` in the first place; the query plan was optimized for the first set of queries and then subsequent calls with the same "shape" used the same index.
https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2677474876

Note that I still don't understand why `metadata.write_ts` is slow for the `manual/*` series.

Regardless, a fairly simple fix which says

> But for now, we can just say that if the sort key is `metadata.write_ts` we can
> make the hint be `metadata.key`.

With this fix, the run against mongodb 8.0.4 is also < 15 mins https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2677583319

Testing done:
https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2677583319

PR #1033, opened 2025-02-27
💩 Have the hints be optional so we can run on DocDB without errors

shankari:
In https://github.com/e-mission/e-mission-server/pull/1028 I added some hints to indicate which indices MongoDB should use for specific queries. When deploying this to staging/production, however, it turns out that DocDB does not support hints on sparse indices.
https://github.com/e-mission/e-mission-docs/issues/1109#issuecomment-2686341450

If we attempt to specify a hint on a sparse index, it is not just ignored, but the operation fails. So this is a blocker for testing out any other changes (unrelated to hints) on staging/production.

To unblock testing against real data while exploring a migration from DocDB to MongoDB, we need to remove hints on DocDB and only on DocDB. I could simply revert the previous change , but that would distort our experimentation with scalability improvements on dev against MongoDB

So I instead plumb through a option (`USE_HINTS`), defaulting to false, that controls this. While running against MongoDB in a dev environment, we should set this to True. Since we will not change the environment variables on production, we will not use hints there.

All the related changes are in this single commit so it can be reverted later.

Testing done:
- Ran the pipeline against the USAID dataset. No related errors.
- Set USE_HINTS=True and then re-ran the pipeline against the USAID dataset. No related errors.
- Set USE_HINTS=True in one of the testing workflows (test-with-manual-install) but leave it as False for the docker testing so we can test both cases.

### DIFF
--- a/.github/workflows/test-with-manual-install.yml
+++ b/.github/workflows/test-with-manual-install.yml
@@ -26,6 +26,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
+    env:
+      USE_HINTS: True
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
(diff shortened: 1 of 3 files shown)


shankari:
@JGreenlee @TeachMeTW please make sure to set `USE_HINTS=True` before you work on any further scalability changes.
Although if you don't, you should see much more impressive improvement!

PR #1040, opened 2025-03-28
🔊 🧪 Additional logs and MRE for the out of order checks

shankari:
This clears the decks for @JGreenlee to take over.

The two changes are consistent with the additional logs from
https://github.com/e-mission/e-mission-docs/issues/1122#issuecomment-2759756014
https://github.com/e-mission/e-mission-docs/issues/1122#issuecomment-2759909027
https://github.com/e-mission/e-mission-docs/issues/1122#issuecomment-2759943743
https://github.com/e-mission/e-mission-docs/issues/1122#issuecomment-2760212688

and

the third point from https://github.com/e-mission/e-mission-docs/issues/1122#issuecomment-2760212688

> use it to also generate single-day "real examples" for exploration by @JGreenlee and for inclusion in the unit tests

PR #1041, opened 2025-03-28
🐛 Fix the "multi out of order" case

JGreenlee:
#### 🐛 drop consecutive out-of-order points before segmenting

Fixes the "multi out-of-order" case in https://github.com/e-mission/e-mission-docs/issues/1122, https://github.com/e-mission/e-mission-server/pull/1040

Using pandas diff to identify out-of-order rows can fail when there are multiple consecutive out of order rows:

> That works fine with a sequence like 1, 3, 2, 4, 5 – we just drop 2
> But with a sequence like 1, 4, 2, 3, 5, we would only drop 2 leaving us with 1, 4, 3, 5 which is still out of order

Numpy has a "cumulative maximum" fn which builds an array of the largest value seen so far, giving [1, 4, 4, 4, 5]
We can then compare this to the original array, and any that don't match are considered out-of-order

Added tests for all the multi_ooo_* examples, ensuring the correct number of raw_trips & confirmed_trips and that none of the raw_trips have negative durations

### DIFF
--- a/emission/tests/analysisTests/intakeTests/TestPipelineRealData.py
+++ b/emission/tests/analysisTests/intakeTests/TestPipelineRealData.py
@@ -812,6 +812,57 @@ class TestPipelineRealData(unittest.TestCase):
             self.assertLessEqual(ct["metadata"]["write_ts"], end_run)
         self.assertEqual(countUntrackedTime, 0)
 
+    def testMultiOutOfOrderAug6(self):
+        # https://github.com/e-mission/e-mission-docs/issues/1122
+        # https://github.com/e-mission/e-mission-server/pull/1040
+        dataFile = "emission/tests/data/real_examples/multi_ooo_aug_6_2024"
+        etc.setupRealExample(self, dataFile)
+        etc.runIntakePipeline(self.testUUID)
+        ts = esta.TimeSeries.get_time_series(self.testUUID)
+
+        raw_trips = ts.find_entries(["segmentation/raw_trip"], None)
+        self.assertEqual(len(raw_trips), 15)
+        # No trips with negative duration (start_ts > end_ts)
+        for rt in raw_trips:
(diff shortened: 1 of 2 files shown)


shankari:
> Numpy has a "cumulative maximum" fn which builds an array of the largest value seen so far, giving [1, 4, 4, 4, 5]
> We can then compare this to the original array, and any that don't match are considered out-of-order

Just to clarify: running this algorithm on [1, 4, 2, 3, 5] will generate [1, 4, 5] instead of [1, 2, 3, 5]
That is the same result as running the previous OOO code until there are no out of order entries, but I am concerned that one OOO entry can result in a whole slew of deleted corrected entries.

For example, if the array was [1, 10, 2, 3, 4, 5, 6, 7, 8, 9, 11], this algorithm would delete 2-9 so that we can keep 10, which could result in loss of trajectory data

emission/tests/analysisTests/intakeTests/TestPipelineRealData.py
```diff
@@ -812,6 +812,57 @@ def testShankariUntrackedTimeJul20(self):
...
+    def testMultiOutOfOrderAug6(self):
+        # https://github.com/e-mission/e-mission-docs/issues/1122
+        # https://github.com/e-mission/e-mission-server/pull/1040
+        dataFile = "emission/tests/data/real_examples/multi_ooo_aug_6_2024"
+        etc.setupRealExample(self, dataFile)
+        etc.runIntakePipeline(self.testUUID)
+        ts = esta.TimeSeries.get_time_series(self.testUUID)
+
+        raw_trips = ts.find_entries(["segmentation/raw_trip"], None)
+        self.assertEqual(len(raw_trips), 15)
```
shankari, line 824: future fix: we can save the expected "ground truth" in and run the same tests as before. That will give us more unusal test cases for all stages, not just this one.

shankari (approved):

PR #1044, opened 2025-04-11
⚡️ preload entries during `CLEAN_RESAMPLING`

JGreenlee:
Loads loc_entries, stops_entries, and sections_entries in the processing range upfront instead of making several per-trip and per-section queries

In get_filtered_points, we filter loc_entries to the ts range of the section instead of making `find_entries` calls for each section

We can also avoid a lot of `get_entry_at_ts` and `get_entry_from_id` calls by looking at the preloaded entries, as well as by using the Timeline of raw trips and places, which we now pass down from `save_cleaned_segments_for_timeline`

### DIFF
--- a/emission/analysis/intake/cleaning/clean_and_resample.py
+++ b/emission/analysis/intake/cleaning/clean_and_resample.py
@@ -41,6 +41,7 @@ import emission.core.wrapper.localdate as ecwld
 import emission.storage.decorations.place_queries as esdp
 
 import emission.storage.timeseries.abstract_timeseries as esta
+import emission.storage.timeseries.timequery as estt
 
 import emission.storage.pipeline_queries as epq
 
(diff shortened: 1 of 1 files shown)


JGreenlee:
These changes reduce DB calls in CLEAN_RESAMPLING by about 3x overall.
There are still some repetitive calls:
- `analysis/smoothing` entries are still queried one by one. They cannot be preloaded over the entire time range because they don't have a timestamp and are only identified by the ID of the section
- Inserting the cleaned stops and cleaned sections is still done on a per-trip basis. I could rework this to bulk insert across all trips, but I haven't done it yet

These changes involve a lot of "parameter drilling" or values being passed down through several layers of functions. I wonder if it would be worthwhile to refactor `clean_and_resample.py` into a class

shankari (approved):
This LGTM!

I see what you mean about parameter drilling and why you would want to create a class. As a more general design pattern for that, I wonder if we could just read all the sensor data once at the beginning during trip segmentation, put that into a context or similar pattern, and pass it through all the stages until we get to the end [1]

Note that trip segmentation will only work on the first 250k location entries, but since each stage depends on the output of the previous stage, the stages after trip segmentation won't really care about location points beyond that in this round anyway.

That will reduce DB call time, and also lower memory churn since we will read the data once and reuse it instead of reading it over and over again. It will mean that we won't free the memory as quickly, only when we get to the end of the run for this user.

But given the advantages of lower DB access and lower memory churn, I think that the tradeoff will prove to be acceptable.

[1] Note that the context can be thread-local-storage, but since we run multiple processes, not threads, I am not sure how useful that would be. We really just need a context class that we can pass between stages.

PR #1048, opened 2025-04-17
Enhance speed and distance calculations with vectorized numpy operations

TeachMeTW:
Updated the `clean_and_resample.py` and `location_smoothing.py` files to improve performance by using vectorized numpy functions for speed and distance calculations. Adjusted precision for validation checks to 7 decimal places to accommodate small differences. Added a new `calHeading_numpy` function for calculating heading angles. Updated tests to use `assertAlmostEqual` for floating-point comparisons to ensure accuracy in results.

### DIFF
--- a/emission/analysis/classification/inference/mode/seed/section_features.py
+++ b/emission/analysis/classification/inference/mode/seed/section_features.py
@@ -10,6 +10,7 @@ import emission.core.get_database as edb
 from emission.core.get_database import get_section_db, get_mode_db, get_routeCluster_db,get_transit_db
 from emission.core.common import calDistance, Include_place_2
 from emission.analysis.modelling.tour_model.trajectory_matching.route_matching import getRoute,fullMatchDistance,matchTransitRoutes,matchTransitStops
+import emission.core.common as ec
 
 Sections = get_section_db()
 from pymongo import MongoClient
@@ -30,24 +31,8 @@ def calSpeed(trackpoint1, trackpoint2):
   else:
     return None
 
-# This formula is from:
-# http://www.movable-type.co.uk/scripts/latlong.html
-# It returns the heading between two points using 
-def calHeading(point1, point2):
-    # points are in GeoJSON format, ie (lng, lat)
-    phi1 = math.radians(point1[1])
-    phi2 = math.radians(point2[1])
-    lambda1 = math.radians(point1[0])
-    lambda2 = math.radians(point2[0])
-
-    y = math.sin(lambda2-lambda1) * math.cos(phi2)
-    x = math.cos(phi1)*math.sin(phi2) - \
-        math.sin(phi1)*math.cos(phi2)*math.cos(lambda2-lambda1)
-    brng = math.degrees(math.atan2(y, x))
-    return brng
-
 def calHC(point1, point2, point3):
-	HC = calHeading(point2, point3) - calHeading(point1, point2)
+	HC = ec.calHeading(point2, point3) - ec.calHeading(point1, point2)
 	return HC
 
 def calHCR(segment):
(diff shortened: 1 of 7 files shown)


JGreenlee (changes requested):
This looks like a great start! The new NumPy implementations in `calHeading_numpy`, `recalc_speed`, and `add_dist_heading_speed` look promising.

I am fine with using `assertAlmostEqual` in the tests because we understand the reason for the discrepancy: we are comparing against stored expected outputs that were generated using an older branch, which didn't use the numpy implementations and rounded differently.

However, the checks in `clean_and_resample` may be a different case. It doesn't look like they are comparing against stored "expected" data but rather data from within the same pipeline run. If so, this is likely because the old implementations are still being used somewhere else in the pipeline.
We don't want distance/speed/heading values within the same pipeline run to be different, even by a tiny amount.

I would recommend:
- Revert your changes to `clean_and_resample.py`
- Go ahead and fully replace `calHeading` with `calHeading_numpy` and `calDistance` with `haversine_numpy`
- See if the assertion errors in `clean_and_resample` go away

TeachMeTW:
in clean_and_resample.py, we see a different pattern:
In get_filtered_points(), the function calls eaicl.add_dist_heading_speed(resampled_loc_df) to calculate distances, speeds, and headings using our new vectorized numpy implementation.
But later, in functions like _extend_section_by_point(), we see:
```
   section["data"]["distance"] += dist_diff
   section["data"]["duration"] += ts_diff
   section["data"]["distances"].append(dist_diff)
   section["data"]["speeds"].append(dist_diff/ts_diff)
```
This is calculating speed on-the-fly with a simple division, not using our vectorized numpy function.
Similarly, in functions like _fix_squished_place_mismatch(), distance is calculated with ecc.calDistance() (the non-vectorized version) rather than the vectorized haversine_numpy.

This means there are places in the pipeline where:
The distances/speeds/headings are calculated using our new vectorized numpy implementations
Other places where they're calculated using the old point-by-point methods
This inconsistency could lead to small floating-point differences within the same pipeline run.


So I would need to 


>     * Go ahead and fully replace `calHeading` with `calHeading_numpy` and `calDistance` with `haversine_numpy`
> 
>     * See if the assertion errors in `clean_and_resample` go away

JGreenlee (changes requested):
> Some of the calDistance usage were point by point basis not vectorized; I put calDistance code it onto the haversine_numpy and had it use numpy functions -- is this a good idea? Or should I modify the places where calDistance was being used to fit the original haversine_numpy instead? Furthermore I implemented your suggestion on merging add_dist_heading_speed and recalc_speed

Good question. I see your point; we have to support a few use cases:
1. one distance between point A and point B; returns a float
  a. coordinates given as `pt` with `pt.lon`, `pt.lat`
  b. coordinates given as `[lon, lat]`
2. multiple distances between sets of points; returns ndarray of floats

So I think you're on the right track, but to me, making `haversine_numpy` support all these use cases depending on the given args adds too much complexity
Generally we do want generic functions and to not repeat ourselves, but we also don't want long bloated functions if we can avoid it

If you are keeping `calDistance` and `calHeading`, I would let them handle the differing formats. That way `haversine_numpy` expects one consistent format like before

<hr>

The `add_dist_heading_speed` / `recalc_speed` refactor lgtm

JGreenlee:
Are you seeing any performance benefit from these changes?

TeachMeTW:
Ran vectorized calHeading vs old calHeading on master vs current
Added calDistance to visualize similar gains

shankari_performance_comparison

TeachMeTW:
I decided on using modifying the existing calDistance and calHeading into wrapper functions that call the implementation functions (haversine_numpy and calHeading_numpy) as it provides several important benefits in my opinion:

- Consistent Interface - The wrappers provide a stable interface that other code depends on, allowing the underlying implementation to change without breaking existing code (not too much at least).

- Input flexibility - The wrappers accept different input formats that could help it be more generalized/handle different cases.

- Future-proofing - If we need to change the algorithm or add features like caching or logging, we can modify the implementation functions without changing all call sites.

- Backward compatibility - Existing code that calls calDistance and calHeading continues to work without modification.

These benefits significantly outweigh the minor overhead of an extra function call.

PR #1047, opened 2025-04-16
🔥 Use the `bulk_write` API to insert the incoming data instead of ins…

shankari:
…erting one by one

As we support larger deployments, the webapp is turning into a bottleneck. Looking at the logs, it looks like the bottleneck is again DocumentDB, which seems to take multiple milliseconds for each write, which adds up when we are saving many thousands of entries. In contrast, MongoDB inserts are < 1 ms each

MongoDB:

```
2025-04-16 07:27:58,093:INFO:140737036695104:Before modifying, keys_to_munge=[]
2025-04-16 07:27:58,093:INFO:140737036695104:(After modifying, entry_doc.keys()=dict_keys(['year', 'month', 'day', 'hour', 'minute', 'second', 'weekday', 'timezone'])
2025-04-16 07:27:58,093:INFO:140737036695104:Before modifying, keys_to_munge=[]
2025-04-16 07:27:58,093:INFO:140737036695104:(After modifying, entry_doc.keys()=dict_keys(['time_zone', 'plugin', 'write_ts', 'platform', 'read_ts', 'key', 'type', 'write_local_dt', 'write_fmt_time'])
2025-04-16 07:27:58,093:INFO:140737036695104:Before modifying, keys_to_munge=[]
2025-04-16 07:27:58,093:INFO:140737036695104:(After modifying, entry_doc.keys()=dict_keys(['year', 'month', 'day', 'hour', 'minute', 'second', 'weekday', 'timezone'])
2025-04-16 07:27:58,093:INFO:140737036695104:Before modifying, keys_to_munge=[]
2025-04-16 07:27:58,093:INFO:140737036695104:(After modifying, entry_doc.keys()=dict_keys(['battery_level_pct', 'battery_status', 'ts', 'local_dt', 'fmt_time'])
2025-04-16 07:27:58,093:INFO:140737036695104:Before modifying, keys_to_munge=[]
2025-04-16 07:27:58,093:INFO:140737036695104:(After modifying, entry_doc.keys()=dict_keys(['metadata', 'data', 'user_id'])
2025-04-16 07:27:58,095:DEBUG:140737036695104:Updated result for user = da63d4f7-ff86-4cca-bfc3-5fb60d81d5d3, key = background/battery, write_ts = 1722957271.6881871 = {'n': 1, 'nModified': 0, 'ok': 1.0, 'updatedExisting': True}

2025-04-16 07:27:58,095:INFO:140737036695104:Before modifying, keys_to_munge=[]
[...]

PR #1046, opened 2025-04-16
Improvements to webapp scalability to support large-scale (1200+) participant deployments

shankari:
@JGreenlee @jgu2 @rubinasingh for visibility

PR #1050, opened 2025-04-20
🔥 💩 Temporarily disable all metrics calls by aborting

shankari:
The webapp crashes that we have seen recently and attempted to mitigate using https://github.com/e-mission/e-mission-server/pull/1046 https://github.com/e-mission/e-mission-server/pull/1047 https://github.com/e-mission/e-mission-server/pull/1049 turned out to be due to the metrics calls.

A couple of the issues are:
1. metrics calls read the aggregate data, which can be large for large studies such as `uw-ebike`. We have seen 50k+ composite trips being read.
2. even for user specific metrics, it looks like the postprocessing can lead to runaway CPU/memory utilization. I am not sure why this is, but it is 100% reproducible when tested on a laptop using docker.

As a temporary fix, we disable all metrics calls. The rest of the phone code will work properly. This should address the crashes we have seen and allow the UW team to continue with recruitment

Testing done:
- without this change, the webapp crashed within 6 calls
- when only aggregate calls were aborted, the webapp crashed within 5 minutes under load
- when all metrics calls were aborted, the webapp has been stable for over 30 minutes

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -355,6 +355,8 @@ def deleteUserCustomLabel():
 
 @post('/result/metrics/<time_type>')
 def summarize_metrics(time_type):
+    logging.debug("Metrics call, finished querying values through skipping")
+    abort(503, "Metrics calls have been temporarily disabled to avoid overloading the server")
     _fill_aggregate_backward_compat(request)
     user_uuid = get_user_or_aggregate_auth(request)
 
@@ -375,6 +377,7 @@ def summarize_metrics(time_type):
     else:
         old_style = True
         is_return_aggregate = True
+
 
     app_config = request.json['app_config'] if 'app_config' in request.json else None
 


shankari:
Docker utilization without the metrics calls in the load; small bump in usage
image

Docker utilization with metrics calls; load spikes and app crashes in a minute (dotted line)
image

Docker utilization with aggregate calls aborting; spike at startup but there is a long period of load stability with low memory utilization is low until there is a runaway increase  
image

Docker utilization with aborting all metrics calls; long-term stability.
image

shankari:
@JGreenlee @rubinasingh @jgu2 for visibility

JGreenlee:
I can prioritize caching and/or optimization of aggregate metrics next week.

Where are we making webapp calls for user-specific metrics? I thought we were only computing those client-side now.

shankari:
After deploying this to the `uw-ebike` deployment yesterday, the service was stable all night with very low utilization.
Removing the rate limits to see what happens, and to allow us to catch up on the backlog.

> Where are we making webapp calls for user-specific metrics? I thought we were only computing those client-side now.

@JGreenlee I did not actually look to see the parameters of the calls on product. I just saw that there were metrics calls, and I assumed that there would be user and aggregate calls. So when generating the synthetic workload I added both aggregate and user calls, alternating with each other, to the workload.

I guess we could go back to only aborting aggregate calls, but given the potentially catastrophic consequences, I'm going to keep this in place until we know what is going on.

I also wonder why we need to read composite trips instead of confirmed trips for the metrics, do we really need the summaries of the trip trajectories to compute the emissions when we already have the computed distance? Not sure if that is the cause of the memory overload, though.

@rubinasingh the only consequence of this that the in-app dashboard will show "no data" - see the screenshot below. All other functionality in the app will continue to work. If you would like, we can send out a push notification to all users to let them know.

| android initial | android after clicking OK | iOS on launch | iOS after clicking OK |
|------|------|--------|------|
| Screenshot 2025-04-20 at 10 10 29 AM | Screenshot 2025-04-20 at 10 10 56 AM |  image | image |

shankari:
@JGreenlee @rubinasingh we could also comment out the error display when the aggregate metrics are not loaded.
I feel like it is not too onerous, particularly if everything else works, and will explain why they don't see the aggregate metrics in the dashboard. And even if I make the change right now, it will take some time for approval from Apple and Google, and then it will take time for the updates to propagate to all phones

PR #1051, opened 2025-04-20
👔 Improve the logic to whether we should skip the pipeline

shankari:
Before this, we used to skip the pipeline run in production settings if there was no new data. But if the pipeline is unstable, we can have people whose pipeline stopped running, and then they uninstalled the app. When we reset the pipeline, it will never re-run because there is no new data, potentially leading to a lot of unprocessed data sitting around.

Now that we store additional metrics in the profile, we can use that for a more principled check that will work in both production and dev settings.

The check is simple:
- if there is unprocessed data, run the pipeline, otherwise skip
- the check for unprocessed data is "do we have any location points beyond the end of the pipeline"?

Note that, since there is a little bit of fuzz in trip end detection, the check is also fuzzy. For example, if I stop moving at 3:00pm, the app will continue collecting data until 3:10 so that it can see that I have stopped moving; otherwise, it will stop at every stoplight, which is not great. But then if we end the trip at 3:00pm (or 3:05), it will look like the data from 3:05 to 3:10 is unprocessed, which will cause us to re-run the pipeline and spin our wheels.

The fuzz allows us to fix this:
- if the user is dormant, we will miss any trips in the last 10 mins, which is NBD
- if the user is not dormant, we will handle the locations as part of the next run

Testing done:
- Ran against a local dev server copy with a mixture of situations

```
For uuid=UUID('...'), last location entry is at 2025-04-03T03:42:00.507721+00:00(1743651720.507721), pipeline has run until 2025-04-03T03:42:00.507721+00:00(1743651720.507721), difference = 0:00:00, skipping
For uuid=UUID('...'), last location entry is at 2025-03-19T21:02:02.588467+00:00(1742418122.588467), pipeline has run until 2025-03-19T21:02:02.588467+00:00(1742418122.588467), difference = 0:00:00, skipping
2025-04-20 21:38:42,192:INFO:140737476424192:For uuid=UUID('...'), last location entry is at 2025-04-03T22:22:59.341000+00:00(1743718979.341), pipeline has run until 2025-04-01T18:16:29.187000+00:00(1743531389.187), difference = 2 days, 4:06:30.154000, continuing
```

### DIFF
--- a/bin/historical/migrations/add_sections_and_summaries_to_trips.py
+++ b/bin/historical/migrations/add_sections_and_summaries_to_trips.py
@@ -65,6 +65,6 @@ if __name__ == '__main__':
     args = parser.parse_args()
     split_lists = eps.get_split_uuid_lists(args.n_workers)
     logging.info("Finished generating split lists %s" % split_lists)
-    eps.dispatch(split_lists, skip_if_no_new_data=False, target_fn=add_sections_to_trips)
+    eps.dispatch(split_lists, target_fn=add_sections_to_trips)
 
 
--- a/bin/historical/migrations/populate_pipeline_range.py
+++ b/bin/historical/migrations/populate_pipeline_range.py
@@ -39,5 +39,5 @@ if __name__ == '__main__':
     args = parser.parse_args()
     split_lists = eps.get_split_uuid_lists(args.n_workers)
     logging.info("Finished generating split lists %s" % split_lists)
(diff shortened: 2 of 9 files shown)


shankari:
and the CPU and memory bump was low
Screenshot 2025-04-20 at 3 21 08 PM

shankari:
@JGreenlee  for visibility

shankari:
@JGreenlee for visibility

PR #1052, opened 2025-04-21
∦ Tighten the mechanism to have only one running copy of the pipeline

shankari:
Our pipeline cannot have two copies running in parallel at the same time; we will end up with two copies or none. The expectation is that we have a logically linear timestream and jumping in and out of it intermittently will lead to incorrect results.

We do this by using the `curr_run_ts` as a semaphore. If the `curr_run_ts` of a particular user was set, then other pipelines running in parallel would detect it and would not process that stage.

However, when running in production, if we launched services every hour, but the pipeline took more than an hour to run, we would end up with two pipelines running in parallel, but one much futher along than the other.

Theoretically, this is OK. The later pipeline will process the initial states properly before failing at the conflict. But then when the stage with the conflict completes, it will reset the `curr_run_ts`.

Concretely, let us assume that the first run (run1) of the pipeline started at 8am and it took 3 hours to complete, so it ended at 11am. It was stuck for a long time at the bottleneck on S5

In the meantime, a new run (run2) started at 9am.

So some steps may be:

```
-  9:10: run1 - > S5, run2 - > S1
-  9:20: run1 - > S5, run2 - > S2
-  9:30: run1 - > S5, run2 - > S3
-  9:40: run1 - > S5, run2 - > S4
-  9:50: run1 - > S5, run2 - > S5 <- - - -  CRASH - - - - - - >
- 10:50: run1 - > S5, run2 - > S6 <- - - -  run2 has leapfrogged S5, but that just means that subsequent steps will not update their `last_processed_ts`   - - - - - - >
- 10:50: run1 - > S6, run2 - > S6 <- - - - `curr_run_ts` of S5 is reset by run1 ---->
- 11:00  run1 - > S7
- 11:10  run3 - > S1 (but after the work that run2 did)
```

So it should work, but:
- we will use up 2x the resources and
- if run1 is killed and we have to reset the pipeline, then all the work that run2 did will be wasted

Let's keep it simple, and not even start `run2` when `run1` is still in progress, since it is likely to fail anyway. This is also consistent with the original vision of the `curr_run_ts` as a semaphore, so it is likely to be more robust.

If we ever build in additional support for manipulating the pipeline at the stage level - for example, by changing the reset to work on a per-stage basis, or add additional resources to the cluster, we can revisit this decision

Testing done:

Tests (including the newly added ones) in `TestPipelineCornerCases` pass

```

Ran 6 tests in 1.940s

OK
```

Running this against a local dump with a few invalid states . Don't see any failures because of the `curr_run_ts.

```
Pipeline for e0e0c5bb-5bfb-4d8f-88b6-2fd602aef234 has in_progress_stages=1, skipping
Pipeline for 220f4503-631a-4d68-a937-431fa36472b5 has in_progress_stages=2, skipping
2025-04-20T22:28:56.049586-07:00**********UUID e9abc855-abfd-413b-877b-6018ba0701ca: moving to long term**********
2025-04-20T22:28:56.111654-07:00**********UUID ed2a2933-95ea-49ea-8b39-29f1ec9758da: segmenting into sections**********
2025-04-20T22:28:56.193898-07:00**********UUID f6adb9f3-4782-4d01-80e6-8722fb47eb93: cleaning and resampling timeline**********
2025-04-20T22:28:56.368976-07:00**********UUID e9abc855-abfd-413b-877b-6018ba0701ca: updating incoming user inputs**********
```

### DIFF
--- a/emission/pipeline/intake_stage.py
+++ b/emission/pipeline/intake_stage.py
@@ -94,12 +94,23 @@ def run_intake_pipeline_for_user(uuid):
         last_loc_ts = arrow.get(user_profile.get('last_location_ts', 0))
         last_proc_time = user_profile.get('pipeline_range', {}).get('end_ts', None)
         last_proc_ts = arrow.get(last_proc_time) if last_proc_time is not None else arrow.get(0)
-        fmt_string = f"For {uuid=}, last location entry is at {last_loc_ts}({last_loc_ts.timestamp()}), pipeline has run until {last_proc_ts}({last_proc_ts.timestamp()}), difference = {last_loc_ts - last_proc_ts}({(last_loc_ts.timestamp() - last_proc_ts.timestamp())})"
-        if  (last_loc_ts.timestamp() - last_proc_ts.timestamp()) <= 10 * 60: # 10 minutes
-            print(f"{fmt_string}, skipping")
+        ts_diff = last_loc_ts.timestamp() - last_proc_ts.timestamp()
+        fmt_dormant_user_check = f"For {uuid=}, last location entry is at {last_loc_ts}({last_loc_ts.timestamp()}), pipeline has run until {last_proc_ts}({last_proc_ts.timestamp()}), difference = {last_loc_ts - last_proc_ts}({(ts_diff)})"
+        if  (ts_diff) <= 10 * 60: # 10 minutes
+            print(f"{fmt_dormant_user_check}, skipping")
             return
         else:
-            logging.info(f"{fmt_string}, continuing")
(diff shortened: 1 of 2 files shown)


shankari:
Increase in DB CPU and memory utilization during the run was not that bad
Screenshot 2025-04-20 at 10 55 58 PM

shankari:
WIth this, both the changes outlined in https://github.com/e-mission/e-mission-docs/issues/1105#issuecomment-2798124327 are fixed. I will proceed with the next steps there.

PR #1054, opened 2025-04-24
🔧 use confirmed trips for agg metrics

JGreenlee:
Step 1 of https://github.com/e-mission/e-mission-docs/issues/1126 is to simply use confirmed trips instead of composite trips to compute metrics. Confirmed trips are smaller than composite trips, so this will use less memory. e-mission-common is already set up to handle both types the same way.

Testing done:
https://github.com/e-mission/e-mission-docs/issues/1126#issuecomment-2828147799


shankari (approved):
This is approved, but I am going to keep it in my "back pocket" while seeing how long it takes to implement the "cache once a day, as part of the public dashboard" solution.

PR #1058, opened 2025-04-24
Non-Test Changes for Rule Engine

TeachMeTW:
This PR implements two main changes:

Rule Engine Migration:
- Migrates mode inference to use rule_engine for improved maintainability and performance
- Enables more consistent handling of mode inference rules
- Addresses issues with existing mode inference implementation

Transit Matching Caching:
- Implements caching mechanism for Overpass API results in match_stops.py
- Significantly reduces API calls to external services
- Improves performance for repeated transit matching operations
- Cache files are stored as JSON for easy inspection and debugging

Related Issues:
- Related to https://github.com/e-mission/e-mission-server/pull/1026#issuecomment-2695408459


shankari:
@JGreenlee I am not sure we need to cache values here; the public server has been able to handle our load from production for a while. And even if we do cache, do we have to check in the cached files? Which trips do they represent? (I assume the ones from the automated tests).

And finally, we are using geofabrik on production, but it is not clear that other deployers are. If they use the public server, we will cache the results on their filesystem. Which is potentially OK, but we should then make sure that we don't run out of disk space.

Note also that in general, if the pipeline is run "serverless" it will not have access to persistent storage other than the database. And even if we do store in the database, caching values precludes getting refreshes when there are new transit stops (e.g. a new BRT line https://metrobrtproject.com/)

So basically, I think that caching the data is engineering overkill and should be implemented only if needed after the batching is in place

JGreenlee:
> So then why are they part of the "non-test changes" for the rule engine. They should be part of https://github.com/e-mission/e-mission-server/pull/1039

These changes were part of #1039 and you had requested it to be split into #1057 and #1058.

> it would be better if they would be handled through mocking instead of modifying the source code for the overpass retrieval.
> 
> e.g. something like https://medium.com/@everythingismindgame/mocking-the-network-call-with-unittest-python-51e9f42a28f9

Yes, much better. This can easily be converted to a mock, since it basically just puts a wrapper around `match_stops.make_request_and_catch`

PR #1060, opened 2025-04-25
📊 use Stage_agg_metrics DB for preloaded aggregate metrics

JGreenlee:
https://github.com/e-mission/e-mission-docs/issues/1126

define indices for Stage_agg_metrics collection, which is where the aggregate metrics will be stored when they are computed by the public dashboard

new function get_agg_metrics_from_db to read precomputed metrics from Stage_agg_metrics

In cfc_webapp.py:
- remove the 503 abort
- switch to using metrics.get_agg_metrics_from_db remove a bunch of old code that was left to support old versions of the phone UI, which are obsolete now

### DIFF
--- a/emission/net/api/cfc_webapp.py
+++ b/emission/net/api/cfc_webapp.py
@@ -346,7 +346,7 @@ def getMetrics(time_type):
     logging.debug("getMetrics with time_type %s and request %s" %
                   (time_type, request.json))
     if time_type != 'yyyy_mm_dd':
-        abort(404, "Metrics calls only supported for yyyy_mm_dd time type")
+        abort(404, "Please upgrade to continue using the app dashboard")
     
     user_uuid = get_user_or_aggregate_auth(request)
     start_ymd = request.json['start_time']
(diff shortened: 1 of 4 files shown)


emission/net/api/cfc_webapp.py
```diff
@@ -354,49 +342,19 @@ def deleteUserCustomLabel():
...
 @post('/result/metrics/<time_type>')
-def summarize_metrics(time_type):
-    logging.debug("Metrics call, finished querying values through skipping")
-    abort(503, "Metrics calls have been temporarily disabled to avoid overloading the server")
-    _fill_aggregate_backward_compat(request)
+def getMetrics(time_type):
+    logging.debug("getMetrics with time_type %s and request %s" %
+                  (time_type, request.json))
+    if time_type != 'yyyy_mm_dd':
+        abort(404, "Metrics calls only supported for yyyy_mm_dd time type")
```
shankari, line 349: In case there is anybody who has still not updated (e.g. have the old app installed, but have turned off auto-updates), I think we should tell them to update so that things keep working. Maybe we can return a different error message?

```suggestion
        abort(404, "Please upgrade to continue using the app dashboard")
```

shankari (approved):
LGTM, and thanks for cleaning this up while you are making changes.
I will merge https://github.com/e-mission/em-public-dashboard/pull/195 first so that the metrics will be in place when the app tries to read it.

PR #1061, opened 2025-04-25
Bug fix to load_timeline_for_day_and_user

TeachMeTW:
Addresses the issue where data wasn't being processed by the pipeline due to date being set to 1970.

This update tracks most recent location timestamp from background/location. Also tracks latest phone data timestamp from metadata.

Some real examples result in key error in case of things, I added checks for fields to prevent these key errors

After loading the data, we now update the user profile with these timestamps to reflect the newly loaded data. It also initializes the pipeline_range field if doesn't exist.

These fixes should ensure that profiles are properly updated with timestamps so it won't be skipped.


JGreenlee (approved):

shankari:
superceded by #1077

PR #1063, opened 2025-05-02
Csv timeseries

KG2468:
#1121


shankari:
@KG2468 one high level comments before @JGreenlee and I review this

**Are the files in `emission/storage/timeseries/data/` the raw files that you downloaded from the TSDC?** If so (I can't view them because they are too large), they should not be checked in to the repo.
   a. The files are not currently fully public; people can download them from the TSDC, but they have to register and click through the terms and conditions
   b. While you can and should test against them, we are building a system that will work with any csv file, not a one-off script to process a particular dataset

When you remove the files, please make sure to edit your commit (`git rebase -i HEAD~3` or similar) to remove the files completely instead of introducing a new commit to remove them. Unless you edit the commit, the files will be in git history so will still be accessible via the correct `git checkout`

PR #1067, opened 2025-05-06
👔 Reset the cached pipeline state in the profile as well

shankari:
Before this change, the pipeline reset script would change the `get_pipeline_state_db()` but not the cached pipeline start and end in the profile.
    
Since we now use the profile to decide whether to run the pipeline, this led to issues where the pipeline for a user would not re-run even after the pipeline was reset.
https://github.com/e-mission/e-mission-docs/issues/1105#issuecomment-2852181010
    
I wonder if we should use only the pipeline state to decide whether or not to run the pipeline, instead of the profile. It seems like restructuring to use a single source of truth would be good.
    
Testing done:
Ran it to reset our test user back to `2024-07-30`; verified that the value in the profile was reset
Verified that the pipeline was not skipped on the run after the reset

### DIFF
--- a/bin/locust/webapp.py
+++ b/bin/locust/webapp.py
@@ -73,7 +73,8 @@ class PhoneAppUser(HttpUser):
 
     @task(2)
     def usercache_get(self):
-       self.client.post("/usercache/get", json={"user": self.opcode})
+       response = self.client.post("/usercache/get", json={"user": self.opcode})
+       print(f"Got {response.text} in response to the call")
 
     @task(2)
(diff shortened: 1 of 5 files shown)


TeachMeTW:
Added test in https://github.com/e-mission/e-mission-server/pull/1068
I'll fix the commits once 1067 is merged
I know copilot advised mocking but I believe using intake pipeline and real examples was more intuitive and easier to grasp

shankari:
Also made some additional logging changes to clean up this branch before I am done

PR #1069, opened 2025-05-07
Tests for restart checking analysis

TeachMeTW:
Unit tests for the restart detection functionality in restart_checking.py to document and validate how np.searchsorted maps timestamps between different data streams.

## Changes

- Created unit tests for both ongoing_motion_in_loc_df and tracking_restarted_in_loc_df functions
- Added synthetic data tests (first two tests) that verify all edge cases (events before/between/after location points) 
- Implemented a real-world test using shankari reboot data as that is the perfect place to test restart checking

## Testing

Ran and passed. Test output includes detailed logging of how timestamps map to indices, providing clear examples:
`Restart timestamp 1673851361.154 maps to location index 29 with ts 1673916151.673`

see:
https://github.com/e-mission/e-mission-docs/issues/1105#issuecomment-2705741540

### DIFF
--- a/emission/tests/analysisTests/TestRestartChecking.py
+++ b/emission/tests/analysisTests/TestRestartChecking.py
@@ -1,0 +1,223 @@
+import unittest
+import pandas as pd
+import numpy as np
+import logging
+import datetime
+
+import emission.core.wrapper.transition as ecwt
+import emission.core.wrapper.motionactivity as ecwm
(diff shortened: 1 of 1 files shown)


emission/tests/analysisTests/TestRestartChecking.py
JGreenlee, line 1: I think this should be under analysisTests.intakeTests

PR #1076, opened 2025-07-11
🧪 add tests for the rule engine implementation of mode inference

JGreenlee:
#### 🧪 add tests for the rule engine implementation of mode inference

There are 2 implementations of the MODE_INFERENCE stage in the codebase: i) old version: pipeline.py (eacimp); and new version: rule_engine.py (eacimr). https://github.com/e-mission/e-mission-server/pull/1026#issuecomment-2695408459
rule_engine is used in the pipeline, while pipeline.py was used in tests, meaning we effectively didn't have tests for the actual behavior of the MODE_INFERENCE stage

Since rule_engine involves a lot of queries to Overpass API, we should be careful about not abusing the API with repetitive queries every time we run tests, so this testing approach uses mocks to implement caching by wrapping the functions that perform API calls.

The RuleEngine tests seek to test several mode predictions such as walking, cycling, driving, etc based on different factors.
Cases include empty sections, AirOrHSR, Motorized, Unknown, Combination
Added tests with shankari_2023-07-18_xmastrain (https://github.com/e-mission/e-mission-docs/issues/1124) to ensure rule engine fails gracefully (i.e. creates an UNKNOWN section rather than quitting)

\- regenerated "expected" outputs since they change now that we are using rule_engine

additional context: https://github.com/e-mission/e-mission-server/pull/1039


JGreenlee:
I am guessing you're referring to `jack_untracked_time_2023-03-12.inferred_section.expected_composite_trips`.
Yes the inferred sensed mode switches from `BUS` to `UNKNOWN`, but the cleaned sensed mode was always `UNKNOWN` to begin with

image

When the cleaned sensed mode is UNKNOWN, the rule engine goes to `get_unknown_prediction` (below). I am pretty sure it ended up in the last else block where we have exhausted our possibilities and give up
https://github.com/e-mission/e-mission-server/blob/c684fca916a63f15d0cc8bd8bcf2f553353e2a1f/emission/analysis/classification/inference/mode/rule_engine.py#L152-L175

I remember this day; I was on a road trip to Gatlinburg TN by car.
I don't know why it was not IN_VEHICLE. This was probably recorded on my OnePlus phone.

JGreenlee:
There's actually no motion activity in `jack_untracked_time_2023-03-12` at all, so that would explain it

PR #1099, opened 2026-07-31
✨ Initial implementation of the vehicle library

shankari:
- Bikeep integration with separate endpoints
- Tests for bikeeep integration

PR #1100, opened 2026-08-04
✨ Plumb through support for modifiable, user-specific state

shankari:
As OpenPATH gets more complex, we need to store additional user-specific state. This is not really a time-series use case, since the state is designed to be updated, and we will want to retrieve the most recent version. It could be implemented as a timeseries if we want to support edit history and/or versioning.

We already store models as user-specific modifiable data. We extend this to store user-specific modifiable state as well.

This allows us to avoid overloading the user profile, and separates out state that is only relevant to a particular feature.

Changes:
➕ implement a new abstract/builtin module under "modifiable" storage ➕ show how it will work for two types of state (vehicles and payments) related to the vehicle library (https://github.com/e-mission/e-mission-server/pull/1099) ➕ add tests for the new module
➕ minor formatting fix for the model storage tests to match the new tests

### DIFF
--- a/emission/core/get_database.py
+++ b/emission/core/get_database.py
@@ -264,6 +264,18 @@ def get_model_db():
     ModelDB.create_index([("metadata.write_ts", pymongo.DESCENDING)])
     return ModelDB
 
+
+def get_state_db():
+    """
+    " Stores modifiable state documents keyed by state type for each user.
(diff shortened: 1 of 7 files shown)


shankari:
@JGreenlee for visibility

